PR #177, opened 2014-11-08
refractor torrent to make it easier to add new sources

miketamis:
moved the storage creation into a module ‘torrent-manager’ this will
allow for further source to be added easily.


feross:
Sorry this PR never got merged, but it's way out of date now so I'm going to close it.

Issue #234, opened 2015-01-04
Use native browser sha1 implementation when possible


feross:
Most of the the browser's time is spent in `git-sha1` when creating a torrent. This is unsurprising, since the file must be split into pieces and hashed. But, this currently locks the entire browser for seconds.

Is the browser's native sha1 implementation faster? It's already shipped in Chrome and Firefox. See: http://caniuse.com/#feat=cryptography

- [x] Test that `crypto.subtle` is actually faster
- [x] Compare `Rusha` and current `git-sha` as fallback for insecure origins and older browsers
- [ ] File bug on Chrome and the spec to complain that `crypto.subtle` is only available on `https:` sites (https://instant.io is https, but other sites that use webtorrent might not be). Comment on this issue: https://code.google.com/p/chromium/issues/detail?id=373032

Related issue: https://github.com/feross/create-torrent/issues/4

astro:
It looks like [simple-sha1](https://github.com/michaelrhodes/simple-sha1) has the right aims: to provide the fastest implementation available. crypto.subtle support should be implemented there.

feross:
This is done now. Released as 0.22.0.

- `crypto.subtle` is actually faster
- `Rusha` is faster than `git-sha1`
- Will file bugs on chrome later

PR #236, opened 2015-01-05
replace git-sha1 with simple-sha1

astro:
See: #234

I would like to use the native crypto module in node.


feross:
The native crypto module is already used in node. git-sha1 handles that case. See: https://github.com/creationix/git-sha1/blob/master/git-sha1.js#L11

I don't mind moving to simple-sha1, but first we need to confirm:

- [ ] simple-sha1 (which uses Rusha internally) is actually faster than the git-sha1 implementation
- [ ] the file size difference when browserified is negligible
- [ ] swap out git-sha1 in all the webtorrent modules as it's used in several, and we don't want to include multiple implementations (I can handle this one)

feross:
Confirmed all 3 things. Thanks for the PR!

PR #240, opened 2015-01-08
fix Piece.verify method and update README

n1t0:
Since sha1 is now async, this was failing with the following error:

```
.../node_modules/webtorrent/node_modules/simple-sha1/index.js:6
    cb(hash)
    ^
TypeError: undefined is not a function
    at .../node_modules/webtorrent/node_modules/simple-sha1/index.js:6:5
    at process._tickCallback (node.js:419:13)
```

I also updated the README to add some available options when adding a torrent.


feross:
Thanks, released as 0.22.1.

Issue #201, opened 2014-12-08
<audio> tag streaming


ArtskydJ:
Like #144.

Is something like this supported?

If it is not yet, might it be in the future?

Somewhere in `index.html`
```html
	<script src="/bundle.js"></script>
	<audio id="audioElement"></audio>
```

Turn this into `bundle.js` using browserify.
```js
var WebTorrent = require('webtorrent')
var client = new WebTorrent()

client.download(magnet_uri, function (torrent) {
	var file = torrent.files[0] // Let's say the first file is an audio file...
	var audio = document.getElementById('audioElement') // Get audio element
	file.createReadStream().pipe(audio) // Stream the audio into the <audio> tag
})
```

feross:
This is fixed now, thanks to @mathiasvr! You can see it on https://instant.io.

ArtskydJ:
Sweet; thanks to you both!

PR #245, opened 2015-01-16
Add <audio> tag streaming support for mp3 files

mathiasvr:
Quick fix for issue #201

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -99,12 +99,15 @@ FileStream.prototype.pipe = function (dst) {
   var self = this
   var pipe = stream.Readable.prototype.pipe
 
-  if (dst && dst.nodeName === 'VIDEO') { // <video> tag
+  if (dst && dst.nodeName === 'VIDEO' || dst.nodeName === 'AUDIO') { // <video> and <audio> tag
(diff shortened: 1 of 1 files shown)


lib/file-stream.js
```diff
@@ -99,12 +99,15 @@ FileStream.prototype.pipe = function (dst) {
   var self = this
   var pipe = stream.Readable.prototype.pipe
 
-  if (dst && dst.nodeName === 'VIDEO') { // <video> tag
+  if (dst && dst.nodeName === 'VIDEO' || dst.nodeName === 'AUDIO') { // <video> and <audio> tag
```
feross: The if statement should be `dst && (dst.nodeName === 'VIDEO' || dst.nodeName === 'AUDIO')` because of order of operations and `dst` could be undefined. I'll fix.

feross:
Awesome, thanks!

feross:
Thanks! This was a simple change with awesome results! Released as webtorrent@0.26.0 and deployed to https://instant.io :+1:

lib/file-stream.js
mathiasvr: Yes you are right! :+1:

PR #247, opened 2015-01-16
Added ./examples/ directory

ArtskydJ:
Added the three examples from the readme
Added an example from a pastebin
Copied and modified an example from the readme


feross:
This is great, @ArtskydJ!

Issue #248, opened 2015-01-17
Reduce memory usage


alex-rokabilis:
hello, i started using webtorrent in a nodejs app and quickly i realized that in order to download a file webtorrent uses the main memory,I understand that this is necessary because it use torrents, pieces in order etc.. But this could cause problems in big files. Let say a file of 2gb would be loaded in main memory causing a large memory usage and in fact causing node to crash because it has a limit of 1gb memory limit, this could be solved with passing extra flags to force node use more memory but I don't think this is really a solution. I think the solution in this is to free buffers when a piece is verified and saved to disk, I already made some changes in my local project to setting null in piece buffer and block properties and then forcing v8's garbage collector to run. This (maybe not very good) solution worked and I managed to download a file of 3gb (before this i couldn't download more than 1gb).

PS. Still it use much memory, but in node working levels. Am I missing a reference to another buffer??

feross:
This is not supposed to happen. In node, webtorrent uses your system's tmp folder to store pieces to disk and only reads them back into memory from disk when a peer requests them. After that, they should be garbage collected.

The browser is another story. Right now, we store the file entirely in memory in the browser because I haven't investigated what storage to use yet (it's not a top priority at the moment).

Could you send a PR with your changes? We definitely can do a better job setting references to `null` to help with GC. I'll merge your PR.

FWIW, I've streamed video files up to 5GB without problems. What version of node/iojs are you using? Platform? Version of webtorrent?

Svenskunganka:
@feross since the Filesystem API support has been dropped, we could maybe look into [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) ([CanIUse](http://caniuse.com/#feat=indexeddb)) for storage?

astro:
For the record: I observed increasing memory usage too. Maybe it's the storage layer not freeing pieces after validation.

alex-rokabilis:
I use 
webtorrent v0.23.0
node v0.10.33
ubuntu 14.04

In the original webtorrent i cannot see freeing any memory, node instance getting bigger and bigger
as the downloading continues( I changed the tmp folder but i dont think this is a problem?). In fact even running webtorrent from the terminal, system monitor shows the same growing memory usage as if i run it from inside my node app. Can you please tell me in which file and line you deallocate the buffers, because i couldn't figure it when I inspected the lib files.

astro:
[...]

PR #249, opened 2015-01-21
Storage memory leak fix

astro:
For your consideration, attempts to solve #248


feross:
Was this actually merged? I don't see it in `master`.

astro:
reopened as #251

Issue #235, opened 2015-01-04
Move SHA1 hashing to another process/worker


feross:
Creating SHA1 hashes for large multi-gigabyte files can take up **a lot** of CPU time.
- [ ] In the browser, we should investigate moving this work to a web worker with [transferrable objects](http://updates.html5rocks.com/2011/12/Transferable-Objects-Lightning-Fast).
- [ ] In node, we should investigate spawning worker processes to do hashing.

astro:
Transferrable objects worked well for Bitford:
* https://github.com/astro/bitford/blob/master/src/store.js#L616
* https://github.com/astro/bitford/blob/master/src/sha1-worker.js#L18

astro:
It is not clear to me how to integrate this into the browserify process. Maybe with [workerify](https://github.com/shama/workerify)?

feross:
Managed to make `client.seed` much faster by skipping the piece verification in `storage.js`. It's unnecessary since we're the first seeder of a new file.

Before: 
```
createTorrent.parseInput: 0.598ms
index.js:206 createTorrent: 6654.848ms
index.js:211 add: 186.781ms
index.js:220 tasks (storage.load): 6906.296ms
index.js:224 seed: 13752.465ms
```

After:

```
createTorrent.parseInput: 0.626ms
index.js:209 createTorrent: 6656.904ms
index.js:214 add: 419.079ms
index.js:223 tasks (storage.load): 256.499ms
index.js:227 seed: 7337.118ms
```

Total time for 250MB file almost cut in half. 13s to 7s.

feross:
With this PR (https://github.com/michaelrhodes/simple-sha1/pull/1) to `simple-sha`, the time to prepare a 250MB file is only 2 seconds.

feross:
Aaaand, using the web crypto API gets us down to <1 second to seed a 250MB file!

PR out: https://github.com/michaelrhodes/simple-sha1/pull/3

We're now officially in the territory of C speeds. We probably can punt on web workers for a while. This is so much faster than it was before (13.7 seconds).

feross:
Released 0.22.0 using `simple-sha1` for *MUCH* faster perf in the browser.

Svenskunganka:
@ferros This is amazing! Really needed! We should try to push the larger BitTorrent client developers to implement WebRTC as WebTorrent is revolutionary to BitTorrent!

feross:
[...]

PR #250, opened 2015-01-21
Continuous hashing

### LINKED ISSUE #237: Continuous/streaming SHA1 hashing
I've had good experience with this in Bitford and would like to work on it. But first I would like to hear your opinion. My arguments for it:
* Shorter and more evenly distributed 100% CPU situations. It will smoothen things much more even with Web Workers, because...
* I just discovered that the 31C3 video torrents come with a piece length of 16 MB. After receiving the last few 16 KB chunks we will be able to deliver the media to the player frontend much faster if there were only a few KB to hash instead of the full piece.

As neither Rusha nor subtle support streaming SHA1 we will have to emulate it in the browser for now. Still, the node.js crypto module can do it well. Rusha could probably be modified in the future. The subtle interface only provides the oneshot method. :-(

astro:
Based on #249. Fixes #237 and provides groundwork for #235.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -451,7 +451,19 @@ function drawTorrent (torrent) {
       }
       var bar = ''
       for (var j = 0; j < piece.blocks.length; j++) {
-        bar += piece.blocks[j] ? '{green:█}' : '{red:█}';
+        switch(piece.blocks[j]) {
+        case 0:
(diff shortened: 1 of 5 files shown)


feross:
Was this actually merged? I don't see it in `master`.

astro:
reopened as #252

PR #251, opened 2015-01-21
Storage leakfix

astro:
For your consideration, attempts to solve #248

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -451,7 +451,19 @@ function drawTorrent (torrent) {
       }
       var bar = ''
       for (var j = 0; j < piece.blocks.length; j++) {
-        bar += piece.blocks[j] ? '{green:█}' : '{red:█}';
+        switch(piece.blocks[j]) {
+        case 0:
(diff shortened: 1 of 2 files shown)


astro:
Hmmm... something is breaking the fs-storage tests.

astro:
Honestly, I would like to keep the piece backing buffer separated in blocks so we can persist them away ASAP, to lift pressure from the GC.

astro:
I would like to know your opionion on the idea. I've got a preliminary implementation here:
https://github.com/astro/webtorrent/tree/per-block-buffers
Block structures cause quite some overhead, so we'll need different handling for pieces with a blank/written state. My experiment has shown really low memory usage by partially loaded pieces. This will benefit performance whenever we can move received data off memory, ie. to files or IndexedDB.

feross:
Trying to debug why this causes the tests to fail.

feross:
I couldn't figure out why the tests are failing, so I had to revert merging this. I'll take another look soon.

PR #252, opened 2015-01-21
Continuous hashing

### LINKED ISSUE #237: Continuous/streaming SHA1 hashing

astro:
Based on #251. Fixes #237 and provides groundwork for #235.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -451,18 +451,22 @@ function drawTorrent (torrent) {
       }
       var bar = ''
       for (var j = 0; j < piece.blocks.length; j++) {
-        switch(piece.blocks[j]) {
-        case 0:
-          bar += '{red:█}';
-          break;
-        case 1:
-          bar += '{blue:█}';
-          break;
-        case 2:
+        if (j < piece.blocksHashed) {
           bar += '{green:█}';
-          break;
-        default:
-          throw 'Invalid block state: ' + piece.blocks[j]
+        } else {
+          switch(piece.blocks[j]) {
+          case 0:
+            bar += '{red:█}';
+            break;
+          case 1:
+            bar += '{yellow:█}';
+            break;
+          case 2:
+            bar += '{blue:█}';
+            break;
+          default:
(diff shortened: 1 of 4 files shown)


bin/cmd.js
```diff
@@ -438,21 +438,43 @@ function drawTorrent (torrent) {
...
+          default:
+            throw 'Invalid block state: ' + piece.blocks[j]
+          }
+        }
       }
       clivas.line('{4+cyan:' + i + '} ' + bar);
       linesremaining -= 1
     }
+    clivas.line(
+      '{red:storage mem:} {bold:' + Math.ceil(storageMem / 1024) + ' KB}  '
```
feross: We should use the `prettyBytes` function instead of calculating this manually.

bin/cmd.js
```diff
@@ -438,21 +438,43 @@ function drawTorrent (torrent) {
...
         continue;
       }
       var bar = ''
       for (var j = 0; j < piece.blocks.length; j++) {
-        bar += piece.blocks[j] ? '{green:█}' : '{red:█}';
+        if (j < piece.blocksHashed) {
+          bar += '{green:█}';
+        } else {
+          switch(piece.blocks[j]) {
+          case 0:
```
feross: Should export the constants from `storage.js` so we don't have magic numbers here.

bin/cmd.js
astro: Ideally, we wouldn't need to output this at all. :-)

PR #253, opened 2015-01-22
torrent: cancelBlocks on wire 'close'

astro:
We may need to add even more logic: malicious peers may hold back responses indefinitely. What do you think about request timeouts?

(Albeit not for malicious reasons, I once got no response from certain peers when choosing a wrong request size (> 16 KB)).

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -514,6 +514,9 @@ Torrent.prototype._onWireWithMetadata = function (wire) {
 
   wire.on('close', function () {
     clearTimeout(timeoutId)
+    wire.requests.forEach(function (req) {
+      self.storage.cancelBlock(req.piece, req.offset)
+    })
   })
 
   wire.on('choke', function () {


feross:
@astro When a wire emits `finish`, all pending request callbacks are called with errors. Code here: https://github.com/feross/bittorrent-protocol/blob/c41d091549dc15261f81d3c22d6d59771a6b9cf1/index.js#L620-L621

This will be handled in `gotPiece` in torrent.js here: https://github.com/feross/webtorrent/blob/25c6a10e3180b4aeea1c36f8d57ad66a0e599b52/lib/torrent.js#L966 where `storage.cancelBlock` is called.

Unless I'm mistaken, this PR shouldn't be necessary! Do you agree?

feross:
> malicious peers may hold back responses indefinitely. What do you think about request timeouts?

Timeouts are already handled by `bittorrent-protocol` here: https://github.com/feross/bittorrent-protocol/blob/c41d091549dc15261f81d3c22d6d59771a6b9cf1/index.js#L450 as long as we call `wire.setTimeout` in torrent.js.

feross:
There's even a test verifying that timeouts work as expected here: https://github.com/feross/bittorrent-protocol/blob/master/test/timeout.js

astro:
you're right.

PR #267, opened 2015-02-15
Display reserved blocks in CLI

grunjol:
Green blocks in CLI cmd.js represent non-BLOCK_BLANK blocks, which includes reserved blocks. 

More info https://github.com/feross/webtorrent/issues/266 

This patch add blue marks for reserved blocks


feross:
Thanks, this will be useful for debugging.

Issue #266, opened 2015-02-15
Unverified pieces get stuck until all pieces finish downloading


adammw:
e.g. piece 59 in the screenshot below

![screen shot 2015-02-16 at 12 28 03 am](https://cloud.githubusercontent.com/assets/153219/6203374/c7432706-b572-11e4-87fa-2819f04129a5.png)

yipperr:
shady need to obfuscate the image

grunjol:
I think you are seeing that because the green represent non-BLOCK_BLANK blocks, which includes reserved blocks.

on cmd.js [line 461] (https://github.com/feross/webtorrent/blob/master/bin/cmd.js#L461) you can replace 
```
bar += piece.blocks[j] ? '{green:█}' : '{red:█}'
```
with 
```
bar += piece.blocks[j] ? (piece.blocks[j] === 1 ? '{blue:█}' : '{green:█}') : '{red:█}'
```
and you will see reserved blocks in blue.



But not all reserved blocks will be shown, because if you have reserved blocks on pieces without any written block, the condition on [line 456] (https://github.com/feross/webtorrent/blob/master/bin/cmd.js#L456) will skip it
```
if (piece.verified || piece.blocksWritten === 0) {
    continue
}
```
so, something like this can be used to skip the condition

```
if (piece.verified || !piece.blocks[0]) {
    continue
}
```
even if it doesn't check all the blocks, as far i can see when starting with a new piece it starts with the first block, so its a kind of better approach

adammw:
I'm not sure exactly what "reserved" means in this context, but yeah with that change I'm seeing that the pieces that get stuck do have 'blue' segments in them, I was really unfortunate with this one as it happened on the first block so I had to wait for the entire file to be downloaded before it would go back and try again.
![screen shot 2015-02-18 at 9 37 29 pm](https://cloud.githubusercontent.com/assets/153219/6245814/8886f1e6-b7b6-11e4-8906-b58c3eb98de7.png)

feross:
@grunjol's fix in https://github.com/feross/webtorrent/pull/272 should resolve this issue! Please give webtorrent@0.27.2 a try, and please let us know if this does the trick.

PR #272, opened 2015-02-18
Allow to cancel reserve on unwritten blocks

grunjol:
Trying to figure out why https://github.com/feross/webtorrent/issues/266 happens and after debugging using the https://github.com/feross/webtorrent/pull/267 PR, it seems like the cancelBlock is called as you mention here https://github.com/feross/webtorrent/pull/253 but if the **block is the first in the piece** and was never written, the 'reserve'  is never really canceled because the buffer is null (because is lazy initialized)

### DIFF
--- a/lib/storage.js
+++ b/lib/storage.js
@@ -104,7 +104,7 @@ Piece.prototype.reserveBlock = function (endGame) {
 
 Piece.prototype.cancelBlock = function (offset) {
   var self = this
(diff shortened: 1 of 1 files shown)


feross:
Good catch! I believe this fixes #266, which is a bug I've been trying to track down for a while. You're the bomb!

feross:
Released as webtorrent@0.27.2

PR #282, opened 2015-03-18
should remove part of memory leaks closures for callback in event listen...

arestov:
...ers in foreight objects

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -101,6 +101,8 @@ function Torrent (torrentId, client, opts) {
   this._servers = []
 
   if (torrentId !== null) this._onTorrentId(torrentId)
+
+  this._destroyers = []
 }
 
 Object.defineProperty(Torrent.prototype, 'timeRemaining', {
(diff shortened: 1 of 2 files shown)


feross:
Thanks for the PR. How did you measure that these changes actually reduce memory usage? How can I confirm on my machine?

arestov:
I did not make any mesures for this. 

This case https://cloud.githubusercontent.com/assets/69734/5448999/74155c90-8505-11e4-9dd4-afbf7cb99ce6.png revealed to me that code has not any unsubscribtion for events.

But them should be because it common case for memory leaks
https://msdn.microsoft.com/en-us/library/ie/dn741342(v=vs.94).aspx
(dev placed some object in closure, function with closure are placed in 
_events = {
  eventName: [func]
} cache. Thats all - now you have to remove this reference chain if you want to remove object from memory

So this should fix not only part of confirmed leaks, but also just potential leaks

feross:
Why did you close the issue?

arestov:
I commited from fork master where I plan to make more changes. (so i did) I want to make new pull request from other branch (wich I still did not make)

arestov:
https://github.com/feross/webtorrent/pull/382

PR #293, opened 2015-03-25
Add comment in browser-download example about explicit seeding

fpinzn:
A little clarification never hurt nobody :)

This would fix the need to explain this https://github.com/feross/webtorrent/issues/292 multiple times.


fbodz:
This is long overdue. Good idea.

feross:
Thanks @fpinzn!

PR #312, opened 2015-05-11
Added an app built with Webtorrent

alexeisavca:
Can I haz my app mentioned here pls?


feross:
Wow! This is amazing!! Reminds me of [PeerCDN](http://www.peercdn.com). Good work :-)

feross:
Just realized the name "webtorrentapp" might be confusing since we're going to be releasing [WebTorrent.app](https://github.com/feross/webtorrent.app) soon.

alexeisavca:
I guess I can think of a new name.
tbh "webtorrenapp" just was the first thing that popped into my mind.

feross:
Not a huge deal, but just wanted to let you know :cat2:

PR #335, opened 2015-05-31
Display info hash on commandline for seed option

### LINKED ISSUE #330: Command line: Make it easier to copy the magnet link / infohash to clipboard
`webtorrent seed` doesn't show the magnet link / infohash in a way that's easy to copy-paste. Let's fix that. This is a good first bug for a new contributor :)

jcytong:
Fix webtorrent seed doesn't show info hash

Fixes #330


feross:
Thanks for the PR! Released as 0.48.4.

PR #336, opened 2015-06-01
fix server.js browser exclude

mvayngrib:


feross:
Good catch - thanks!

feross:
Released as 0.48.3.

PR #351, opened 2015-06-07
error handling for duplicate torrents

devTristan:
I'm mostly happy with this solution.

I'd be even happier if I could detect the error message by a safer means than message string comparison, but I didn't want to over-complicate.

Tests pass.

### DIFF
--- a/index.js
+++ b/index.js
@@ -168,8 +168,17 @@ WebTorrent.prototype.download = function (torrentId, opts, ontorrent) {
   if (ontorrent) self.on('torrent', clientOnTorrent)
 
   torrent.on('error', function (err) {
-    self.emit('error', err, torrent)
-    self.remove(torrent)
+    var alreadyExistsMessage = 'There is already a swarm with info hash'
+    if (ontorrent && err.message && err.message.substr(0, alreadyExistsMessage.length) === alreadyExistsMessage) {
+      self.torrents.splice(self.torrents.indexOf(torrent), 1)
+      torrent = self.torrents.filter(function (_torrent) {
+        return torrent.infoHash === _torrent.infoHash
+      })[0]
+      process.nextTick(clientOnTorrent.bind(null, torrent))
+    } else {
+      self.emit('error', err, torrent)
+      self.remove(torrent)
+    }
   })
 
   torrent.on('listening', function (port) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -168,8 +168,17 @@ WebTorrent.prototype.download = function (torrentId, opts, ontorrent) {
   if (ontorrent) self.on('torrent', clientOnTorrent)
 
   torrent.on('error', function (err) {
-    self.emit('error', err, torrent)
-    self.remove(torrent)
+    var alreadyExistsMessage = 'There is already a swarm with info hash'
+    if (ontorrent && err.message && err.message.substr(0, alreadyExistsMessage.length) === alreadyExistsMessage) {
```
devTristan: Is it okay for this line to be this long?

index.js
```diff
@@ -168,8 +168,17 @@ WebTorrent.prototype.download = function (torrentId, opts, ontorrent) {
...
+      self.torrents.splice(self.torrents.indexOf(torrent), 1)
```
devTristan: can't call self.remove because it does other stuff like destroying the torrent

index.js
```diff
@@ -168,8 +168,17 @@ WebTorrent.prototype.download = function (torrentId, opts, ontorrent) {
...
+      torrent = self.torrents.filter(function (_torrent) {
+        return torrent.infoHash === _torrent.infoHash
+      })[0]
+      process.nextTick(clientOnTorrent.bind(null, torrent))
```
devTristan: calling this immediately causes [this test](https://github.com/devTristan/webtorrent/blob/duplicate-handling/test/duplicates.js#L23) to fail

devTristan:
Just noticing - running `client.seed` and then later `client.add` causes the torrent to stop seeding. This is beyond my ken. @feross any ideas?

feross:
I think we can solve this more easily by just checking if `client.get(torrentId)` is truthy before we go forth with creating a new torrent. I don't like the idea of searching the error string.

I used your test cases (thanks!) and solved this using `client.get`. Released as 0.49.0. Thanks!

devTristan:
Great. I wanted to use client.get for this but couldn't quite figure out how. I'm glad you could.

PR #358, opened 2015-06-13
use input file/dir path when seeding

### LINKED ISSUE #357: When seeding a file/folder from disk, don't perform extra copy to /tmp
Hi !

If I want to seed a file from (just an example) */home/me/videos/awesome.webm* I see that webtorrent copy it as */tmp/webtorrent/xxx/xxx.webm*. I wanted to know if this was really needed, since the file is already on the disk. 

This could be annoying with big videos (1Go), disk space would rapidly decrease and the seeding would take much time since the disk should copy 1Go before it.

Thanks :)

janza:
closes #357

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ var parallel = require('run-parallel')
 var parseTorrent = require('parse-torrent')
 var speedometer = require('speedometer')
 var zeroFill = require('zero-fill')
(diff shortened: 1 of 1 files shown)


feross:
Released as 0.50.1. Thanks for the PR!

PR #363, opened 2015-06-25
extension support

devTristan:
I made this thing support extensions. Thoughts?


feross:
Yes, this is good idea. Without this, the only way to use a custom protocol extension is to send a PR to webtorrent core, or to use `bittorrent-protocol` directly.

However, I'd rather emit `'wire'` as an event than pass it in as an option to the `WebTorrent` constructor, because we currently don't have any `on*` style options – those are all events. I'll make this change to your PR.

Eventually, we might want to add an `extensions` option to the constructor to make it easier than: 

```js
var MyExtension = require('./my-extension')

torrent.on('wire', function (wire) {
  wire.use(MyExtension)
})
```

feross:
Released as 0.50.0.

devTristan:
:+1:

I wasn't super happy with the onWire option. I was thinking maybe `client.use(Extension)` but this works too.

PR #373, opened 2015-07-15
fixed #322 blobURL cleanup

### LINKED ISSUE #322: Track and cleanup Blob object urls that are given out
There are two place where we give out Blob object URLs:

- `file.getBlobURL`
- `torrent.torrentFileURL`

OlaviSau:
Couldn't find tests to go along with the changes


feross:
Nice - looks good!

feross:
Released as 0.52.1.

Issue #316, opened 2015-05-13
Event 'done' isn't called when there are deselected files


sudoaza:
This checks for all the files to be downloaded instead it should check only for the "selected" files/pieces.
https://github.com/feross/webtorrent/blob/master/lib/storage.js#L597

So when all the wanted files are downloaded the torrent wont be done if there are deselected files.

feross:
Released as 0.65.1.

PR #378, opened 2015-07-16
issue #316 now checks only for selected files

OlaviSau:


DiegoRBaquero:
As this PR is outdated, I put up a new one: https://github.com/feross/webtorrent/pull/528.

All credits to you.

PR #379, opened 2015-07-16
Fix appendTo Readme

Chocobozzz:
It seems that we must specify a callback


feross:
Thanks for fixing the readme. I will also make it so the callback is not required.

PR #382, opened 2015-07-19
should remove part of memory leaks 

arestov:
memory leaks through closures of callback in event listeners in foreight objects.

`destroy` should unsubscribe current object from any events in foreight objects

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -29,11 +29,15 @@ var sha1 = require('simple-sha1')
 var uniq = require('uniq')
 var ut_metadata = require('ut_metadata')
 var ut_pex = require('ut_pex') // browser exclude
+var subscribe = require('subscribe-ev')
 
 var File = require('./file')
 var RarityMap = require('./rarity-map')
 var Server = require('./server') // browser exclude
 var Swarm = require('./swarm')
+
+var initSubscribe = subscribe.init
+var toDestory = subscribe.toDestory
 
 var MAX_BLOCK_LENGTH = 128 * 1024
 var PIECE_TIMEOUT = 30000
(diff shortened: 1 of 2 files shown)


arestov:
also `self.swarm.listen(` here can be problem

rom1504:
I think it would be interesting to have some way to measure how much the memory leaks decreased.

arestov:
You can use this trick.
https://github.com/feross/webtorrent/issues/712#issuecomment-203593883

Add this in every file and Class (instances of which could be destroyed) in every webtorrent module. 
That run webtorrent for few hours. Or just add 10 torrent and then remove them all.
Take snapshot and you will see how many objects should be GCed by counting of MemoryLeak (filter by name)
```javascript
var MemoryLeak = function(){}


Item.die = function() {
  // ...
  this.leak = new MemoryLeak()
  // ...
}
```
I guess it can be added in master branch with `process.NODE_ENV == 'DEBUG'`

```javascript
var isDebugging = process.NODE_ENV == 'DEBUG';

//...

Item.die = function() {
  // ...
  if (isDebugging) {
    this.leak = new MemoryLeak()
  }
  // ...
}
```

(If your code has no leaks than GC will remove whole `item`, but if not — you will see `MemoryLeak` in memory profiler (snapshot))

arestov:
We, I've put `MemoryLeak` in some of files
leak_detect

Started webtorrent-desktop (whithout code in commit) download and got some leaks
leaks1

Click start/stop and got more leaks
leaks2

So looks like this commit https://github.com/feross/webtorrent/pull/382/commits/44e486b5227f91f3eb4c2ee854e0e28b7bcf5818 is not enough


I guess that this code is also problem:
https://github.com/feross/torrent-discovery/blob/master/index.js#L76 
https://github.com/feross/torrent-discovery/blob/master/index.js#L58

PR #385, opened 2015-07-23
Some documentation for _onMetadata

OlaviSau:
Writing documentation to understand webtorrent ^^

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1,18 +1,3 @@
-Skip to content
-This repository  
-Search
-Pull requests
-Issues
-Gist
- @OlaviSau
- Watch 395
-  Unstar 4,992
-  Fork 332
-feross/webtorrent
- branch: master  webtorrent/lib/torrent.js
-@ferossfeross 7 days ago torrent.destroy() remove from client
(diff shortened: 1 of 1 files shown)


OlaviSau:
All suggestions open

feross:
Thanks for sending this in. This PR has a few problems:

1. The tests don't pass. Run `npm test` to make sure your changes don't break any of the tests.
2. I prefer succinct inline comments. If you wanted to move some of what you wrote into those, that would be fine.
3. You also changed some code, so the PR name is misleading. I think this probably introduces a bug.

OlaviSau:
Thanks for responding :)
Sorry about the tests, pushed it at like 2 am and forgot about them since i had copy pasted from github , appareantly it put some github gibberish into it ^^ last test only required newlines at the end.
About the comments, yeah whatever you like, I wrote it to understand webtorrent better, so perhaps I can contribute in some larger things rather than minor bugfixes :) 
And about the code change, the PR is name is misleading indeed sorry about that, but the code change itself is valid, you don't need to check if the object exists if you are checking a property on it.

The following values are always falsy:

false
0 (zero)
"" (empty string)
null
undefined
NaN (a special Number value meaning Not-a-Number!)

PR #389, opened 2015-07-29
Added command line hook scripts(done/exit).

solderzzc:
Give a chance for the programs use webtorrent cli to run scripts when download done or exit.

For example:
exit.sh

#!/bin/bash
echo "exiting" > info.txt
echo "Torrent filename  :  $1" >> info.txt
echo "Torrent path      :  $2" >> info.txt
echo "Torrent magentURI :  $3" >> info.txt

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -102,7 +102,7 @@ function checkPermission (filename) {
       errorAndExit('Your script ' + filename + ' is not exist')
     }
     // check if the script has executable permission
-    if (!(1&parseInt((stats.mode & parseInt('777', 8)).toString(8)[0]))) {
+    if (!(1 & (parseInt((stats.mode & parseInt('777', 8)).toString(8)[0])))) {
       errorAndExit(filename + ' don\'t have executable permission')
     }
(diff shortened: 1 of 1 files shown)


feross:
Thanks for the PR! Released as 0.55.0.

I renamed `--exit` to `--on-exit` and `--done` to `--on-done` for clarity and removed the shorthands `-e` and `-d` since these won't be commonly used options. Also, I don't think `cmd.js` is the right place to be saving the .torrent file, so I removed that code for now.

Out of curiosity, what do you plan to use these hooks for?

solderzzc:
These hooks are designed for running on single board computer(such as
Raspberry Pi or some TV boxes with Linux running in container).
Such environment has limited resource. For 512M memory, it's not suitable
to run multiple webtorrent instances.
The --on-done script can be used to move/link/copy the downloaded files to
'Transmission' for seeding to save memory. Transmission can serve multiple
torrents easily.
And the --on-exit script can be used to clean up downloaded folder if
there's storage limitation(Same as popcornTime Android behavior).

PR #399, opened 2015-08-12
	Default download destination should be current directory

gillesdemey:
Fixes https://github.com/feross/webtorrent/issues/390

I'm not sure if this would be the best way to go about it, but I felt like it didn't belong in the `FSStorage` module.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -255,7 +255,7 @@ var client, href, playerName, server, serving
 
 function runDownload (torrentId) {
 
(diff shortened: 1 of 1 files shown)


feross:
Thanks for the PR! Released as 0.57.0.

I also added you as a GitHub collaborator so you can [contribute](https://github.com/feross/webtorrent/blob/master/CONTRIBUTING.md) as you see fit. :+1: Nice work!

PR #402, opened 2015-08-16
Respect the "private" flag in .torrent files

### LINKED ISSUE #271: Respect the "private" flag in .torrent files
When torrents are marked "private", they should not announce to any source but the tracker (i.e. no DHT, no peer exchange).

gillesdemey:
Fixes #271 

Couple of caveats:

* The DHT client still throws the ready event, even though no-one is listening for it. This could be patched in `bittorrent-dht` to check if the server hasn't been destroyed yet before we emit the event.

* No tests to check if PEX is correctly disabled.


lib/torrent.js
```diff
@@ -207,6 +207,12 @@ Torrent.prototype._onParsedTorrent = function (parsedTorrent) {
     })
   }
 
+  // destroy DHT client if the torrent is private
+  if (self.parsedTorrent.private && self.client.dht) {
+    self.client.dht.destroy()
+    self.client.dht = null
+  }
```
feross: Other torrents might not be private, and we want the DHT to be available to them. Best to solve this a different way that preserves the DHT for use by other torrents.

gillesdemey: Oops, yeah that doesn't seem right.

Then disabling DHT for the discovery and setting DHT to false when passing the handshake options to bittorrent-protocol should suffice?

I'll take another stab at this.

feross:
Excellent PR!

Unfortunately, lots of WebTorrent code changed recently as part of efforts to make WebTorrent and torrent-stream/peerflix share more code in common. Would you mind rebasing these changes so they merge cleanly?

gillesdemey:
I'll take another look, I had noticed that `parsedTorrent` had been removed.

feross:
Right. Instead of doing `torrent.parsedTorrent`, all those properties now exist directly on the `torrent` object.

gillesdemey:
Rebased and updated to reflect the changes in master

feross:
Nice work! Released as 0.62.0.

I also added you as a GitHub collaborator. Feel free to contribute as you see fit :)

feross:
Scratch that, hehe. I already added you from a previous PR :+1:

PR #417, opened 2015-09-02
Switching to mafinstosh/chromecasts

### LINKED ISSUE #414: Replace "chromecast-js" package with "chromecasts"
@mafintosh's package is probably better and I like the idea of reusing code with playback/peerflix.

gillesdemey:
Fixes #414, tested on a Chromecast with the latest firmware.


feross:
Nice work! Released as 0.61.0.

PR #425, opened 2015-09-10
[Feature proposition] Allow to render a torrent into an existing element

valeriangalliat:
Like `file.appendTo`, but taking an existing element as parameter, for example:

```html
<img src="magnet:?xt=urn:btih:d2346c943f770eba4994e25fa7b478692bcf0064&amp;dn=test.png">
```

```js
const WebTorrent = require('webtorrent')

Array.from(document.querySelectorAll('[src^=magnet:]')).forEach(elem =>
  client.add(elem.src, torrent =>
    torrent.files[0].renderTo(elem)
  )
)
```

See the [`Implement 'render-to' method`](https://github.com/feross/webtorrent/commit/41dc75b8e7420297043d2fb294ce334a498498df) commit to have a proper diff, because `append-to.js` was renamed to `render.js` and it's messing the full PR diff.

Related: #531.


valeriangalliat:
Any comment on this feature?

devTristan:
I like it. I've wanted this myself.

RomanEmelyanov:
I'm work under https://github.com/feross/webtorrent/issues/531

render-to method does not include yet to main branch :(

Do you have the patch as standalone javascript file to include as the second after:
https://cdn.jsdelivr.net/webtorrent/latest/webtorrent.min.js
?

valeriangalliat:
@feross I rebased the PR, ready to merge in master! I'd like a review and approval of other WebTorrent contributors before merging, do you know who to ping? Thanks. :)

lib/handle.js
```diff
@@ -0,0 +1,150 @@
...
+    file.getBlobURL(function (err, url) {
+      if (err) return fatalError(err)
+      elem.src = url
+      if (extname !== '.pdf') elem.sandbox = 'allow-forms allow-scripts'
+      cb(null, elem)
+    })
+  }
+
+  function fatalError (err) {
+    err.message = 'Error appending file "' + file.name + '" to DOM: ' + err.message
```
devTristan: This won't be accurate anymore, because "appending"

devTristan:
It looks pretty good to me, but if you don't mind a little unsolicited feedback, I'd call `handle.js` something different, maybe `render.js`, just to make it easier for newcomers to know where to look when exploring the source tree.

valeriangalliat:
@devTristan Thanks for your comments, I fixed the error message and renamed `handle.js` to `render.js`, it indeed makes more sense.

lib/render.js
```diff
@@ -0,0 +1,150 @@
...
+    })
+  }
+
+  function renderImage () {
+    elem = getElem('img')
+    file.getBlobURL(function (err, url) {
+      if (err) return fatalError(err)
+      elem.src = url
+      elem.alt = file.name
+      cb(null)
```
feross: This shouldn't be here

lib/render.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  function renderIframe () {
+    elem = getElem('iframe')
+
+    file.getBlobURL(function (err, url) {
+      if (err) return fatalError(err)
+      elem.src = url
+      if (extname !== '.pdf') elem.sandbox = 'allow-forms allow-scripts'
+      cb(null, elem)
```
feross: This shouldn't be here

lib/render.js
feross: Nevermind

feross: This should be `cb(null, elem)`

lib/append-to.js
```diff
@@ -28,141 +8,28 @@ module.exports = function appendTo (file, rootElem, cb) {
...
-    rootElem.appendChild(elem)
-    file.getBlobURL(function (err, url) {
-      if (err) return fatalError(err)
-      elem.addEventListener('error', fatalError)
-      elem.addEventListener('playing', onPlaying)
-      elem.src = url
-      elem.play()
-    })
+    elem.autoplay = true // for chrome
+    elem.play() // for firefox
```
feross: `createMedia` needs to append to the DOM too. Right now, only `createElem` does.

feross:
@valeriangalliat Thanks for the good PR! It has a few issues (mentioned as inline comments), but I can fix those up after merge.

You rock!

feross:
Released as 0.65.0.

valeriangalliat:
Awesome, thanks for fixing the issues!

feross:
No problem. Thanks for taking the initiative to fix the issue :) I really appreciate that.

RomanEmelyanov:
Thanks a lot! Any plans to add "renderTo" method to "https://cdn.jsdelivr.net/webtorrent/latest/webtorrent.min.js" ?

feross:
@RomanEmelyanov The new code should be available at that url. You might need to clear your cache to see it.

RomanEmelyanov:
What I'm do wrong?
Tested on Google Chrome.

```
<html><head>
<script src="https://cdn.jsdelivr.net/webtorrent/latest/webtorrent.min.js"></script>
</head><body><script>
var client = new WebTorrent()
var torrentId = 'magnet:?xt=urn:btih:6a9759bffd5c0af65319979fb7832189f4f3c35d'
client.add(torrentId, function (torrent) {
var file = torrent.files[0]
file.renderTo('#V1', function (err, elem) { // does not work :(
// file.appendTo('#V2', function (err, elem) { // works fine
 if (err) throw err
 console.log('New DOM node with the content', elem)
})

})
</script>
<div id="V2"></div>
<video id="V1"></video>
</body></html>

**Comment by RomanEmelyanov** [comment]
SOLVED: I'm forget to add "controls" and "autoplay" for <video id="V1"> tag, sorry to trouble

**Comment by nextgenthemes** [comment]
So if I put controls and autoplay on a element and you `renderTo` its teh same as if I use `opts` on `renderTo`.

Also what does it do when the torrent file is not a video and I set it to render to a video tag?

**Comment by feross** [comment]
> So if I put controls and autoplay on a element and you renderTo its the same as if I use opts on renderTo.

The default is `true`, so you don't need to set these options. I would rely on webtorrent to autoplay for you, since there are some browser differences. On Firefox, you have to explicitly call `video.play()` after the metadata has been added to the tag.

> Also what does it do when the torrent file is not a video and I set it to render to a video tag?

It'll try to put the file data into the video tag which will just make the video tag fire an `'error'` event.

**Comment by nextgenthemes** [comment]
@feross Thanks. But I need to disable autoplay and I was thinking it would be convenient to use `renderTo` in a video that my php code already generates that also has a class that I wanted to use. Now I am thinking this entire `renderTo` thing is probably a not so good idea and just complicates things.

Between the lines you answered my question with "No". So the `renderTo` is not smart to read the video tag attributes as opts and then doing kind of the same thing as `appendTo`. I feel more comfortable with the `appendTo` and opts I will just read from data-attributes now.

**Comment by feross** [comment]
You can disable autoplay by passing an `{ autoplay: false }` option to either method.

**Comment by nextgenthemes** [comment]
I know and I do this now.

```
			// Stream the file in the browser
			torrent.files[0].appendTo( $container, {
				autoplay: $wrapper.hasAttribute('data-arve-autoplay'),
				controls: $wrapper.hasAttribute('data-arve-controls')
			} );
```

PR #426, opened 2015-09-10
Always pass 'elem' to 'append-to' callback

valeriangalliat:
Looks like it was forgotten for 'img' and 'iframe'.


feross:
Thanks for the PR! Released as 0.62.1.

Since you've submitted a few good PRs at this point, I've added you as a GitHub collaborator. Feel free to contribute as you see fit. :+1:

PR #427, opened 2015-09-10
Add an 'announce' option to announce to custom trackers

valeriangalliat:
```sh
webtorrent seed somefile -a http://localhost:8000/announce -a udp://localhost:8000 -q
```

Output:

```
magnet:?xt=urn:btih:066ed351804835cf0fec21e3b352f58e270a26d5&dn=somefile&tr=http%3A%2F%2Flocalhost%3A8000%2Fannounce&tr=udp%3A%2F%2Flocalhost%3A8000
```

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -43,6 +43,7 @@ var argv = minimist(process.argv.slice(2), {
     s: 'select',
     i: 'index',
     o: 'out',
+    a: 'announce',
     q: 'quiet',
     h: 'help',
     v: 'version'
@@ -63,6 +64,7 @@ var argv = minimist(process.argv.slice(2), {
   ],
   string: [ // options that are always strings
     'out',
+    'announce',
     'blocklist',
     'subtitles',
     'on-done',
@@ -201,6 +203,7 @@ Options (advanced):
     -p, --port [number]     change the http server port [default: 8000]
     -t, --subtitles [path]  load subtitles file
     -b, --blocklist [path]  load blocklist file/http url
+    -a, --announce [url]    tracker URL to announce to
     -q, --quiet             don't show UI on stdout
     --on-done [script]      run script after torrent download is done
     --on-exit [script]      run script before program exit
@@ -459,7 +462,7 @@ function runSeed (input) {
   })
   .on('error', fatalError)
 
-  client.seed(input)
+  client.seed(input, { announce: argv.announce })
 
   client.on('torrent', function (torrent) {
     if (argv.quiet) console.log(torrent.magnetURI)


valeriangalliat:
Hey, I would really appreciate a review or any opinion before merging!

yciabaud:
I believed [magnet-uri](https://github.com/feross/magnet-uri) was converting tr attributes (like "tr=http%3A%2F%2Flocalhost%3A8000%2Fannounce&tr=udp%3A%2F%2Flocalhost%3A8000") to announce URL list on its own. Am I wrong or is it a convenience argument to achieve the same goal?

valeriangalliat:
@yciabaud I'm not sure to understand, how would you seed a file from the filesystem with custom announce URLs then?

yciabaud:
Oh I may have misunderstood your example. I believed the magnet uri was an argument. 

I understand now the purpose of this evolution and I think it is very useful to anyone running a custom tracker.

The PR looks good to me.

valeriangalliat:
I updated the PR comment to make it more clear, indeed the magnet URI wasn't part of the command.

That's effectively while running a custom tracker I encountered this issue and made the PR, I had to use the JS API to announce on my tracker otherwise.

Thanks for reviewing @yciabaud.

Anything needed before mergin the PR? Like docs somewhere maybe, but I don't see other options documented in the readme except for media output options?

cc @feross

gillesdemey:
Out of curiosity, will this overwrite the existing trackers listed in the torrent file / magnet URI? This may be undesirable if you simple wish to add a tracker on top of the existing ones.

valeriangalliat:
@gillesdemey Hmm, I didn't consider this before, the idea was just to allow to configure the `announce` option, whatever this option is doing.

I didn't tried but according to the [documentation](https://github.com/feross/webtorrent/blob/d7a88f3ebe314fd2db3507898040b0d5f897183d/README.md#clientaddtorrentid-opts-function-ontorrent-torrent-), it is "added to list in .torrent or magnet uri". :smile_cat:

DiegoRBaquero:
It does overwrite the default trackers. But that's the way create-torrent works atm. If you specify any tracker, at least one, the default are overwritten. https://github.com/feross/create-torrent/blob/master/index.js#L279-L281

We should allow two ways. "announce" and "announceList", as they are already used somehow but both replace. If you use announceList, it overwrites. If you use "announce" it concats them.

What do you think?

valeriangalliat:
@DiegoRBaquero, I agree it would be nice to have a way to add trackers to the default list instead of overwriting, but in this PR I just want the CLI options to reflect the JS API (so overwriting defaults in any case).

If `create-torrent` is updated to support a way to append trackers, I'd happily reflect the changes in the CLI options, but modifying `create-torrent` to support this probably involves quite some discussions and reflection on the right API, and probably better have its place in another PR. :)

feross:
Thanks everyone for the thoughtful discussion. Sorry for taking so long to review this PR. I've been traveling.

The CLI change looks good to me, but we should be consistent and pass the `announce` option into other commands that support an `announce` option, like `webtorrent create` and `webtorrent download`. I'll make this change after merging.

(FWIW, any collaborator should feel free to merge a PR once there's sufficient support from other collaborators and it's been at least 3 days so everyone has a chance to weigh in with their thoughts. Power to the people :+1:)

feross:
Released as 0.64.0.

Issue #404, opened 2015-08-17
webtorrent seed command should show magnet link


browny:
I install the webtorrent cli by `npm install webtorrent -g`

Open one terminal to run `webtorrent seed /data`, output as below

    seeding: img
    info hash: 75fa47b07d23f24a77ab7f01f5c8ff51fb38d8c9
    speed: 0 B/s  downloaded: 0 B/193.59 kB  uploaded: 0 B  peers: 0/0  hotswaps: 0
    time remaining: Infinity years remaining  total time: 9s  queued peers: 2  blocked: 0

And open another terminal to download that seed (`webtorrent download 75fa47b07d23f24a77ab7f01f5c8ff51fb38d8c9`), but nothing happen, the log just shows 

    fetching torrent metadata from 0 peers

Anything I am doing wrong?

joshuaterrill:
It appears that the seeding hasn't finished yet. To my knowledge, you'll need to wait until the seeding has finished before you can start downloading. Unfortunately it appears that you have a long time to wait: ```time remaining: Infinity years remaining``` Maybe I'm understanding this wrong, though?

feross:
@browny Since you're seeding a new torrent and you didn't specify any trackers, some default ones are added for you automatically. When downloading from the second client, you'll want to use the full magnet link or .torrent file which contains the same list of trackers.

Just using the info hash makes it less likely that the new client will be able to find the seeder. Without trackers, the DHT is used. But if UDP tracker is filtered or blocked on your network, then no peers will be found.

Right now, there's not an easy way to copy the magnet link from the `webtorrent seed` command. If we add that feature, I think your example will work.

feross:
This is fixed now. Just pass `-q` to the command.

`webtorrent seed files -q`

PR #439, opened 2015-09-21
Show torrent magnetURI when seeding

Stiveknx:
From issue #404.


whitef0x0:
Is there any reason this wasn't merged @feross?

feross:
Yeah, sorry for not merging this earlier. This looks like it doesn't merge cleanly anymore.

The other issue with it (and it's not the PR's fault) is that it'll break the `clivas` package because magnet uris will almost always be longer than the terminal width and the package can't handle rendering that.

I think a better solution is to pause with the uri visible on the screen with a "Press ENTER to start seeding." message. This gives the user the chance to copy the uri before the terminal GUI takes over.

PR #447, opened 2015-09-28
VerifyPieces: Removed Parallel execution to prevent excessive memory consumption...

ngjermundshaug:
…and also to prevent UI/Process to hang when verifying large files.

Also introduced a verifyingPieces flag on Torrent - in order to let consumers of the library detect when files are being verified.
self._onStore() is now being invoked before Torrent is verified - so that consumers can poll on the verifyingPieces flag.

UI Demo: http://screencast.com/t/NgbUPvQLik1O

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -80,7 +80,7 @@ function Torrent (torrentId, opts) {
   self.numBlockedPeers = 0
   self.files = null
   self.done = false
-  self.verifyingPieces = false;
+  self.verifyingPieces = false
 
   self._amInterested = false
   self._selections = []
@@ -358,9 +358,8 @@ Torrent.prototype._onMetadata = function (metadata) {
 
     self._onWireWithMetadata(wire)
   })
-    
-  
-  //verify pieces
+
+  // Verify pieces
   var verifyPiece = function (index) {
     self.store.get(index, function (err, buf) {
       if (!err) {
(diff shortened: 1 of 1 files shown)


joseph-onsip:
It looks like this change causes the travis builds to stall, and they also stall on my local machine. I know you said [the tests won't run on Windows](https://github.com/feross/webtorrent/issues/454#issuecomment-145978233), but you could try rebasing your changes onto `master` (or merging `master` into your branch) and then testing locally? There were some [recent changes](https://github.com/feross/webtorrent/pull/450) that aim to allow the tests to run on Windows.

IngoValente:
Thanks for this, indeed keeps my UI usable while verifying.

However the sha1 hash check is async, so shouldn't the 'verify next piece check' be inside the sha1 callback function? I got stuck using your code because it would mark verifying as done before the last piece was actually verified, and that fixed it for me.

feross:
@ngjermundshaug @josephfrazier @IngoValente Thanks all, for your contributions to this PR!

The simplest – and fastest – solution is to limit the number of concurrent verifications to one per CPU core so we go as fast as possible, but no faster.

Here's what I decided to go with: https://github.com/feross/webtorrent/commit/664eb30006216dfd6c6a62c78c74c16f3e5f8809

feross:
Released as 0.72.1.

PR #449, opened 2015-10-04
Use path.resolve in tests

joseph-onsip:
See https://github.com/feross/webtorrent/issues/429#issuecomment-139878312


feross:
Closing in favor of #450

Issue #429, opened 2015-09-11
Tests don't work on Windows


feross:
Got this email from a user:

```
Hi feross,

 

I have been looking at the WebTorrent library and I am interested in being a part of it. However I am having some issues getting set up.

This is on windows. Once I have clone the repo and I try npm install it runs but it doesn’t install everything.

I think ran a loop to loop thru the sub directories and run npm install on them.

I got that to work but when running npm test it doesn’t seem to be working correctly.

 

Output

C:\Users\Shavaunh\Documents\webtorrent>npm test

 

> webtorrent@0.62.0 test C:\Users\Shavaunh\Documents\webtorrent

> standard && node ./bin/test.js

 

events.js:85

      throw er; // Unhandled 'error' event

            ^

Error: spawn npm ENOENT

    at exports._errnoException (util.js:746:11)

    at Process.ChildProcess._handle.onexit (child_process.js:1053:32)

    at child_process.js:1144:20

    at process._tickCallback (node.js:355:11)

    at Function.Module.runMain (module.js:503:11)

    at startup (node.js:129:16)

    at node.js:814:3

npm ERR! Test failed.  See above for more details.

Any help would be lovely

Thanks
```

Without doing any investigation, my guess is that tests don't work on Windows because we're using `/` instead of `path.join` all over the tests.

shavyg2:
So i found the problem and solved it. Just needed to use path.resolve and also do child.spawn("node"...[CMD,....,...]

This is because a windows system doesn't use the shebang so it might not know how to execute a js file. Spawning using node and node will understand how to read this.

Small issue silly as it sounds, is i don't know how to send a pull request.

joseph-onsip:
My curiosity got the best of me and I tried this out in a [Window 10 VM](https://dev.modern.ie/tools/vms/mac/):

<img width="553" alt="screen shot 2015-10-06 at 10 17 19 pm" src="https://cloud.githubusercontent.com/assets/6473925/10327656/1f163b6e-6c78-11e5-8c4b-0a9dfb15d6b6.png">

It turns out this is a [long-standing issue](https://github.com/nodejs/node-v0.x-archive/issues/2318), but [`cross-spawn-async`](https://www.npmjs.com/package/cross-spawn-async) provides a workaround. I'll be submitting a PR shortly :)

PR #450, opened 2015-10-04
exec/spawn cmd.js with 'node'

joseph-onsip:
See https://github.com/feross/webtorrent/issues/429#issuecomment-139878312

> This is because a windows system doesn't use the shebang so it might
> not know how to execute a js file. Spawning using node and node will
> understand how to read this.

(includes and resolves #449)

### DIFF
--- a/test/basic-node.js
+++ b/test/basic-node.js
@@ -1,16 +1,17 @@
 var WebTorrent = require('../')
 var fs = require('fs')
 var http = require('http')
+var path = require('path')
 var parseTorrent = require('parse-torrent')
 var test = require('tape')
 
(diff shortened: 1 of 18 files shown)


feross:
Thanks – great work!

joseph-onsip:
Sure thing! Unfortunately, I'm not certain that #429 is completely fixed, as I don't have a Windows machine to test on. Perhaps @shavyg2 could verify?

feross:
Okay, let's leave #429 open until we receive confirmation that it works.

@joseph-onsip Also, I've added you as a collaborator to WebTorrent so you can contribute as you see fit. :)

joseph-onsip:
Sounds good, cheers!

PR #451, opened 2015-10-04
travis: cache node_modules for faster builds

joseph-onsip:
This makes builds around 90 seconds faster. See here for an example: https://travis-ci.org/joseph-onsip/webtorrent/builds/83604415

This command finished after 5.75 seconds.

http://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/


feross:
Sure, we can try this.

PR #455, opened 2015-10-07
tests use cross-spawn-async instead of child_process.spawn

joseph-onsip:
This allows Windows to run the tests, as promised in
https://github.com/feross/webtorrent/issues/429#issuecomment-146061289

Note that there are a couple of outstanding failures on Windows:

    # client.seed: filesystem path to folder with one file, string
    not ok 8 should be equal
      ---
        operator: equal
        expected: '3a686c32404af0a66913dd5f8d2b40673f8d4490'
        actual:   'e6887b78d89d995876c7cef3476b6fb32c4c4e3d'
      ...
    not ok 9 should be equal
      ---
        operator: equal
        expected: |-
          'magnet:?xt=urn:btih:3a686c32404af0a66913dd5f8d2b40673f8d4490&dn=folder&tr=udp%3A%2F%2Ftracker.webtorrent.io%3A80'
        actual: |-
          'magnet:?xt=urn:btih:e6887b78d89d995876c7cef3476b6fb32c4c4e3d&dn=folder&tr=udp%3A%2F%2Ftracker.webtorrent.io%3A80'

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.txt text eol=lf
--- a/bin/test.js
+++ b/bin/test.js
@@ -1,14 +1,14 @@
 #!/usr/bin/env node
 
-var cp = require('child_process')
+var spawn = require('cross-spawn-async')
 
 var runBrowserTests = !process.env.TRAVIS_PULL_REQUEST ||
   process.env.TRAVIS_PULL_REQUEST === 'false'
 
-var node = cp.spawn('npm', ['run', 'test-node'], { stdio: 'inherit' })
+var node = spawn('npm', ['run', 'test-node'], { stdio: 'inherit' })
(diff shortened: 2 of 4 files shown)


joseph-onsip:
> Note that there are a couple of outstanding failures on Windows:

Hmm, this appears to have been due to a bad [`core.autocrlf`](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#coreautocrlf-7kTgIgFwI4) setting I had in the VM. `core.autocrlf` was set to `true`, so `./test/content/folder/file.txt` had an extra CR character in it, causing the `infoHash` to differ. Running `git config core.autocrlf input` and refreshing `file.txt` allowed the tests to pass. See here for more info: https://help.github.com/articles/dealing-with-line-endings/#platform-linux

Issue #411, opened 2015-08-30
Pick an available port instead of crashing with "listen EADDRINUSE"


markhilton:
DEBUG INFO: webtorrent 0.60.1, node v0.12.7, darwin x64, exit 1
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: listen EADDRINUSE
    at exports._errnoException (util.js:746:11)
    at Server._listen2 (net.js:1156:14)
    at listen (net.js:1182:10)
    at Server.listen (net.js:1267:5)
    at runDownload (/opt/local/lib/node_modules/webtorrent/bin/cmd.js:313:10)
    at Object.<anonymous> (/opt/local/lib/node_modules/webtorrent/bin/cmd.js:147:3)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)

feross:
This error means that the port WebTorrent wants to use (port 8000) is already in use. This could be from an other instance of WebTorrent, or any another program.

You can run WebTorrent on a different port by using the `--port` argument:

```
webtorrent --port 8001
```

I'll leave this issue open since we can do better here -- by either providing a friendlier error, or preferably, automatically picking an available port.

markhilton:
Thanks!

feross:
Leaving this issue open until we support automatically picking an available port instead of crashing.

Stiveknx:
@feross I'm whiling to help on this.
It's ok to use an external package for that?
(I think it would be too much work doing this without).

https://github.com/baalexander/node-portscanner

gillesdemey:
@Stiveknx It's not too much work actually, it can be done in just a few lines of code.

https://github.com/gillesdemey/grop/blob/master/index.js

There are countless modules on npm already, though it looks like node-portscanner has a few more bells and whistles.

feross:
@Stiveknx Cool, happy to accept a PR from you that implements this. You might not need to use an open port finder, but instead pass `0` in as the port so the OS will automatically pick a port. 

@gillesdemey `grop` is a module done right – looks nice.

yciabaud:
I made a little contribution on this, I believe this is what you were talking about...
Let me know if there is anything.

PR #445, opened 2015-09-27
#411 Pick an available port on "listen EADDRINUSE" or "listen EACCES"

yciabaud:
Let the OS choose a port when the default or given one is used or unauthorized.
 
Fixes: #411

PR #462, opened 2015-10-14
#411 Choose random port when port not specified

Krizz:
Webtorrent will choose random unused port at start. Still throws an error if port is specified and in use.


joseph-onsip:
In my opinion, #445 does a better job of fixing #411 in the spirit of https://github.com/feross/webtorrent/issues/411#issuecomment-136194502

PR #485, opened 2015-11-19
clone.sh should use bash due to bashisms

ivucica:
Bashisms pushd/popd don't work so well if /bin/sh is dash, as is the case on a Debian machine.


feross:
Thanks for the fix!

PR #483, opened 2015-11-09
travis: test Node 4.2

joseph-onsip:

PR #494, opened 2015-11-27
refactored travis configuration

udnisap:
sudo is removed as it is the default configuration. 
added v4 and v5 for building environments
caching is removed as it might have problems when we change the dependency tree.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,10 @@ language: node_js
 node_js:
   - '4.2'
   - '0.12'
-sudo: false
+  - v4
+  - v5
+
 env:
   global:
   - secure: AJsEWFnWC5W8hcF3hJzm3PT7heazJpKg85xiSvIWVzLHZU/s0h4+WfJ6t0F9v3L4awaowm62vy8CRaxRkB4lJyJg+JK2K0QN7lNFGj2f8Jx2cFlVJ1IyY959GY4iUg66JrNj1yzS02+yQfweDngyifqzb7IlxnowiveDjUO2gyo=
   - secure: hvihwLUqlPchrGFXKWFF7iKRugISU7r/gLBo6O63nPeg0OwnYqYcC2BnBWoSiOdu9oR5bM4a5u0os04XL+bP3dqt324g0uBTqvyyxD6NhBsphVFkUmdUH3HMe7IQY6JTns96KT/6UkQapKhIuW4CUDeidR+5NFKvyRdKIjSawS4=
-cache:
-  directories:
-  - node_modules


rom1504:
"caching is removed as it might have problems when we change the dependency tree.", I doubt this is a problem https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/

udnisap:
Well the problem is say we upgrade a version of a dependency and since we are caching we will have both the libraries and may have problems when importing them as well when there is a peer dependency breaks a version of another library with a previous build.

rom1504:
> upgrade a version of a dependency and since we are caching we will have both the libraries

I'm not sure how travis caching work, but since they specifically put it as an example in that article up there, caching is probably smarter than that. I bet if you change the dependencies then the node_modules dir is deleted.

udnisap:
https://github.com/travis-ci/travis-ci/issues/3758

udnisap:
and how about adding `iojs` too in the list ?

josephfrazier:
> sudo is removed as it is the default configuration. 

Not quite, see https://docs.travis-ci.com/user/workers/container-based-infrastructure/#Routing-your-build-to-container-based-infrastructure

> caching is removed as it might have problems when we change the dependency tree.

I think there are only problems if `npm install` were to not overwrite old dependencies properly. See here for details on how caching works https://docs.travis-ci.com/user/caching/#How-does-the-caching-work%3F In any case, I think we should wait until there are *actually* problems with caching before removing it.

> added v4 and v5 for building environments

Shouldn't `v4` always resolve to the same version as `4.2`? `v5` and `iojs` sound nice though.

udnisap:
>sudo is removed as it is the default configuration.
Not quite, see https://docs.travis-ci.com/user/workers/container-based-infrastructure/#Routing-your-build-to-container-based-infrastructure

I was referring to 
https://docs.travis-ci.com/user/migrating-from-legacy/#tl%3Bdr


>Shouldn't v4 always resolve to the same version as 4.2? v5 and iojs sound nice though.

It is but this will use the major version with the latest update

josephfrazier:
> I was referring to 
https://docs.travis-ci.com/user/migrating-from-legacy/#tl%3Bdr

That link points out that `sudo: false` is needed in order to use the container-based infrastructure (for pre-2015 repos). According to the Travis builds for this PR, removing `sudo: false` reverts to the legacy infrastructure: https://travis-ci.org/feross/webtorrent/jobs/93513185

> It is but this will use the major version with the latest update

Oh, I see now. I had assumed in #483 that since 4.2 was an LTS release, there would be no further 4.x.0 releases, but that's not necessarily the case, according to items 7 of https://github.com/nodejs/LTS/blob/d576207e6aff6f10006a47c7f7cfd02906352b10/README.md#lts-plan

feross:
Hey guys,

Thanks for all the discussion about how to make the travis configuration better.

Enabling the tests for node v5 sounds good. I'd like to support all node LTS versions (currently 0.12 and 4) as well as the latest node version. (There's a cool trick to get the latest version of node by using "node" as the version.) I have a PR out for that here: https://github.com/feross/webtorrent/pull/501

We should remove the npm cache. These builds don't take very long and it's causing test failures. I [removed the cache](https://github.com/feross/webtorrent/commit/08ae55cb702ae1c397e295297a8dd2505e6cdb1f) in #500 after it caused a failure. An old version of `create-torrent` continued to be used even though a new one exists on npm. Removing the cache option fixed it.

We should leave "sudo: false" alone since it's required for projects created before the travis container infrastructure existed.

PR #500, opened 2015-12-02
Update npm dependencies

feross:


gillesdemey:
:+1: seems like this is also required for #501 to be green on Travis.

feross:
thanks @gillesdemey - merged.

PR #504, opened 2015-12-02
timeRemaining should be 0 once done (Fix #503)

### LINKED ISSUE #503: Time Remaining is Infinity even if done
timeRemaining torrent property is Infinity once done as downloadSpeed() is 0.

DiegoRBaquero:
timeRemaining property should be 0 once a torrent has been downloaded (or is being seeded)


feross:
@DiegoRBaquero – thanks for the bugfix. Released as 0.63.1.

PR #506, opened 2015-12-02
Update README (Fix #495)

### LINKED ISSUE #495: File path seeding problems
I'm trying this one 

```
var client = new WebTorrent()
client.seed('foo')
```

And getting back 
```
Uncaught TypeError: fs.stat is not a function
```

Documentation clearly says that the seed method accepts file path as a string. What am I doing wrong?

[JS Bin example](http://jsbin.com/baqasixuje/edit?html,output)

DiegoRBaquero:
The path to file/folder only works in Node.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ Start seeding a new torrent.
 
 `input` can be any of the following:
 
-- path to the file or folder on filesystem (string)
+- path to the file or folder on filesystem (string) (Node.js only)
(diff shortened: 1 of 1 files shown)


feross:
Thanks – can you also send a PR to `webtorrent-www`? The docs are here: https://github.com/feross/webtorrent-www/blob/master/server/views/docs.jade

I'll make an issue to unify these into one file so we don't need to keep them manually in sync.

feross:
Test failure is due to create-torrent changes. We should change the test to not assume certain trackers will exist.

PR #509, opened 2015-12-03
browser: exclude two unused packages

feross:


feross:
This saves a few KB in the browser bundle size.

PR #520, opened 2015-12-09
Initialize pieces on constructor

DiegoRBaquero:
Many prototype functions and properties use it and if it hasn't been initialized it will throw error as it's undefined.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -81,8 +81,9 @@ function Torrent (torrentId, opts) {
   this.numBlockedPeers = 0
   this.files = null
   this.done = false
-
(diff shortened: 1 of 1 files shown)


feross:
Good catch. Released as v0.63.4.

Thanks for taking the initiative on this, and sorry for not releasing this sooner. I'm currently on vacation and have spotty internet. I'm back on Dec 22.

But next time, please don't commit the built `webtorrent.min.js` or push any git tags. That should be done when I run `npm publish`. I'm going to continue keeping the power to declare releases for myself on this for now, though I may add others in the future.

PR #525, opened 2015-12-12
Issue 163

imsnif:
Global upload and download throttle for TCP peers as discussed in the below issue:

https://github.com/feross/webtorrent/issues/163


gpetrov:
@feross is this PR still considered as solution for issue https://github.com/feross/webtorrent/issues/163
We really want to be able to limit the upload speed.

chtrinh:
Correct me if I'm wrong but I don't think this PR works anymore. `throttle()` is never invoked.

PR #533, opened 2015-12-21
Deselect done fix #316

### LINKED ISSUE #316: Event 'done' isn't called when there are deselected files

DiegoRBaquero:
If the files are deselected then they are "done"


feross:
This will work most of the time (and is definitely an improvement), but it won't work if the user implicitly deselects the file by deselecting all the pieces that make up the file.

```js
torrent.deselect(0, torrent.pieces.length)
```

feross:
Here's a different approach that doesn't have the issue I mentioned above: https://github.com/feross/webtorrent/pull/544

@DiegoRBaquero What do you think?

feross:
Going with PR https://github.com/feross/webtorrent/pull/544

PR #544, opened 2015-12-28
Fix torrent.done state when there are deselected files (fix #316)

### LINKED ISSUE #316: Event 'done' isn't called when there are deselected files

feross:
Supersedes PR https://github.com/feross/webtorrent/pull/533

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1154,8 +1154,20 @@ Torrent.prototype._checkDone = function () {
     debug('file done: ' + file.name)
   })
 
-  // is the torrent done?
-  if (self.files.every(function (file) { return file.done })) {
+  // is the torrent done? (if all current selections are satisfied, or there are
+  // no selections, then torrent is done)
+  var done = true
+  for (var i = 0; i < self._selections.length; i++) {
+    var selection = self._selections[i]
+    for (var piece = selection.from; piece <= selection.to; piece++) {
+      if (!self.bitfield.get(piece)) {
+        done = false
+        break
+      }
+    }
+    if (!done) break
+  }
+  if (!self.done && done) {
     self.done = true
     self.emit('done')
     debug('torrent done: ' + self.infoHash)


lib/torrent.js
```diff
@@ -1154,8 +1154,20 @@ Torrent.prototype._checkDone = function () {
...
+    var selection = self._selections[i]
+    for (var piece = selection.from; piece <= selection.to; piece++) {
+      if (!self.bitfield.get(piece)) {
+        done = false
+        break
+      }
+    }
+    if (!done) break
+  }
+  if (!self.done && done) {
```
feross: This is an extra safety-net to ensure we don't emit 'done' twice.

lib/torrent.js
```diff
@@ -1154,8 +1154,20 @@ Torrent.prototype._checkDone = function () {
...
-  if (self.files.every(function (file) { return file.done })) {
+  // is the torrent done? (if all current selections are satisfied, or there are
+  // no selections, then torrent is done)
+  var done = true
+  for (var i = 0; i < self._selections.length; i++) {
+    var selection = self._selections[i]
+    for (var piece = selection.from; piece <= selection.to; piece++) {
+      if (!self.bitfield.get(piece)) {
+        done = false
+        break
```
feross: This isn't strictly necessary – just an optimization

lib/torrent.js
```diff
@@ -1154,8 +1154,20 @@ Torrent.prototype._checkDone = function () {
...
+      }
+    }
+    if (!done) break
```
feross: Same here – this is just an optimization

DiegoRBaquero:
Looks good

feross:
Released as 0.65.1.

Issue #534, opened 2015-12-23
Increase web seed request size from 16kb


hualiwang:
Everytime it return 16.5kb data size, I think it is too small .

rom1504:
It's not possible currently, but @DiegoRBaquero is working on it. I think he was pretty close to have it working.

DiegoRBaquero:
At the moment, you can't. I managed to get it up to the piece length, which you can test here: https://rawgit.com/DiegoRBaquero/BMagnet/master/demo/bmagnet.html

I don't really know when I'm going to make the PRs, because I did the changes manually in the node modules and have been sick.

feross:
@DiegoRBaquero Do you mind quickly describing the changes you made? I can go through and make all the changes to the various modules. I have free time at the moment.

rom1504:
I guess https://github.com/DiegoRBaquero/BMagnet/commit/5b9e389d59c31c1963420543bfcbc75a5c4661e8

DiegoRBaquero:
@feross Going to do that now.

feross:
Make the changes or describe them? ;)

DiegoRBaquero:
@feross Make the changes and PR to the respective modules

feross:
nice :+1:

DiegoRBaquero:
@feross Any updates or changes that should be made? Maybe allow the users to selected through the opts the max size?

feross:
This is released as 0.81.0. The web seed request size is now equal to the piece size, which is much better.

PR #545, opened 2015-12-28
Download larger chunks from webseeds.

DiegoRBaquero:
This is the last part of the changes required to download larger chunks from webseeds.

1st: https://github.com/feross/bittorrent-swarm/pull/29
2nd: https://github.com/feross/torrent-piece/pull/1

Basically the changes done to torrent.js mean:
- If the peer from which we are going to make a request is a webseed, then reserve all remaining pieces; if the request fails, cancel all those pieces.
- set the chunks and check for piece completion

This should fix issue #534 :)


feross:
I'm taking a look at this now. Super sorry for the huge delay :blush:

lib/torrent.js
```diff
@@ -1063,6 +1063,8 @@ Torrent.prototype._request = function (wire, index, hotswap) {
   var self = this
   var numRequests = wire.requests.length
 
+  var webPeer = wire.remotePort === 80 || wire.remotePort === 443
```
feross: Normal peers can listen on port 80 or 443. There are no rules about what ports a client can use, so this might match a normal peer. Not a huge deal right now, but it's better to be explicit since we might add other custom code for web seeds in the future and it could cause trouble.

feross: I fixed this up!

feross:
Excellent news! This is released as 0.81.0 :+1: :sunglasses:

PR #548, opened 2015-12-29
Fix maxPeers

imsnif:
Basically, the maxPeers option was ignored by the constructor.
I'm not sure if this was intentionally left not working or not - but this will be really helpful. So I'm hoping this was not intentional.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,6 +64,8 @@ function WebTorrent (opts) {
   self.downloadSpeed = speedometer()
   self.uploadSpeed = speedometer()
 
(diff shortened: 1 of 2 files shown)


imsnif:
Also, for order's sake: https://github.com/feross/webtorrent/issues/395

feross:
Thanks for the PR. I'm going to rename this option to `maxConns` since it's more descriptive. The `bittorrent-swarm` package only limits the number of connections, not peers.

feross:
Released as 0.66.0.

PR #549, opened 2015-12-30
Fix timeRemaining issue

DiegoRBaquero:
When the swarm hasn't been created but the timeRemaining property is accessed, it causes an error.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -97,7 +97,7 @@ function Torrent (torrentId, opts) {
 Object.defineProperty(Torrent.prototype, 'timeRemaining', {
   get: function () {
     if (this.done) return 0
-    if (this.swarm.downloadSpeed() === 0) return Infinity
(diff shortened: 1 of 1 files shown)


lib/torrent.js
```diff
@@ -97,7 +97,7 @@ function Torrent (torrentId, opts) {
 Object.defineProperty(Torrent.prototype, 'timeRemaining', {
   get: function () {
     if (this.done) return 0
-    if (this.swarm.downloadSpeed() === 0) return Infinity
+    if (this.swarm === undefined || this.swarm.downloadSpeed() === 0) return Infinity
```
feross: This would also fail if swarm is `null`. It's better to just to do:

```js
if (!this.swarm || this.swarm.downloadSpeed() === 0) return Infinity
```

feross:
Released as 0.67.1.

PR #553, opened 2016-01-03
File select with priority - fixes #550

### LINKED ISSUE #550: File set priority
Just leaving this for reference, will make PR soon.

DiegoRBaquero:
Allows selecting with a set priority, if not passed as argument, it will default to false/0


feross:
Released as 0.69.0.

PR #555, opened 2016-01-04
Fix downloaded property access

DiegoRBaquero:
When bitfield is not yet created, accessing downloaded property will throw an error.


feross:
Released as 0.71.1.

PR #560, opened 2016-01-07
add Seedshot to apps in production

shime:


feross:
Nice, would be cool if you added a bit more info about how this works to the website or GitHub repo.

PR #568, opened 2016-01-10
show help if there is not exactly one argument for info, create, down…

### LINKED ISSUE #567: Command line crashes on wrong number of arguments
Type `webtorrent create`. This probably happens for other commands too.

We should print the help message instead.

rom1504:
…load, add and seed, fix #567


feross:
Nice, thanks!

feross:
Released as webtorrent 0.73.0.

PR #595, opened 2016-02-04
Provide a way to give extra data to the tracker on announce

yciabaud:
This PRs is dependent on feross/bittorrent-tracker#107 and  feross/torrent-discovery#15 to provide a callback for the client to get updated params and send it along with the announce message to the tracker.

It sends updated download stats (uploaded/downloaded/letf and adds a callback in th options to extend it with additionnal data.

The callback is called before every announce to get updated data from the client.


feross:
What's your use case for sending extra parameters to the tracker? Some kind of authentication?

lib/torrent.js
```diff
@@ -90,6 +90,7 @@ function Torrent (torrentId, opts) {
   this.pieces = []
   this._selections = []
   this._critical = []
+  this._getExtraAnnounceOpts = opts.getExtraAnnounceOpts
```
feross: I'd rather name this consistently with the `bittorrent-tracker` option. `getAnnounceOpts`

lib/torrent.js
```diff
@@ -280,7 +281,15 @@ Torrent.prototype._onSwarmListening = function () {
...
     port: self.client.torrentPort,
     rtcConfig: self.client._rtcConfig,
-    wrtc: self.client._wrtc
+    wrtc: self.client._wrtc,
+    getAnnounceOpts: function () {
+      return {
+        uploaded: self.uploaded,
+        downloaded: self.downloaded,
+        left: self.length - self.downloaded,
+        extraAnnounceOpts: self._getExtraAnnounceOpts && self._getExtraAnnounceOpts()
```
feross: Instead of passing the user's custom announce opts separately in this `extraAnnounceOpts` property, can we merge their properties with the `uploaded`, `downloaded`, and `left` params?

Thinking something like this:

```js
getAnnounceOpts: function () {
  return extend({
    uploaded: self.uploaded,
    downloaded: self.downloaded,
    left: self.length - self.downloaded
  }, self._getExtraAnnounceOpts && self._getExtraAnnounceOpts())
})
```

This ordering will mean that the user's values for `uploaded`, `downloaded`, and `left` (if they exist) will take precedence.

feross:
LGTM :+1: after the inline comments and https://github.com/feross/bittorrent-tracker/pull/107 comments are addressed.

yciabaud:
For now I am sending per wire statistics in order to track data transfer between peers, the bitfield to compute video playing stats (see when the consumer seeks and stops the video...) but later I want my tracker to provide an optimized peerlist to reduce latency. So I will probably retrieve more metrics.

feross:
Good news! This is released as 0.80.0.

yciabaud:
Thank you for finishing this.

feross:
Of course! Thanks for the PR!

PR #598, opened 2016-02-06
Chromecast torrent title

Sitebase:
Attach the torrent name so chromecast shows the name of the torrent file on the preload screen.


feross:
Nice! Released as 0.73.1.

PR #600, opened 2016-02-06
Couple of changes to cmd.js

zcecc22:
Commits documented and signed off. Ready to discuss as needed.

### DIFF
--- a/bin/cmd.js
+++ b/bin/cmd.js
@@ -48,7 +48,8 @@ var argv = minimist(process.argv.slice(2), {
     a: 'announce',
     q: 'quiet',
     h: 'help',
-    v: 'version'
+    v: 'version',
+    e: 'exit'
   },
   boolean: [ // options that are always boolean
     'airplay',
@@ -62,7 +63,8 @@ var argv = minimist(process.argv.slice(2), {
     'quiet',
     'help',
     'version',
-    'verbose'
+    'verbose',
+    'exit'
   ],
   string: [ // options that are always strings
     'out',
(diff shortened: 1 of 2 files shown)


feross:
I'm curious why you need the `--exit` option? The auto-closing behavior seems to work fine for me. What's your use case?

bin/cmd.js
```diff
@@ -538,58 +548,59 @@ function drawTorrent (torrent) {
...
-        '{green:Blocked peers:} {bold:' + torrent.numBlockedPeers + '}  ' +
-        '{green:Hotswaps:} {bold:' + hotswaps + '}'
-      )
-    }
+    line(
+    '{green:Queued peers:} {bold:' + torrent.swarm.numQueued + '}  ' +
+    '{green:Blocked peers:} {bold:' + torrent.numBlockedPeers + '}  ' +
+    '{green:Hotswaps:} {bold:' + hotswaps + '}'
+    )
+
```
feross: This is pretty advanced information. Why do you want to show it to all users?

zcecc22:
Seeding the torrent when done downloading.

So far the only work around Ive found was to relaunch with seed instead of
download which is not efficient on several gb torrents (long rehash).

Is'nt it best practice to keep sharing by default?

BR,
V

feross:
Yeah, you make a good point. My concern is that lots of folks use the command line to fetch files and then do other things with them. Example:

```bash
webtorrent magnet_uri --stdout | some-other-program
```

We can make it so that `--stdout` implies `--exit`.

In the future, please send separate PRs for each change you want to make so they can be discussed and merged individually.

feross:
I like this PR, except for the changes you made to the UI. Can you remove those? It doesn't make sense to me to hide the detailed peer info under `--verbose` when we're already clearing the entire terminal and taking up the whole terminal buffer.

In another PR/issue, I'm open to making a more minimalist download UI similar to the npm3 one:



And we can hide the current one behind `--verbose`.

Sound good?

zcecc22:
Sounds good.

I ll open new PRs for the port and exit including the implicit for stdout.

Feel free to drop this one.

Cheers,
V

feross:
Thanks for opening new PRs. That's really awesome of you. I'll look for those.

PR #602, opened 2016-02-08
Add appveyor.yml for automated Windows testing

josephfrazier:
https://github.com/feross/webtorrent/issues/429 reminded me that webtorrent doesn't have any automated Windows testing. [AppVeyor](http://www.appveyor.com/) is a service that can provide that for us. Here's an example build: https://ci.appveyor.com/project/josephfrazier/webtorrent/build/1.0.6

Note that simply merging this isn't quite enough to get builds running. AppVeyor also needs to be configured [here](https://ci.appveyor.com/projects/new). I should be able to do this if/once this is merged.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,35 @@
+# Explicitly ensure line endings aren't mangled
+# http://www.appveyor.com/docs/lang/nodejs-iojs#line-endings
+init:
+  - git config --global core.autocrlf input
+
+# Test against these versions of Node.js
+# https://www.appveyor.com/docs/lang/nodejs-iojs#testing-under-multiple-versions-of-node-js-or-io-js
+environment:
+  matrix:
+  - nodejs_version: "" # https://www.appveyor.com/docs/installed-software#node-js
+  - nodejs_version: "4"
+  - nodejs_version: "0.12"
+
+# http://www.appveyor.com/docs/build-cache
+cache:
+  - node_modules
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
+  - ps: Install-Product node $env:nodejs_version
+  # install modules
+  - npm install
+
+# Post-install test scripts.
+test_script:
+  # Output useful info for debugging.
+  - node --version
+  - npm --version
+  # run tests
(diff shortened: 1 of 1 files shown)


feross:
Nice, hopefully this will help us catch when there's a regression on Windows! Much appreciated @josephfrazier!

feross:
It looks like it's all configured, if you go here: https://ci.appveyor.com/project/feross/webtorrent

Also, added a badge to the readme :)

josephfrazier:
Sweet, thanks! There's a noticeable delay on the builds, but they're better late than never.

One thing that I thought you might be interested in is that [the `node_modules` directory is cached for the AppVeyor builds](https://github.com/feross/webtorrent/commit/aa8c10cf840376debe874c1eb5cbb06fed948bc1#diff-180360612c6b8c4ed830919bbb4dd459R14). I had done [the same for Travis builds](https://github.com/feross/webtorrent/commit/8b2e16cdc647bd4076af3ea3a83f04a7e01f6772), but that changes, but that change was [undone](https://github.com/feross/webtorrent/commit/08ae55cb702ae1c397e295297a8dd2505e6cdb1f) (for reasons [I didn't quite agree with](https://github.com/feross/webtorrent/commit/08ae55cb702ae1c397e295297a8dd2505e6cdb1f#commitcomment-14744485)). Do you want to remove the caching here as well?

feross:
@josephfrazier Just saw your comment - sorry for missing it before. I think you're generally right. We should update `package.json` when a newer version is *required* for webtorrent to work correctly. Let's try to do that going forward :+1: 

That said, I prefer not to cache the `node_modules` folder on CI since I want to catch bugs that exist in the latest published packages, i.e. what a user will experience when running `npm install webtorrent` for the first time.

feross:
For example, if we fix a bug and add a test for it, but forget to update `package.json` to be greater than the buggy version, that's bad (since users could still have the bug), but not as bad as if `npm install webtorrent` gets completely broken by a bad patch version of a dependency. I care more about catching the latter. :)

josephfrazier:
Oh, that makes sense now, given that the new user experience is prioritized. I'll update appveyor.yml accordingly. Thanks for the explanation!

feross:
@josephfrazier Thanks for caring so much about testing and reliability! :)

PR #603, opened 2016-02-09
Clarify `opts.name` default behaviour

retrohacker:
First attempt at documenting the behaviour of `create-torrent` when `opts.name` isn't defined.

This could use some work.


retrohacker:
This is from https://github.com/feross/create-torrent/issues/23

feross:
Thanks for documenting this. Ideally, we can just improve the core behavior so we don't need to warn people about anything in the readme. But this works for now!

@retrohacker Thanks!

PR #605, opened 2016-02-09
Use Electron to run browser tests with Travis

josephfrazier:
This makes the automated testing applied to pull requests a little more comprehensive. A couple caveats:

* This doesn't work on AppVeyor
* Most of the browser-specific tests are skipped

I'm hoping both of these can be resolved later, but this seems like a worthwhile change regardless. Here are example builds from my personal repo: [Travis](https://travis-ci.org/josephfrazier/webtorrent/builds/107940116), [AppVeyor](https://ci.appveyor.com/project/josephfrazier/webtorrent/build/1.0.10)


feross:
Nice work! I see a one way we can improve this further:

- Run sauce tests locally. With this PR, these are only run when certain environment variables are set, but locally most people their credentials in `.zuulrc`. Locally can we just run all 3: node, electron, and sauce labs?

josephfrazier:
Great idea! I'll open a PR for that.

PR #606, opened 2016-02-09
Don't cache node_modules in AppVeyor tests

josephfrazier:
See the discussion at
https://github.com/feross/webtorrent/pull/602#issuecomment-181987938

> I prefer not to cache the node_modules folder on CI since I want to
> catch bugs that exist in the latest published packages, i.e.  what a
> user will experience when running npm install webtorrent for the first
> time.


feross:
LGTM :+1:

PR #607, opened 2016-02-09
can't download a magnet URI

substack:
It's entirely possible that I'm setting up webtorrent wrong, but I can't get this simple seed+download test to work in both node, electron, and I was getting the same behavior with a similar setup for the browser. Originally I thought the problem was related to streaming, but the problem is present with files and buffers too.

If I change:

``` js
client.add(torrent.magnetURI, function (dl) {
```

to:

``` js
client.add(torrent, function (dl) {
```

then this test works. Maybe there is an issue with trackers or magnet links? I haven't made much progress on what might be causing this issue.

### DIFF
--- a/test/stream.js
+++ b/test/stream.js
@@ -1,0 +1,63 @@
+/* global Blob */
+
+var test = require('tape')
+var Readable = require('readable-stream').Readable
+var WebTorrent = require('../')
+var concat = require('concat-stream')
+var Tracker = require('bittorrent-tracker/server')
+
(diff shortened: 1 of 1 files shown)


feross:
@substack Thanks for the test case! I'll see what I can figure out.

feross:
@substack I figured it out! Fixed in v0.74.0. 

Thanks for the test - it was really helpful while debugging this. Let me know if you run into any other issues!

PR #619, opened 2016-02-15
Use .zuulrc credentials for Sauce Labs tests

josephfrazier:
As discussed in https://github.com/feross/webtorrent/pull/605#issuecomment-183823968


josephfrazier:
I pushed the `use-zuulrc-credentials` branch [into this repo](https://github.com/feross/webtorrent/tree/use-zuulrc-credentials) as well, so we can see the build that has access to the Sauce Labs credentials in .travis.yml: https://travis-ci.org/feross/webtorrent/builds/109267142

feross:
It's too bad that the specific way that zuul looks for .zuulrc had to be inlined in here, but at least the zuul tests run locally again!

feross:
Attempted a simpler solution here: https://github.com/feross/webtorrent/pull/626

PR #620, opened 2016-02-15
Wrap addrToIPPort in try/catch

niieani:
It was throwing an error when addr was not parsable, causing webtorrent to die.
See related: https://github.com/feross/bittorrent-swarm/pull/31.


feross:
Thanks for the PR. I needed to make some changes to it. You can see them here: https://github.com/feross/webtorrent/commit/42f94e89b66f39d5f0079d7d22f902f4b2013ad3

This is released as 0.77.0.

PR #626, opened 2016-02-19
test: Always run saucelabs tests locally

feross:
I think this is a simpler approach for #619. Always run the sauce labs
tests locally. If users lack the right sauce labs credentials, they’ll
get an error and hopefully set up zuul correctly before sending a PR.


feross:
Tests all pass. Merging!

PR #630, opened 2016-02-20
call unref only when present

substack:
This patch to the command-line client guards calls to `.unref()` so unref is only called when present. Electron doesn't appear to have `.unref()` methods. Now with this patch I can do:

```
$ electron-spawn `which webtorrent` MAGNET
```

to download a magnet over webrtc from the command-line using webtorrent.


feross:
0.78.1

Issue #653, opened 2016-03-02
Consider moving webtorrent CLI to its own package


feross:
This would let us remove lots of dependencies that are only used in the CLI.

rom1504:
Sounds good but then cli users would have to npm install -g webtorrent-cli (and then use webtorrent-cli in the command line) ? Not sure how much of a problem that would be. It's certainly a breaking change though.

feross:
Released as 0.85.0.

PR #656, opened 2016-03-03
remove webtorrent-cli

rom1504:
would be #653 

This PR is an experiment, to see what removing the cli means.

If that's what we want to do, then a webtorrent-cli package should be created out of what is deleted in that PR.

With this PR node_modules contains 615 top level directories which sums up to a 16285 files (1m56.962s install time)
Without it, it contains 638 top level directories which sums up to 20089 files (2m21.811s install time)

(these numbers are for a full installation including devDependencies and optionalDependencies)

I think I removed all the dependencies used in the cli, and only them.


rom1504:
(is `continuous-integration/appveyor/branch` supposed to work ?)

feross:
Thanks for the PR! This looks good, but I can't merge it now because there are conflicts.

feross:
Thanks again for the excellent PR. I opened a new PR that includes your changes and without the merge conflict since I can't push to your branch.

Also, I'll add you as a contributor to this repo :)

PR #661, opened 2016-03-08
Fix links to CONTRIBUTING.md

mortonfox:


feross:
Thanks for the PR, but this file is moving back to it's original location as soon as https://github.com/feross/webtorrent/pull/660 is merged, so the links will fix themselves :)

Issue #650, opened 2016-03-01
Cleaner web seed implementation


feross:
Ideally web seeds would be handled directly as a special-case instead of treating the web seed as another peer and interpreting wire protocol requests as http requests. That was a clever hack to get this shipped quickly, but we can do better.

We can find pieces that have no blocks reserved and give those entire pieces to the web seed. The request size can even span multiple pieces.

Once this is done, we should revert this change to `torrent-piece`, which amounts to a hack: https://github.com/feross/torrent-piece/pull/1

grunjol:
Last week, while testing Diego's changes in webseed handling (nice to see they are merged now!) I found two 'issues'

1) As you mention, the request size to the server is right now the size of the piece, so depending on the piece size in the torrent definition the web server gets hammered.

2) If the web server has some kind of connection limitations (like nginx's limit_conn addr), once you complete the first pieces, the server start sending 50x errors and the web peer is discarded. The special-case should consider the number of connections to the webseed.

DiegoRBaquero:
Another issue I found yesterday and shared on Gitter is that when we have 10+ connections each of 128KB requests and your internet slow, one or more requests will timeout causing the web seed to be removed.

grunjol:
Exactly, i was trying to understand why is requesting 10+ connections... tried to mitigate with a choke() but all the requests get removed from wire and then web seed gets removed from client...

I think the solution is not try to handle web seeds as a peer in the swarm. Maybe get back to last year's approach to start a new parallel 'http download service' that reserves the pieces and return them if something gets wrong with the request.

grunjol:
@DiegoRBaquero : the fix you sent reduce the amount of requests if it is a slow webseed, but it still calculates the number of connections based on getPipelineLength using speed/piece size, but if the webseed is FAST (like in a CDN), the number of connections increase fast and in the third request you hit again with the limit per IP.

Instead i tried 

`if (isWebSeed) maxOutstandingRequests = Math.min(maxOutstandingRequests, MAX_CONCURRENT_WEBSEED_CONNECTIONS) // A webseed will handle it's real max requests`

...and it works like a charm.

In my quick test i made MAX_CONCURRENT_WEBSEED_CONNECTIONS = 2, but it can be passed as paramater in torrent opts.

DiegoRBaquero:
@grunjol The idea of the PR is to set the max requests based on the current speed, not limit to a "reasonable" http max requests limit (which I'd say is in 16-32 concurrent).

[...]

PR #663, opened 2016-03-09
Lower max requests of webseeds to real value

DiegoRBaquero:
`getPipelineLength` adds 2 extra more requests possible to each wire. See [L1303-L1305](https://github.com/feross/webtorrent/blob/master/lib/torrent.js#L1303-L1305)

In the case of a webseed, if it already has speed, let the webseed handle its real max requests.

This is merely a temporary attempt to make webseeds more stable and don't let them timeout

#650 is still on the works.


feross:
Released 0.83.0.

PR #667, opened 2016-03-11
BREAKING: Move command line to new package: webtorrent-cli

### LINKED ISSUE #653: Consider moving webtorrent CLI to its own package

feross:
This allows us to remove lots of dependencies that are only used in the CLI, and makes the WebTorrent.app bundle a bit lighter.

Supersedes @rom1504’s excellent PR (#656) which wasn’t merging cleanly.

Fixes #653.


feross:
Released as 0.85.0.

PR #673, opened 2016-03-14
add default webrtc trackers when there is webrtc support 

### LINKED ISSUE #672: Add websocket trackers directly in webtorrent
websocket trackers are added to existing torrents in all the webtorrent dependencies that support webrtc:
* [webtorrent-hybrid](https://github.com/feross/webtorrent-hybrid/blob/d920fa1e01a6ecaabbcbf040d666740bf74569b4/lib/global.js#L6)
* [instant.io](https://github.com/feross/instant.io/blob/c8924fe4cea665c938911edf5b7efaa2c4b1e9ab/client/index.js#L15)
* [webtorrent-app](https://github.com/feross/webtorrent-app/blob/6ba0d65daf8d767d500faf8829650839df6b2339/renderer/index.js#L34)

It would probably make sense to put that directly in webtorrent, if there is webrtc support.

Also see that discussion https://github.com/feross/webtorrent-hybrid/pull/22#discussion_r55977228

rom1504:
and global.WEBTORRENT_ANNOUNCE is not defined (fix #672)

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -30,6 +30,8 @@ var Swarm = require('bittorrent-swarm')
 var uniq = require('uniq')
 var ut_metadata = require('ut_metadata')
 var ut_pex = require('ut_pex') // browser exclude
+var Peer = require('simple-peer')
+var createTorrent = require('create-torrent')
 
 var File = require('./file')
 var RarityMap = require('./rarity-map')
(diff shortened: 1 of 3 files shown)


rom1504:
It would also be possible to just remove `global.WEBTORRENT_ANNOUNCE` and use a non global variable instead.

I guess always adding webrtc default trackers is okay until torrent with webrtc trackers are more common.

feross:
Thanks for the pull request. I appreciate the work you put in, but I don't think this is right for WebTorrent. See my comment here for more details: https://github.com/feross/webtorrent/issues/672#issuecomment-196614860

PR #680, opened 2016-03-16
added Buffer and stream.Readable for client.seed

KenRmk:
Added few missing details from create-torrent


Autarc:
In addition to the documentation update I suggest that `create-torrent` also [accepts buffer and stream inputs without a name](https://github.com/feross/create-torrent/pull/31) to simplify its usage and prevent setup overhead.

PR #692, opened 2016-03-25
provide the torrent instance in the opts.store() options

substack:
Currently, `opts.store()` is called with the chunkLength, the `opts.length` containing the size of the torrent file, and the `opts.files` which is an array of the files in the torrent. This patch adds the torrent instance itself to the chunk store options, so that you can procedurally generate a path based on for example the infoHash for the storage of each torrent.


feross:
0.87.0.

PR #702, opened 2016-03-28
Only add new ut_pex peers when we're not seeding

feross:
Otherwise, we get into a loop of connect, get pex messages, disconnect
because they don't have pieces we need (because we're seeding), try
connecting to the pex peers, etc. etc.


feross:
cc @dcposch @mafintosh This seems to help the situation quite a bit!

PR #710, opened 2016-03-30
Max web conns default

DiegoRBaquero:
After a lot of testing and profiling (because I was naive) I found that it's best to keep the concurrent connections small and rotate them as they become available.

This PR sets the maxWebConns to 4 by default. Making a webseed's max request vary from 1 to 4 depending on its speed & piece length.


DiegoRBaquero:
It's safe to merge. Damn you AppVeyor.

feross:
There's an even bigger issue that @dcposch and I just discovered with web seeds. PR incoming.

feross:
https://github.com/feross/webtorrent/pull/723

feross:
0.90.2

PR #715, opened 2016-04-02
Optimization: don't re-verify unchanged files

dcposch:
Let the user specify known-good file modtimes. If the files modtime is at least
that old, then the file hasn't changed and does not need to be re-verified.

This is only valid in node when using FS backing storage, not in the browser

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -5,12 +5,12 @@ module.exports = Torrent
 var addrToIPPort = require('addr-to-ip-port')
 var BitField = require('bitfield')
 var ChunkStoreWriteStream = require('chunk-store-stream/write')
-var cpus = require('cpus')
 var debug = require('debug')('webtorrent:torrent')
 var Discovery = require('torrent-discovery')
 var EventEmitter = require('events').EventEmitter
 var extend = require('xtend')
 var extendMutable = require('xtend/mutable')
+var fs = require('fs')
 var FSChunkStore = require('fs-chunk-store') // browser: `memory-chunk-store`
 var ImmediateChunkStore = require('immediate-chunk-store')
 var inherits = require('inherits')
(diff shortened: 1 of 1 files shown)


lib/torrent.js
```diff
@@ -403,7 +408,55 @@ Torrent.prototype._onMetadata = function (metadata) {
...
+  if (self._fileModtimes && self._store === FSChunkStore) {
+    // don't verify if the files haven't been modified since we last checked
+    self.getFileModtimes(function (err, fileModtimes) {
+      if (err) return self._onError(err)
+
+      var unchanged = self.files.map(function (_, index) {
+        return fileModtimes[index] === self._fileModtimes[index]
+      }).reduce(function (a, b) {
+        return a && b
+      })
```
feross: This will probably crash when there's only one file, since a default starting value was not specified for reduce.

feross: I think `arr.every()` works better here.

lib/torrent.js
```diff
@@ -5,12 +5,12 @@ module.exports = Torrent
 var addrToIPPort = require('addr-to-ip-port')
 var BitField = require('bitfield')
 var ChunkStoreWriteStream = require('chunk-store-stream/write')
-var cpus = require('cpus')
```
feross: This needs to be removed from `package.json`

lib/torrent.js
```diff
@@ -45,6 +45,8 @@ var PIPELINE_MAX_DURATION = 1
 var RECHOKE_INTERVAL = 10000 // 10 seconds
 var RECHOKE_OPTIMISTIC_DURATION = 2 // 30 seconds
 
+var FILESYSTEM_CONCURRENCY = 2
```
feross: Why 2?

lib/torrent.js
```diff
@@ -403,7 +408,55 @@ Torrent.prototype._onMetadata = function (metadata) {
...
+      var unchanged = self.files.map(function (_, index) {
+        return fileModtimes[index] === self._fileModtimes[index]
+      }).reduce(function (a, b) {
+        return a && b
+      })
+      if (unchanged) {
+        for (var index = 0; index < self.pieces.length; index++) {
+          self._markVerified(index)
+        }
+        self._onStore()
```
feross: This could be optimized further to handle the case where only some files are modified, i.e. mark pieces that are wholly-contained within unchanged files as verified, and reverify the rest.

feross:
Thanks for the PR, @dcposch! Looks great.

PR #723, opened 2016-04-06
Base web seed pipeline length on piece length

feross:
Before this, the web seed pipeline length was based on the block size,
just like it is for wire connections, which are block-based.

This meant that we were massively over-estimating the number of http
requests to make to the web seed servers.

Now we use the piece length, since each web seed request is a piece
length in size.


feross:
Reviewed in-person by @dcposch. :

feross:
0.90.2

dcposch:
:+1:

PR #747, opened 2016-04-10
Added torrent.on events to documentation

Tercus:
Added the other torrent.on events to the documentation.


feross:
Thanks! This is a good start. I will add some more information after merge :)

feross:
Changes here: https://github.com/feross/webtorrent/commit/7a7c4a8b8c49f5c92b7c20ff439bc8614f7d607e

PR #750, opened 2016-04-12
API improve

DiegoRBaquero:
infohash, metadata and ready note. Added uploaded

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -334,21 +334,6 @@ listen to this event, but it may aid in debugging.
 Emitted when the torrent encounters a fatal error. The torrent is automatically destroyed
 and removed from the client when this occurs.
 
-## `torrent.on('infoHash', function () {})`
-
-Emitted when the info hash of the torrent has been determined.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -400,6 +399,21 @@ See the `bittorrent-protocol`
 [extension api docs](https://github.com/feross/bittorrent-protocol#extension-api) for more
 information on how to define a protocol extension.
 
+## `torrent.on('infoHash', function () {})`
+
+Emitted when the info hash of the torrent has been determined. Can only be accessed synchronously as client.add calls onTorrent callback after 'infoHash' event has been emitted.
```
feross: I don't think we need the extra text. It makes it more confusing IMO. It's already clear from the fact that this is an event that it's async.

feross:
Documenting the `'upload'` event is great, but I'm -1 for the other changes.

DiegoRBaquero:
@feross removed extra comments

PR #756, opened 2016-04-15
Api for upload trigger on torrent.

Lunik:
I use this trigger but it's not referenced into the api docs !


DiegoRBaquero:
I'm guessing you closed it because it was in this pending PR https://github.com/feross/webtorrent/pull/750

Thank you for the PR nonetheless.

PR #762, opened 2016-04-21
BREAKING: Many fixes; all leaks fixed

### LINKED ISSUE #248: Reduce memory usage

### LINKED ISSUE #317: Memory doesn't get freed after removing torrent from client
Memory goes up and never goes down even after client.remove is called and nothing is downloading. Any feedback on this?

### LINKED ISSUE #635: There is already a swarm listening on port (error)
I don't have reproduction steps for this. This happened when destroying a running torrent (not sure if that's related to the error or not).

```ts
events.js:154
      throw er; // Unhandled 'error' event
      ^

Error: There is already a swarm with info hash 3f19b149f53a50e14fc0b79926a391896eabab6f listening on port 62108
    at node_modules/bittorrent-swarm/lib/tcp-pool.js:100:22
    at nextTickCallbackWith0Args (node.js:453:9)
    at process._tickCallback (node.js:382:13)
```

### LINKED ISSUE #712: Memory leaks [META]
Webtorrent and its modules has a lot of code similar to this simple example, which is actually a good example of memory leak:
```javascript
var Item = function(node) {
  var self = this

  this.node = node

  window.addEventListener('resize', function resizeHandler() {
    if (!self.node) {return}
    self.node.textContent = 'window width: ' + window.width
  })
}

Item.die = function() {
  this.node = null
}
```
Using:
```javascript
var item = new Item(document.createElement('div'))
item.die()
```

resizeHandler still in window resize events, it still has referrence to variable `self`,
yes we made `node` free, but `item` stuck in memory;

### To fix it we should not have closures in event listeners after `die`:
```javascript
var Item = function(node) {
  var self = this

  this.node = node

  this.resizeHandler = resizeHandler
  window.addEventListener('resize', resizeHandler)
  function resizeHandler() {
    if (!self.node) {return}
    self.node.textContent = 'window width: ' + window.width
  }
  
}

Item.die = function() {
  window.removeEventListener('resize', this.resizeHandler)
  this.node = null
}
``` 

(you can imaging that `setTimeout(fn, 30000)` and even `process.nextTick`/`setImmidiate` will be problem for GC too)

### LINKED ISSUE #721: When torrent emits 'error', client should not too
Otherwise, users who listen for errors in both places will get the error twice.

feross:
### Added

- `client.listening` property to signal whether TCP server is listening for incoming
  connections.

- `client.dhtPort` property to show the actual chosen DHT port (parallel to `client.torrentPort` for the TCP torrent listening server)

### Changed

- Merged `Swarm` class into `Torrent` object. Properties on `torrent.swarm` (like
  `torrent.swarm.wires`) now exist on `torrent` (e.g. `torrent.wires`). Deleted a ton of code!

- Deprecate: Do not use `torrent.swarm` anymore. Use `torrent` instead.

- `torrent.addPeer` can no longer be called before the `infoHash` event has been
  emitted.

- Remove `torrent.on('listening')` event. Use `client.on('listening')` instead.

- Remove support from `TCPPool` for listening on multiple ports. This was not used by
  WebTorrent and just added complexity. There is now a single `TCPPool` instance for the
  whole WebTorrent client.

- Deprecate: Do not use `client.download()` anymore. Use `client.add()` instead.

- Only pass `torrent.infoHash` to the Chunk Store constructor, instead of the `Torrent`
  instance itself, to prevent accidental memory leaks of the `Torrent` object by the
  store. (Open an issue if you were using other properties. They can be re-added.)

- Non-fatal errors with a single torrent will be emitted at `torrent.on('error')`. You
  should listen to this event. Previously, all torrent errors were also emitted on
  `client.on('error')` and handling `torrent.on('error')` was optional. The new design is
  better since now it is possible to distinguish between fatal client errors
  (`client.on('error')`) when the whole client becomes unusable versus recoverable errors
  where only a single torrent fails (`torrent.on('error')`) but the client can continue to
  be used. However, if there is no `torrent.on('error')` event, then the error will be
  forwarded to `client.on('error')`. This prevents crashing the client when the user
  only has a listener on the client, but it makes it impossible for them to determine
  a client error versus a torrent error.

### Fixed

- If `client.get` is passed a `Torrent` instance, it now only returns it if it is present
  in the client.

- Errors creating a torrent with `client.seed` are now emitted on the returned `torrent`
  object instead of the client (unless there is no event listeners on
  `torrent.on('error')` as previously discussed). The torrent object is now also destroyed
  automatically for the user, as was probably expected.

- Do not return existing torrent object when duplicate torrent is added. Fire an
  `'error'` event instead.

- Memory leaks of `Torrent` object caused by many internal subclasses of WebTorrent,
  including `RarityMap`, `TCPPool`, `WebConn`, `Server`, `File`, `Torrent`.

- `client.ratio` and `torrent.ratio` are now calculated as `uploaded / received` instead
  of `uploaded / downloaded`.

Closes #721 
Closes #712
Closes #635 
Closes #317 
Closes #248


test/swarm.js
```diff
@@ -1,82 +1,82 @@
-var hat = require('hat')
-var Swarm = require('../lib/swarm')
-var test = require('tape')
+// var hat = require('hat')
```
dcposch: Might want to delete this file. Or are you planning to port the tests to use `torrent` instead of `swarm` later?

dcposch:
This is awesome and when you exclude commented out tests and the new release notes, it's removing a lot more code than it adds.

feross:
Swarm tests are disabled for now, since there's no Swarm class anymore to test. It's part of Torrent. I'll see what testing is worth keeping and what is redundant in a future PR.

test/swarm.js
feross: Planning to port the tests later.

feross:
I want to get a PR ready for `webtorrent-cli` ready before merging this. Will release tomorrow.

grunjol:
I think this is the PR everybody was waiting for :D.

feross:
PR to webtorrent-cli wasn't so bad: https://github.com/feross/webtorrent-cli/pull/6

PR #767, opened 2016-04-23
add p2pdrop to app using webtorrent list

ajainvivek:
P2PDrop is secure file sharing application for desktop and browser.


feross:
Nice work!

ajainvivek:
Thanks @feross :)

feross:
thanks! 😀

PR #774, opened 2016-04-27
Test Node v5 on Travis/Appveyor (v6 is stable)

josephfrazier:


feross:
We only need to officially support the LTS and Current versions, i.e. the versions that are linked from https://nodejs.org/en/ But we can leave the node 5 tests in for a while -- doesn't hurt.

josephfrazier:
Ah, ok, I wasn't sure which platforms we're targeting here. I see that v5 will be supported for only two more months, so that might be a good time to remove the tests: https://nodejs.org/en/blog/release/v6.0.0/#what-about-node-js-v5

feross:
@josephfrazier Sounds good

PR #775, opened 2016-04-27
Add 'noPeers' event to torrents

harperj:
The 'noPeers' event is emitted after tracker or DHT announce events when there are still no peers in the swarm.


feross:
Thanks for this PR!

However before merging, I feel like we should work on making this event more useful to the user. If I listen to `'noPeers'`, there's no guarantee that more peers aren't about to be found from another discovery mechanism.

feross:
But this is much better than nothing, so let's merge it for now!

feross:
0.92.0

harperj:
I agree.  I'll try to take a look at this again shortly.

Issue #649, opened 2016-03-01
Move tracker options into `opts.tracker`


feross:
This will help to control the explosion of options in the WebTorrent API.

I already made the same change to `torrent-discovery` here: https://github.com/feross/torrent-discovery/commit/eb45af0999df9071e7ad51bf0e5b17595ed56b67

nkittsteiner:
Can I have a look at this? Is still required?

feross:
Yes.

PR #791, opened 2016-05-07
Move tracker options into `opts.tracker` (#649)

nkittsteiner:
Moving wtrc and rtcConfig opts to opts.tracker. Hope it helps :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,12 +68,14 @@ function WebTorrent (opts) {
   self.listening = false
   self.torrentPort = opts.torrentPort || 0
   self.dhtPort = opts.dhtPort || 0
-  self.tracker = opts.tracker !== undefined ? opts.tracker : true
+  self.tracker = opts.tracker !== undefined ? opts.tracker : {}
   self.torrents = []
   self.maxConns = Number(opts.maxConns) || 55
 
-  self._rtcConfig = opts.rtcConfig
-  self._wrtc = opts.wrtc || global.WRTC // to support `webtorrent-hybrid` package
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -68,12 +68,14 @@ function WebTorrent (opts) {
...
+  self.tracker = opts.tracker !== undefined ? opts.tracker : {}
   self.torrents = []
   self.maxConns = Number(opts.maxConns) || 55
 
-  self._rtcConfig = opts.rtcConfig
-  self._wrtc = opts.wrtc || global.WRTC // to support `webtorrent-hybrid` package
+  if (typeof self.tracker === 'object') {
+    self.tracker.rtcConfig = opts.rtcConfig
+    self.tracker.wrtc = opts.wrtc || global.WRTC // to support `webtorrent-hybrid` package
+  }
```
feross: These options should be passed into the tracker constructor, not set directly on the tracker object like this.

feross: Nevermind.

nkittsteiner:
Hi Feross, do you mean with nevermind leave it like this? Because I can fix that

feross:
@nkittsteiner Thanks a lot for this PR. It illuminated a few issues that I hadn't though of, like how to handle `opts.rtcConfig` and `opts.wrtc`, as well as how to handle duplicate `getAnnounceOpts` functions passed into `opts.tracker` and also into the opts for `client.add`.

I'll send a PR that builds on yours.

feross:
See: https://github.com/feross/webtorrent/pull/793

PR #798, opened 2016-05-12
On duplicate torrent add, don't emit 'infoHash'

feross:
Private '_infoHash' event allows client.add to check for duplicate
torrents and destroy them before the normal 'infoHash' event is
emitted. Prevents user applications from needing to deal with duplicate
'infoHash' events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # WebTorrent Version History
 
+## UNRELEASED
+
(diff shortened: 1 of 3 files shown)


feross:
This is useful for WebTorrent Desktop, where we don't want to save a torrent to the local `config.json` file (which happens on the 'infoHash' event) if an 'error' event already fired.

Issue #714, opened 2016-04-01
Proper support for "ws=", "as=" and "xs=" magnet link parameters


feross:
Full spec here: https://wiki.theory.org/BitTorrent_Magnet-URI_Webseeding
- [X] Retrieving payload data - "ws=", "as=" parameter
- [x] Retrieving metadata - "xs=" parameter

DiegoRBaquero:
This is what I shared you? I'll grab this issue and work on it!

feross:
@DiegoRBaquero Ah, yes. I forgot who originally sent it to me :)

feross:
'xs=' is implemented now, will be released after https://github.com/feross/webtorrent/pull/809 is merged

RationalCoding:
How can the `ws` parameter retreive metadata? I thought it was reserved for ordinary HTTP servers that don't "know" about the torrent protocol at all?

DiegoRBaquero:
@RationalCoding ws doesn't retrieve metadata. It just serves the files of the torrent.

RationalCoding:
I know it doesn't currently, but there's an item on the list "Retrieving metadata - "ws=" parameter". I was wondering how that could be implemented.

DiegoRBaquero:
Ahhhh, I see your point now. It shouldn't. `ws` and `as` are for direct files

feross:
This issue is done. We support all forms of "ws=", "as=" and "xs="

PR #799, opened 2016-05-13
Implement exact source (xs) for magnet URIs

Sebmaster:
Since I saw #714 stalled a bit, I wanted to check in if no one's working on this and started off with a minimal working prototype.

I know I still need to move the fixture over (also don't rely on webtorrent.io, but create a local server instead) to the other repo, but does this look okay / is anyone else working on this?

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -12,6 +12,7 @@ var extend = require('xtend')
 var extendMutable = require('xtend/mutable')
 var fs = require('fs')
 var FSChunkStore = require('fs-chunk-store') // browser: `memory-chunk-store`
+var get = require('simple-get')
 var ImmediateChunkStore = require('immediate-chunk-store')
 var inherits = require('inherits')
 var MultiStream = require('multistream')
@@ -113,6 +114,7 @@ function Torrent (torrentId, client, opts) {
 
   // for cleanup
   this._servers = []
+  this._xsRequest = null
 
   // TODO: remove this and expose a hook instead
   // optimization: don't recheck every file if it hasn't changed
@@ -269,6 +271,38 @@ Torrent.prototype._onParsedTorrent = function (parsedTorrent) {
       self._onListening()
     })
   }
+
+  if (parsedTorrent.xs && !self.info) {
+    // TODO: Could try multiple in parallel here
+    var src = parsedTorrent.xs instanceof Array ? parsedTorrent.xs[0] : parsedTorrent.xs
+    var isHTTP = src.substring(0, 7) === 'http://' || src.substring(0, 8) === 'https://'
+
+    if (isHTTP) {
+      var opts = {
+        url: src,
+        method: 'GET',
+        headers: {
+          'user-agent': 'WebTorrent (http://webtorrent.io)'
+        }
+      }
+      this._xsRequest = get.concat(opts, function (err, res, torrent) {
+        if (err || res.statusCode !== 200 || !torrent.length) return
+
+        var parsedTorrent
+        try {
+          parsedTorrent = parseTorrent(torrent)
+        } catch (err) {}
+
+        if (parsedTorrent) {
+          if (parsedTorrent.infoHash !== self.infoHash) {
+            return
+          }
+
+          self._onMetadata(parsedTorrent)
+        }
+      })
+    }
+  }
 }
 
 Torrent.prototype._processParsedTorrent = function (parsedTorrent) {
@@ -371,6 +405,10 @@ Torrent.prototype._onListening = function () {
  */
 Torrent.prototype._onMetadata = function (metadata) {
   var self = this
+  if (self._xsRequest) {
+    self._xsRequest.abort()
+    self._xsRequest = null
+  }
   if (self.metadata || self.destroyed) return
   self._debug('got metadata')
 
@@ -553,6 +591,9 @@ Torrent.prototype._destroy = function (err, cb) {
   self.destroyed = true
   self._debug('destroy')
 
+  if (self._xsRequest) {
+    self._xsRequest.abort()
+  }
   self.client._remove(self)
 
   clearInterval(self._rechokeIntervalId)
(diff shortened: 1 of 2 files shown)


lib/torrent.js
```diff
@@ -269,6 +271,34 @@ Torrent.prototype._onParsedTorrent = function (parsedTorrent) {
...
+      url: parsedTorrent.xs,
+      method: 'GET',
+      headers: {
+        'user-agent': 'WebTorrent (http://webtorrent.io)'
+      }
+    }
+    try {
+      this._xsRequest = get.concat(opts, function (err, res, torrent) {
+        if (err || !torrent.length) return
+
```
feross: Should check that `res.statusCode === 200`

feross:
Just took a quick look. Looks pretty good to me. I don't think we to/can easily support ftp:// so I would remove that test case.

Sebmaster:
Fixed. 

> I don't think we to/can easily support ftp:// so I would remove that test case.

I added that mainly to check if unsupported protocols would crash the process, since the spec says to silently ignore.

feross:
I think we should check for http:// https:// explicitly before passing into `simple-get`. That would be more reliable and we could get rid of the try-catch around `simple-get`, right?

Sebmaster:
Done.

> That would be more reliable and we could get rid of the try-catch around `simple-get`, right?

For the protocol case yeah, just hope there's nothing else `http` will throw on...

Sebmaster:
Okay, also migrated the tests over to a local server and reused the fixtures from `webtorrent-fixtures`. Should be ready (for xs, at least).

feross:
Thanks -- will take a look after getting WebTorrent Desktop v0.5 out.

Sebmaster:
I just checked the malformed URL case some more - a URL like `http://example.org:2455665555/` would now crash, which isn't great, so I'm thinking leaving the try / catch there would be better (either instead, or in addition to the protocol check).

lib/torrent.js
```diff
@@ -269,6 +271,38 @@ Torrent.prototype._onParsedTorrent = function (parsedTorrent) {
       self._onListening()
     })
   }
+
+  if (parsedTorrent.xs && !self.info) {
+    // TODO: Could try multiple in parallel here
+    var src = parsedTorrent.xs instanceof Array ? parsedTorrent.xs[0] : parsedTorrent.xs
```
feross: `Array.isArray()` is the right way to check if something is an Array.

lib/torrent.js
```diff
@@ -269,6 +271,38 @@ Torrent.prototype._onParsedTorrent = function (parsedTorrent) {
...
+    if (isHTTP) {
+      var opts = {
+        url: src,
+        method: 'GET',
+        headers: {
+          'user-agent': 'WebTorrent (http://webtorrent.io)'
+        }
+      }
+      this._xsRequest = get.concat(opts, function (err, res, torrent) {
+        if (err || res.statusCode !== 200 || !torrent.length) return
```
feross: No need to check for `!torrent.length` since that will make `parseTorrent` throw below anyway.

feross: Would be nice to add a log here with `debug` for debugging purposes.

lib/torrent.js
```diff
@@ -269,6 +271,38 @@ Torrent.prototype._onParsedTorrent = function (parsedTorrent) {
...
+
+        var parsedTorrent
+        try {
+          parsedTorrent = parseTorrent(torrent)
+        } catch (err) {}
+
+        if (parsedTorrent) {
+          if (parsedTorrent.infoHash !== self.infoHash) {
+            return
```
feross: Debug log here too

lib/torrent.js
```diff
@@ -371,6 +405,10 @@ Torrent.prototype._onListening = function () {
  */
 Torrent.prototype._onMetadata = function (metadata) {
   var self = this
+  if (self._xsRequest) {
+    self._xsRequest.abort()
+    self._xsRequest = null
+  }
```
feross: This should go after the next line.

test/node/download-metadata.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+test('Download metadata for magnet URI with xs parameter', function (t) {
+  t.plan(2)
+
+  var client = new WebTorrent({ dht: false, tracker: false })
+
+  createServer(fixtures.leaves.torrent, function (url, next) {
+    client.add(fixtures.leaves.magnetURI + '&xs=' + encodeURIComponent(url), function (torrent) {
+      t.equal(torrent.name, 'Leaves of Grass by Walt Whitman.epub')
```
feross: The name is also in the magnet URI as the `&dn` parameter. Might be better to check for another parameter.

test/node/download-metadata.js
```diff
@@ -0,0 +1,65 @@
...
+  })
+})
+
+test('Download metadata magnet URI with unsupported protocol in xs parameter', function (t) {
+  t.plan(2)
+
+  var client = new WebTorrent({ dht: false, tracker: false })
+  client.add(fixtures.leaves.magnetURI + '&xs=' + encodeURIComponent('invalidurl:example'))
+  setTimeout(function () {
+    t.ok(true, 'no crash')
```
feross: No need for this. The tests will fail if there's a thrown exception. We should add listeners for 'error' and 'warning' events to all these tests though.

Sebmaster:
I totally fixed all of those just now :smile: 

Also what about the try-catch situation with malformed urls?

feross:
Ah, thanks. I'm fixing all that up too. Sorry, should have communicated that better.

feross:
I'm adding support for multiple &xs= params too.

test/node/download-metadata.js
feross: Nevermind, I see what you were doing.

feross:
@Sebmaster See https://github.com/feross/webtorrent/pull/809

PR #809, opened 2016-05-19
Fixes for PR #799

feross:
- Support multiple &xs= params in parallel
- Fail on 'error' 'warning' events

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -114,7 +114,7 @@ function Torrent (torrentId, client, opts) {
 
   // for cleanup
   this._servers = []
-  this._xsRequest = null
+  this._xsRequests = []
 
   // TODO: remove this and expose a hook instead
   // optimization: don't recheck every file if it hasn't changed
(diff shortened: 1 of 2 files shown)


feross:
@Sebmaster Can you please review this PR when you get the chance?

lib/torrent.js
```diff
@@ -395,23 +363,96 @@ Torrent.prototype._onListening = function () {
...
 
-  // if full metadata was included in initial torrent id, use it immediately. Otherwise,
-  // wait for torrent-discovery to find peers and ut_metadata to get the metadata.
-  if (self.info) self._onMetadata(self)
+  if (self.info) {
+    // if full metadata was included in initial torrent id, use it immediately. Otherwise,
+    // wait for torrent-discovery to find peers and ut_metadata to get the metadata.
+    self._onMetadata(self)
+  } else if (self.xs) {
+    self._getMetadataFromServer()
```
feross: Moved to after listening event since it's really fast and it's when the other discovery happens. This may prevent future bugs where we expect to already be listening.

lib/torrent.js
```diff
@@ -395,23 +363,96 @@ Torrent.prototype._onListening = function () {
...
+    function onResponse (err, res, torrent) {
+      if (self.destroyed) return cb(null)
+      if (self.metadata) return cb(null)
+
+      if (err) {
+        self._debug('http error from xs param: %s', url)
+        return cb(null)
+      }
+      if (res.statusCode !== 200) {
+        self._debug('non-200 status code from xs param: %s', url)
```
Sebmaster: Print the specific status code here?

Sebmaster:
LGTM otherwise.

lib/torrent.js
feross: Good idea!

PR #819, opened 2016-05-29
Fix exception in file.js

feross:
If file is destroyed and stream ends afterwards, then an exception is
thrown because self._torrent is undefined.


feross:
0.94.2

PR #824, opened 2016-06-05
Add PeerTube to projects that use WebTorrent

Chocobozzz:
Now there is a demo :)

Closes https://github.com/Chocobozzz/PeerTube/issues/2


feross:
Nice demo!

screen shot 2016-06-05 at 17 46 55

PR #827, opened 2016-06-06
Add url to cinematrix.one

v0dy:
Added entry to cinematrix.one

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -76,6 +76,7 @@ WebTorrent is still pretty new, but it's already being used in cool ways:
 - **[Instant-Share][instant-share]** - File sharing over WebTorrent
 - **[P2PDrop][p2pdrop]** - Securely share files between peers ([source code][p2pdrop-source])
 - **[Twister][twister]** - Decentralized microblogging service, using WebTorrent for media attachments ([source code][twister-source])
+- **[Cinematrix][cinematrix]** - Popcorn time in the browser with free content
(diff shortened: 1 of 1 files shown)


feross:
I took a look at your site. It looks like it only contains public domain, creative commons, legitimate content. Is your plan to keep it that way?

v0dy:
Yes i will keep the content free! I will also add the sourcecode later on. 
Any suggestion for content is welcome!

feross:
I'm hoping to spend some time compiling some free content into a github repo soon for a gallery I want to put together. We should collaborate on putting that together, and finding free sources. Related: https://github.com/feross/webtorrent-www/issues/27

feross:
Can you update this PR and remove the reference to PT? I think that's not necessary -- your site has nothing to do with piracy.

v0dy:
Yes! I was thinking about adding archive.org, vodo.net and CCC videos as project butter does but the problem is that there is no one seeding from a webtorrent client yet. 
Also updated PR.

feross:
Yeah -- that's something I want to fix. For all the free content out there, I think we can raise a WebTorrent fund to seed it to web browsers.

Merged.

v0dy:
Yeah that's a good idea, or get archive.org to seed them as they already host most of the video content via http. 

For now i will probably seed all the content on my site from a 100mbit connection, just need to set up a server with the webtorrent-client.

PR #834, opened 2016-06-11
Null check pendingConns

DiegoRBaquero:
Prevent arrayRemove if the pool is already destroyed. Fixes possible error `TypeError: Cannot read property 'indexOf' of null`


feross:
Good catch!

PR #851, opened 2016-07-02
Fix version string

mathiasvr:
`.map` calls zeroFill with the index as `pad` argument, causing incorrect results such as `'1.2.5' -> '0112'`, instead of `'0102'`.


feross:
Good catch! Released as 0.95.3.

PR #852, opened 2016-07-05
Added Squidlink to the list of 'whos using webtorrent'

darkenvy:
Thanks feross for all your hard work. Without it, my project would not have came to fruition. Thanks for letting me add my project to your list as well.


feross:
Nice work!

PR #855, opened 2016-07-07
Updated all deps, except electron, to latest version

amilajack:


amilajack:
Sorry about the naming mistake!

feross:
Most of these changes are not strictly necessary because of how the caret `^` operator works, but this doesn't hurt.

feross:
Thanks for the PR!

PR #871, opened 2016-07-21
Use memory chunk store for unit tests

yciabaud:
The webseed tests were not using fixtures module but passed anyway because the fs store was already populated with the data of a previous test.

This PR uses memory chunk store to ensure the store is empty before running a test.

### DIFF
--- a/test/node/download-tracker-magnet.js
+++ b/test/node/download-tracker-magnet.js
@@ -1,6 +1,7 @@
 var extend = require('xtend')
 var fixtures = require('webtorrent-fixtures')
 var fs = require('fs')
+var MemoryChunkStore = require('memory-chunk-store')
 var series = require('run-series')
 var test = require('tape')
 var TrackerServer = require('bittorrent-tracker/server')
@@ -15,7 +16,7 @@ test('Download using HTTP tracker (via magnet uri)', function (t) {
 })
 
 function magnetDownloadTest (t, serverType) {
-  t.plan(10)
+  t.plan(11)
 
   var tracker = new TrackerServer(
     serverType === 'udp' ? { http: false, ws: false } : { udp: false, ws: false }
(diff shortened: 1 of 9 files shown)


test/node/download-tracker-magnet.js
```diff
@@ -15,7 +16,7 @@ test('Download using HTTP tracker (via magnet uri)', function (t) {
 })
 
 function magnetDownloadTest (t, serverType) {
-  t.plan(10)
+  t.plan(11)
```
yciabaud: This test was not passing with an empty store, I assume this is the good number.

test/node/download-webseed-magnet.js
```diff
@@ -10,7 +11,7 @@ var WebTorrent = require('../../')
 test('Download using webseed (via magnet uri)', function (t) {
   t.plan(9)
 
-  var serve = serveStatic(path.join(__dirname, 'content'))
+  var serve = serveStatic(path.dirname(fixtures.leaves.contentPath))
```
yciabaud: Assets are now in the fixtures module.

yciabaud:
I focused on downloading tests but maybe all the tests are concerned.

test/node/download-tracker-magnet.js
feross: It looks like 10 is the right number, we should figure out why this needed to be changed to 11 to pass with an empty store.

test/node/download-webseed-magnet.js
feross: Good catch!

test/node/download-tracker-magnet.js
feross: Okay, so this was a timing thing. Basically now `noPeers` is getting emitted twice because the tracker gets sent a 'started' then a 'completed' message and both times there are no peers in the response. I changed the `.on('noPeers')` to `.once('noPeers')` so this will be more reliable.

yciabaud: Ok thank you

feross: 👍

PR #874, opened 2016-07-24
Add proxy options to allow proxying tracker and peer connections

### LINKED ISSUE #807: SOCKS Proxy Support
Can you please add SOCKS proxy support? I am working on a crawler that parses metadata, but I have received two fraudulent DMCA requests this week even though I never shared a single byte of the actual files (destroys torrent on metadata). The only way to avoid having my internet shut off while using this software is to push all the traffic through a proxy or VPN so my lSP is hidden. My primary concern is support in node.js, but if it can be supported client side as well that would help people who use proxies in their browsers for similar reasons.

P.S. This is great software, and I intend to add a bounty to this feature request.

yciabaud:
This PR is based on feross/bittorrent-tracker#157 to proxy tracker connnections and adds proxying on TCP peers and webseeds.

WebRTC peers in webtorrent-hybrid is ok now electron-webrtc has been updated.

Options  are structured like :

```
proxyOpts: {
    socksProxy: {
      proxy: {
        ipAddress: 'localhost',
        port: 1080,
        type: 5
      }
    },
    proxyTrackerConnections: true,
    proxyPeerConnections: false,
    httpAgent: {},
    httpsAgent: {}
}
```

Any thoughts?

Fixes #807

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,9 +122,9 @@ function WebTorrent (opts) {
       if (!socksProxy.proxy.type) socksProxy.proxy.type = 5
 
       // Ensure electron-wrtc is used in electron with socks proxy
-      if (self.tracker && socksProxy && self.proxyOpts.proxyPeerConnections &&
+      if (self.tracker && self.proxyOpts.proxyPeerConnections &&
         process && process.versions['electron'] &&
-        (!self.tracker.wrtc || (self.tracker.wrtc && !self.tracker.wrtc.electronDaemon))) {
+        self.tracker.wrtc && !self.tracker.wrtc.electronDaemon) {
         console.warn('You need to provide an electron-wrtc instance in opts.wrtc to use Socks proxy in electron -> WebRTC is disabled')
         self.tracker.wrtc = false
       }
--- a/index.js
+++ b/index.js
@@ -113,6 +113,13 @@ function WebTorrent (opts) {
     self.proxyOpts.proxyTrackerConnections = self.proxyOpts.proxyTrackerConnections !== false
     self.proxyOpts.proxyPeerConnections = self.proxyOpts.proxyPeerConnections !== false
 
+    if (self.proxyOpts.proxyTrackerConnections && !self.tracker.proxyOpts) {
+      if (!self.tracker) {
+        self.tracker = {}
+      }
+      self.tracker.proxyOpts = self.proxyOpts
+    }
+
     if (self.proxyOpts.socksProxy) {
       if (!self.proxyOpts.socksProxy.proxy) self.proxyOpts.socksProxy.proxy = {}
       if (!self.proxyOpts.socksProxy.proxy.type) self.proxyOpts.socksProxy.proxy.type = 5
--- a/index.js
+++ b/index.js
@@ -120,17 +120,24 @@ function WebTorrent (opts) {
       if (!self.proxyOpts.socksProxy.proxy) self.proxyOpts.socksProxy.proxy = {}
       if (!self.proxyOpts.socksProxy.proxy.type) self.proxyOpts.socksProxy.proxy.type = 5
 
+      // Ensure electron-wrtc is used in electron with socks proxy
+      if (self.proxyOpts.socksProxy && self.proxyOpts.proxyPeerConnections &&
+        process && process.versions['electron'] &&
+        self.tracker.wrtc && !self.tracker.wrtc.electronDaemon) {
+        return self.emit('error', 'You need to provide an electron-wrtc instance in opts.wrtc to use socks in electron')
+      }
+
       // Convert proxy opts to electron API in webtorrent-hybrid
       if (self.tracker.wrtc && self.tracker.wrtc.electronDaemon &&
-          self.proxyOpts && self.proxyOpts.socksProxy && self.proxyOpts.proxyPeerConnections) {
-        if (!self.proxyOpts.socksProxy.proxy.authentication && !self.proxyOpts.socksProxy.proxy.userid) {
+          self.proxyOpts.socksProxy && self.proxyOpts.proxyPeerConnections) {
+        if (!self.proxyOpts.socksProxy.proxy.authentication && !self.proxyOpts.socksProxy.proxy.userid && self.proxyOpts.socksProxy.proxy.type === 5) {
           var electronConfig = {
             proxyRules: 'socks' + self.proxyOpts.socksProxy.proxy.type + '://' + self.proxyOpts.socksProxy.proxy.ipAddress + ':' + self.proxyOpts.socksProxy.proxy.port
           }
           self.tracker.wrtc.electronDaemon.eval('window.webContents.session.setProxy(' +
                 JSON.stringify(electronConfig) + ', function(){})', networkSettingsReady)
         } else {
-          self.emit('error', 'SOCKS Proxy authentication is not available in electron-wrtc')
+          self.emit('error', 'SOCKS Proxy must be version 5 with no authentication to work in electron-wrtc')
         }
       } else {
         networkSettingsReady(null)
--- a/index.js
+++ b/index.js
@@ -113,10 +113,7 @@ function WebTorrent (opts) {
     self.proxyOpts.proxyTrackerConnections = self.proxyOpts.proxyTrackerConnections !== false
     self.proxyOpts.proxyPeerConnections = self.proxyOpts.proxyPeerConnections !== false
 
-    if (self.proxyOpts.proxyTrackerConnections && !self.tracker.proxyOpts) {
-      if (!self.tracker) {
-        self.tracker = {}
-      }
+    if (self.tracker && self.proxyOpts.proxyTrackerConnections && !self.tracker.proxyOpts) {
       self.tracker.proxyOpts = self.proxyOpts
     }
 
--- a/index.js
+++ b/index.js
@@ -116,7 +116,7 @@ function WebTorrent (opts) {
       self.tracker.proxyOpts = self.proxyOpts
     }
 
-    var socksProxy = self.proxyOpts.socksProxy;
+    var socksProxy = self.proxyOpts.socksProxy
     if (socksProxy) {
       if (!socksProxy.proxy) socksProxy.proxy = {}
       if (!socksProxy.proxy.type) socksProxy.proxy.type = 5
--- a/index.js
+++ b/index.js
@@ -124,7 +124,8 @@ function WebTorrent (opts) {
       if (self.proxyOpts.socksProxy && self.proxyOpts.proxyPeerConnections &&
         process && process.versions['electron'] &&
         self.tracker.wrtc && !self.tracker.wrtc.electronDaemon) {
-        return self.emit('error', 'You need to provide an electron-wrtc instance in opts.wrtc to use socks in electron')
+        self.emit('error', 'You need to provide an electron-wrtc instance in opts.wrtc to use Socks proxy in electron -> WebRTC is disabled')
+        self.tracker.wrtc = false
       }
 
       // Convert proxy opts to electron API in webtorrent-hybrid
@@ -137,7 +138,9 @@ function WebTorrent (opts) {
           self.tracker.wrtc.electronDaemon.eval('window.webContents.session.setProxy(' +
                 JSON.stringify(electronConfig) + ', function(){})', networkSettingsReady)
         } else {
-          self.emit('error', 'SOCKS Proxy must be version 5 with no authentication to work in electron-wrtc')
+          self.emit('error', 'SOCKS Proxy must be version 5 with no authentication to work in electron-wrtc -> WebRTC is disabled')
+          self.tracker.wrtc = false
+          networkSettingsReady(null)
         }
       } else {
         networkSettingsReady(null)
--- a/index.js
+++ b/index.js
@@ -16,7 +16,6 @@ var parseTorrent = require('parse-torrent')
 var path = require('path')
 var Peer = require('simple-peer')
 var randombytes = require('randombytes')
-var Socks = require('socks')
 var speedometer = require('speedometer')
 var zeroFill = require('zero-fill')
 
@@ -120,18 +119,6 @@ function WebTorrent (opts) {
     if (self.proxyOpts.socksProxy) {
       if (!self.proxyOpts.socksProxy.proxy) self.proxyOpts.socksProxy.proxy = {}
       if (!self.proxyOpts.socksProxy.proxy.type) self.proxyOpts.socksProxy.proxy.type = 5
-
-      // Create HTTP agents from socks proxy if needed
-      if (!self.proxyOpts.httpAgent) {
-        var httpOpts = extend(self.proxyOpts.socksProxy)
-        httpOpts.proxy = extend(opts.proxy)
-        self.proxyOpts.httpAgent = new Socks.Agent(httpOpts, false)
-      }
-      if (!self.proxyOpts.httpsAgent) {
-        var httpsOpts = extend(self.proxyOpts.socksProxy)
-        httpsOpts.proxy = extend(opts.proxy)
-        self.proxyOpts.httpsAgent = new Socks.Agent(httpsOpts, true)
-      }
 
       // Convert proxy opts to electron API in webtorrent-hybrid
       if (self.tracker.wrtc && self.tracker.wrtc.electronDaemon &&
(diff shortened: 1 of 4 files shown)


yciabaud:
This PR is messed up, I will restart it.

yciabaud:
I believe it is ready now, some feedbacks?

kewde:
@yciabaud

Thank you for your work on this.
Exactly what I was looking for :)

Kudos to you.

romaincointepas:
@yciabaud Looks amazingly good! Any idea when this will be merged?

yciabaud:
Well, it seems that @feross is not around these days, I am waiting for some feedbacks before merging feross/bittorrent-tracker#157.

Then we will merge this... (hopefully)

kewde:
Whilst this might not be on topic, proxy support is great for people who wish to run WebTorrent over [I2P](https://geti2p.net/en/).

feross:
@yciabaud I'm around, just really busy these days with other projects. I'll try to take a look soon.

yciabaud:
No problem, this will wait.

romaincointepas:
@yciabaud Testing this right now, aren't we missing something like this to propagate the proxy options to the tracker client?

```js
if (self.proxyOpts.proxyTrackerConnections && !self.tracker.proxyOpts) {
  self.tracker.proxyOpts = self.proxyOpts;
}
```

yciabaud:
@romaincointepas The options should be passed to torrent-discovery moddule here: https://github.com/feross/webtorrent/blob/master/lib/torrent.js#L307

And from there to the tracker: https://github.com/feross/torrent-discovery/blob/master/index.js#L159

Do you have an issue with both this PR and the tracker one?

romaincointepas:
@yciabaud Passing proxy settings via `proxyOpts` when creating a WebTorrent instance with `proxyTrackerConnections` enabled should make the tracker client automatically use those options without having to pass them again in `tracker` no?

At the moment, `proxyTrackerConnections` is simply ignored.

yciabaud:
@romaincointepas yes it should, are you testing with the bittorrent-tracker pr enabled?

romaincointepas:
@yciabaud Yep, using https://github.com/yciabaud/webtorrent/tree/socks2 and https://github.com/yciabaud/bittorrent-tracker/tree/proxy.

Just re-read the changes for both PRs, `proxyOpts.proxyTrackerConnections` is indeed never read/used.

yciabaud:
Oh sorry @romaincointepas I did not get it at the first time. I am adding your snippet, thank you ;)

index.js
```diff
@@ -108,6 +108,13 @@ function WebTorrent (opts) {
     self.proxyOpts.proxyTrackerConnections = self.proxyOpts.proxyTrackerConnections !== false
     self.proxyOpts.proxyPeerConnections = self.proxyOpts.proxyPeerConnections !== false
 
+    if (self.proxyOpts.proxyTrackerConnections && !self.tracker.proxyOpts) {
+      if (!self.tracker) {
```
romaincointepas: self.tracker is already set to an empty object above (see https://github.com/yciabaud/webtorrent/blob/405a084da5406a59acf168460d038b449d26a5a3/index.js#L86), no need for this here?

yciabaud: self.tracker can still be false, but I should skip the forwarding if it is.

romaincointepas:
After some testing, there seems to be a problem with using a single instance of Socks.Agent for all http (or https) requests. When a request is done using the Socks.Agent instance, the instance updates some of its own properties (its `proxy.target` property containing the hostname and port), breaking all requests if multiple requests are done simultaneously with that same instance.

I made it to work by deep-copying the instance before each request, but there may be a better way (can't think of one though).

I've also noticed that the Socks.Agent instance (httpAgent) can end up having `proxy.command: 3` (which is for UDP only), so there may be something going on there as well, I'm investigating (UDP trackers do not seem to be working on my side in a real life scenario with with a mix of http, https and udp trackers).

index.js
romaincointepas: Indeed, good catch!

yciabaud:
You must be right, let's try to clone the options ourselves before passing it to socks.

romaincointepas:
@yciabaud Got a working implementation: using the same instance of Sock.Agent for http and https proxying and Socks.Client modifying proxyOpts was indeed messing things up (not 100% sure about the Socks.Client one though, but it wasn't safe for sure).

If you are busy and if you are okay with it, I would like to take a shot at fixing this.

I also see two minor improvements that could be done, tell me what you think:

1. Remove the ability to pass http(s)Agents. I feel like HTTP proxying, if ever supported (it's rarely used for torrenting AFAIK), should be set the same way as SOCKS (by passing host, port and other options). *For future reference, this looks good: https://github.com/TooTallNate/node-https-proxy-agent.*

2. Not do anything that rely on UDP networking when using SOCKS 4/4a (only UDP trackers at the moment).

yciabaud:
@romaincointepas good you made it work, sure you can push your work on my repo and I will merge it here.

Otherwise I can do it, tell me how you did it? Simply cloning the options?

For the improvements, I would rather keep the http agents to avoid having a new dependency in the project. I wanted to keep socks outside too but it was to complicated to handle.

I can do the improvement 2. by checking on socks version for UDP.

Tell me if you can manage to add your code to the PR.

yciabaud:
OK @romaincointepas I made some changes in bittorrent-tracker: 

- Check for socks version: https://github.com/feross/bittorrent-tracker/pull/157/commits/55fa45a544532ade0d98546dde4429ea645cf7a3
- Clone options: https://github.com/feross/bittorrent-tracker/pull/157/commits/c1e9640f710c59c9de53eea22948e89452fc7af9

Is it what you had in mind? I can do the same here too...

romaincointepas:
@yciabaud Great!

Using https://www.npmjs.com/package/clone is cleaner as it offers deep-cloning of an object and will be needed for cloning the Socks.Agent "master" instances. It's a very light and commonly used dependencies, I think it's okay.

About the agents, I think it shouldn't be passed as options in proxyOpts (for both Webtorrent and the tracker client) but instead be generated as *private property* when instantiating the Webtorrent and tracker client (as _http(s)Agent), see my reasoning in my previous comment about HTTP proxying. Then we can clone it when needed in http-trackers.js, webconn.js, etc.

Also, I didn't get your point about httpAgent and adding a dependency, I assume it's a misunderstanding?

About disabling some protocols/features, WebRTC should also be disabled if SOCKS has userid or authentication (forgot about that one :)).

yciabaud:
@romaincointepas are you saying you need to clone agents? Cloning options is not enough?

can you test the last commits I made?

romaincointepas:
@yciabaud Yep, an agent instance should be unique to a protocol/hostname/port combination (because the instance itself on createConnection will modify its property `target`). I find it quite clean to keep the *master* http(s)Agent instances we have right now, and clone it when needed (instead of recreating it by cloning options).

For example, in `http-tracker.js`:
```js
var agent
  if (self.client._proxyOpts) {
    agent = parsedUrl.protocol === 'https:'
      ? self.client._proxyOpts.httpsAgent : self.client._proxyOpts.httpAgent
  }
```

would become (using the `clone` module):
```js
var agent
  if (self.client._proxyOpts) {
    agent = clone(parsedUrl.protocol === 'https:'
      ? self.client._proxyOpts.httpsAgent : self.client._proxyOpts.httpAgent)
  }
```
And self.client._proxyOpts.httpsAgent will never actually be used directly (meaning it really shouldn't, otherwise it will break everything).

yciabaud:
Maybe we should have a factory to generate the agents, or make the socks agent support multiple connection...

I am not sure what to do here... Cloning agents seems weird to me.

romaincointepas:
@yciabaud You could also generate agents on-the-fly and remove the current http(s)Agent logic from your code.

```js
var agent
if (self.client._proxyOpts) {
  agent = new Socks.Agent({
    proxy: clone(self.client._proxyOpts.socksProxy.proxy)
  }, (parsedUrl.protocol === 'https:'));
}
```

yciabaud:
This looks better to me, I am removing agent generation from webtorrent and do like you said in bittorrent-tracker.

yciabaud:
Pushed agent generation on bittorrent-tracker: https://github.com/feross/bittorrent-tracker/pull/157/commits/8747f7edc0fab5db7df44ea3754bb342303169e3

And removed generation from here.

I don't think cloning proxy is required so I just copy the options object (where target is modified)

Can you tell me if it is working with multiple trackers? (Maybe we sould share a test case)

romaincointepas:
@yciabaud Good! Don't forget`webconn.js` as well in `webtorrent`.

I really don't like using `extend` to deep-copy an object with depth > 1 though, even if we know `authentication` is never modified, it's a recipe for future bugs. But your call 😄

There is no easy test case, but trying a *popular* torrent and adding a lot of public trackers (https://github.com/ngosang/trackerslist/blob/master/trackers_all.txt) and checking how many succeed/fail compared to the same setup without proxyOpts is a good way to know if something is broken.

I will test this right now and report here.

yciabaud:
Nice catch @romaincointepas , I ve added agent generation on webconns too.

I agree with you regarding deepcopies, clone is already in the dependency tree anyway so I use it now.

Regarding http(s)Agent options, I like the idea to have an extension point here, http proxy support would be nice and the agents can be reused most of the time.

I am thinking about keeping these options and report a bug to the socks module.

What do you think?

romaincointepas:
`clone` was not in the dependency tree? (but 👍 for using it, it's pretty standard stuff).

About agents tied to host/port, It's not really a bug though as node http agents behaves the same way, or at least used to.

I also think HTTP proxying should be done later if needed and be built the same way SOCKS is (to be honest, I think this will never happen, HTTP proxy for torrenting is not really useful). Not having httpAgents keep the code simpler and clearer IMO.

Also, you have a bug, you replaced extend by clone in a place where you shouldn't have:
https://github.com/feross/webtorrent/pull/874/commits/124bfc930944ff827cbc370f3a386f5f5fd7bb74#diff-65b0a569f99e647ae33e8c31b9bf72baR1664

romaincointepas:
@yciabaud Looks good! What happens if:
- in electron
- proxyPeerConnections == true
- using SOCKS proxy
- Webtorrent opts.tracker is undefined

It seems that right now it will not return an error but instead use WebRTC while ignoring the proxy. (self.tracker being undefined should trigger the error in https://github.com/feross/webtorrent/pull/874/commits/158330896a76d30b787d9e71f48a31e89ee3e29f#diff-168726dbe96b3ce427e7fedce31bb0bcR126?)

Same if SOCKS < 5, or SOCKS has userid or authentication, it seems that it doesn't disable WebRTC.

yciabaud:
@romaincointepas it should emit an error, maybe throwing an exception would be better here...
Are you listnening to the error event?

romaincointepas:
@yciabaud it does emit an error but doesn't stop WebRTC from working and ignoring the proxy settings.

The question is: should using a SOCKS proxy with userid/auth or SOCKS 4 disable some capabilities (WebRTC only at the moment) or let those capabilities bypass the proxy?

yciabaud:
@romaincointepas thank you very much for your help!

I believe too that we should just stop webrtc when the proxy cannot be used. Here is a commit in this way.

romaincointepas:
- [] Replace https://github.com/yciabaud/webtorrent/blob/2a410fdd0f06bbe97269a192ceaaad2cc4214f44/index.js#L139-L140 with:
```js
self.tracker.wrtc.electronDaemon.eval('require("electron").remote.getCurrentWebContents().session.setProxy(' +
    JSON.stringify(electronConfig) + ', function(){})', networkSettingsReady)
```

yciabaud:
Ok i see why the second issue is occurring, my commit on simple-websocket is excluding electron... https://github.com/feross/simple-websocket/commit/fcabf152358147000f98eaea5b0dfc74103a3382

yciabaud:
@romaincointepas can you test it with this revision? I used the electron-eval option to execute in the main process instead of pass it to the window and then use remote to come back in the main process.

I am looking forward to fix the second issue.

romaincointepas:
@yciabaud Works!

index.js
```diff
@@ -106,51 +106,106 @@ function WebTorrent (opts) {
...
+    var socksProxy = self.proxyOpts.socksProxy
+    if (socksProxy) {
+      if (!socksProxy.proxy) socksProxy.proxy = {}
+      if (!socksProxy.proxy.type) socksProxy.proxy.type = 5
+
+      // Ensure electron-wrtc is used in electron with socks proxy
+      if (self.tracker && socksProxy && self.proxyOpts.proxyPeerConnections &&
+        process && process.versions['electron'] &&
+        (!self.tracker.wrtc || (self.tracker.wrtc && !self.tracker.wrtc.electronDaemon))) {
+        console.warn('You need to provide an electron-wrtc instance in opts.wrtc to use Socks proxy in electron -> WebRTC is disabled')
```
romaincointepas: I don't think this warning should be thrown if `self.tracker.wrtc === false`.

yciabaud: You are right

romaincointepas:
@yciabaud after some more testing, I've found something weird.

In `torrent.js`, [the `onGotSocket` function](https://github.com/yciabaud/webtorrent/blob/04399a6759aae830b893c670a649ebdfa17e59da/lib/torrent.js#L1675) is being called with one of the 3 following errors 99% of the time:

`torrent.js:1682 Error: Socket Closed`
`torrent.js:1682 Error: Negotiation Error (5)`
`torrent.js:1682 Error: Negotiation Error (4)`

The result is very low download speed (amazingly not 0) with peers that seem to connect then immediately drop (the peer count goes from around 55 to 0 then back every 5 seconds).

Note:
- I've tested with many different "torrent-optimized" SOCKS5 proxy (PIA, torguard, btguard), always the same problem.
- I've tested those same proxy in the same setup with another torrent client (qBittorrent), works perfectly.
- I've tested with master webtorrent and your fork without `proxyOpts` set, all good.
- I've also tested the 2 above cases with a running torrent-friendly VPN on the side, it also works perfectly (tried this as I was assuming it was a port/networking problem).

romaincointepas:
DHT doesn't seem to be currently proxied, is that correct?

(After taking a quick look, it seems that it should be done in https://github.com/mafintosh/k-rpc-socket/blob/master/index.js)

yciabaud:
This is correct, I have not played around with dht... yet

romaincointepas:
Can you also confirm that right now, when using a proxy, the client is not directly reachable by peers (meaning upload can only happens with peers that the client connects to)? (the listening port the client is identified by on the trackers, DHT and PEX can't be reached as it's the local listening port, whereas the IP is the proxy IP).

I assume most of the work involved would be with `tcp-pool.js` (by using SOCKS BIND instead of net.createServer(), announce that new proxy tcp port to trackers/DHT/PEX, and listen to new connections from that socket).

@yciabaud Am I on the right path? Do you intend to support this?

yciabaud:
I agree with you @romaincointepas , Peer connexions and DHT/PEX should work with Socks BIND. I don't think we can open only one connection in this mode, do you have more information about this?

I wanted to merge this PR before going deep in inboud proxying but this is taking so loong!

@feross please can you have a look at this? I have worked hard to make it work, can you review the tracker part? feross/bittorrent-tracker#157 I don't want to waste more time if you believe I am in the wrong way.

JoshGlazebrook:
@yciabaud v2 of 'socks' is now available. Not sure if you're up for taking this on again, but usage of BIND/ASSOCIATE is now more sane than the prior version.

yciabaud:
Thanks @JoshGlazebrook for this information, unfortunately I worked a lot on this one year ago and never got any feedback from the team so I left it.

I can still update my PRs if it gets some interest but I don't want to waste more time on this.

romaincointepas:
@yciabaud I would definitely be interested in an updated PR (assuming it's not too much work on your side).

kewde:
@yciabaud I think there are quite a few silent appreciators here..
:+1:

Fenny:
@yciabaud Please don't give up! Trust me, this is the only thing webtorrent really needs at this point!

brian123zx:
FWIW, I've been watching this pr for years now too 😄 Sad to see contribution die because @feross seems mia. Is there an active fork?

feross:
We're slowly trying to catch up on backlogged PRs and issues. It's a lot of work! Sorry for the delays everyone.

SahAssar:
Was there any specific reason this was not merged?

Issue #876, opened 2016-07-28
Documentation Request: Embedded Webtorrent HOW-TO (as found on webtorrent.io)


jrnolan2525:
Hello & thank you for such an exciting technology. 

How would one demonstrate a single-page setup as found on https://webtorrent.io's embedded stream, including those lovely progress & speed indicators? I've used the Quick Example's 'Downloading a torrent (in the browser)' as found on https://webtorrent.io/intro but I would be interested in seeing the dynamic resize as demostrated.

My skill level: QBasic is still vogue, yes? That said, I did manage to get Node.js & NPM installed though I'm using _webtorrent.min.js_ as preference for dependency mitigation.

TIA.

yciabaud:
Hello jrnolan2525,

I have set a [codepen](http://codepen.io/yciabaud/pen/XdOeWM) up for you with some of the styles and the status.

Basically, you need to trigger a function to refresh the status using the torrent API. In the example, a onProgress function is triggered periodically after the ontorrent event and the HTML elements are refreshed with `torrent.progress`, `torrent.numPeers`, `torrent.downloaded`, `torrent.length` and `torrent.timeRemaining`.

Is that what you expected?

jrnolan2525:
Bonjour Mon Ami -

I am gobsmacked. Stunning; absolutely _stunning_ how elegant & impactful of your work. To say the very least, it brings clarity to many of the points of confusion I'm grappling.

I assume it's safe to delete these entries fr codepen:

- 
- 
- 
- 
- <!-- script src='//assets.codepen.io/assets/common/stopExecutionOnTimeout-53beeb1a007ec32040abaf4c9385ebfc.js'></script>

Please clarify if not it as appears to function fine without them in Chrome 49.0.2623.87 (64-bit).

Now seeing how to use _client.add_ , which the existing documentation states will also accept a magnet URI, I assume the same magnet URI must be reflected in the HTML's _torrentLink_ anchor reference? My needs dictate magnet URIs over .torrent files.

I think I'm going to need another coffee break to get over the shock of this spectacular result!

yciabaud:
Oh that is codepen stuff, all you really need besides the JS code showed in the window is the webtorrent and moment.js dependencies. (moment.js is used to print human readable remaining time)

In the head:
```
<script src="https://cdn.jsdelivr.net/webtorrent/latest/webtorrent.min.js"></script>
<script src="http://momentjs.com/downloads/moment.min.js"></script>
```

[...]

PR #877, opened 2016-07-30
Add an HTML example showing status using the torrent API

yciabaud:
May be helping some users willing to use the API outside of node / browserify like in the issue #876.


feross:
Thanks, nice!

Issue #168, opened 2014-10-28
Publish a BEP explaining how to add support for WebTorrent


kumavis:
from the readme:

```
We hope established torrent clients (uTorrent, Transmission, Vuze, etc.) will add support for WebRTC and the WebTorrent extensions so they can swarm with peers from both the normal and web networks.
```

Is there an existing BEP for this or should we create one?

tghosgor:
I think it is much better to have a standalone separate solution to this first. Implementing the new extension wide-spread would be long away IMO. I think a browser plugin will be a better and necessary solution for the short-term.

Is there any ongoing browser plugin projects to be a shim layer between webtorrent swarm and bittorrent swarm?

eliooses:
+1

wI2L:
@feross Any news about this?

frandavid100:
I'm a Transmission user and I'd also like to know if theres any news regarding a BEP that Transmission can adopt.

bbarrows:
Sorry, this is a pessimistic comment. I saw an email about this thread and I was thinking that one big reason that BitTorrent works is because people are giving as they are receiving right?

But to connect the WebTorrent and "regular" BitTorrent networks you would basically be asking people to proxy other peoples BitTorrenting connections through their networks. They are not getting anything in return for allowing this. Also, what if someone downloads something illegal through on of these "proxies" and the proxy owner gets in trouble for it (as their IP would be the one exposed).

Obviously it would be very attractive to connect WebTorrent with the TCP/UDP BitTorrent network but because the only option available for peer to peer connections in a browser (minus Chrome Extensions I believe) is WebRTC I just do not see any way to connect the two networks without a proxy. But maybe I am missing something?

I guess you might find some people that would be willing to run these proxy servers to connect WebRTC to TCP/UDP but they would probably end up dealing with the same issue TOR Exit node owners have to deal with.

yciabaud:
Hi @bbarrows ,
We want some clients to add webtorrent support for communicating with webrtc peers directly, not using them to proxy anything. They will be able to share their content with both networks acting as a normal peer. The client has control about what it shares, he makes the bridge between the two networks because he is able to send blocks of he torrents he is owning on both sides.

With this solution, the whole network is not reachable from the web but we are sticking with the original architecture of bittorrent, sharing the content we need with peers on both sides.

Does it answers your concerns or have I misunderstood something?

bbarrows:
[...]

PR #881, opened 2016-08-08
Publish a BEP explaining how to add support for WebTorrent

yciabaud:
This PR proposes a BEP document as discussed in #168

I would like to have some feedback on the content and on some changes I made in the messages.
I changed some properties to have a cleaner protocol, if we keep it, we will have to update bittorrent-tracker to fit with this document.

You can view the html output on github: https://github.com/yciabaud/webtorrent/blob/486a94d2b651e7aed86ba27b82e679f5d1d1e700/bep_webrtc.rst

@feross @DiegoRBaquero @bbarrows

### DIFF
--- a/bep_webrtc.rst
+++ b/bep_webrtc.rst
@@ -204,7 +204,8 @@ If the tracker encounters an error, it might send an error message.
 error response::
 
   {
-    "error": ""
+    "action":   "error"
+    "message":  ""
   }
 
 
--- a/bep_webrtc.rst
+++ b/bep_webrtc.rst
@@ -1,0 +1,239 @@
+:BEP: XXX
+:Title: Websocket tracker protocol for WebRTC transport
+:Version: $Revision$
+:Last-Modified: $Date$
+:Author:  Yoann Ciabaud <y.ciabaud@gmail.com>
+:Status:  Draft
+:Type:    Standards Track
+:Content-Type: text/x-rst
+:Created: 02-Aug-2016
+:Post-History:
+
+
+Abstract
+========
+Webapps are more and more used to publish and consume content over the
+internet, Bittorrent could be used to optimize the server and network resources
+needed to distribute files but the need for a client software prevents its
+usage in a browser environment.
+
+This BEP documents how to add support for WebRTC signalling on a tracker in
+order to make the webapps able to connect peers using the WebRTC API [1].
+Once a communication channel is open between the peers using WebRTC, the
+standard Bittorent peer connection protocol is used.
+
+Due to the WebRTC signalling process, the tracker needs to push messages to the
+clients with a full-duplex communication channel available in a web browser.
+At the moment of this BEP, Websockets [2] are the only compatible standard
+technology.
+
+WebRTC signalling process
+=========================
+The diagram below explains the messages involved in the WebRTC signalling
+process.
+
+::
+
+  peer 1                 tracker                  peer 2
+
+    |                       |                       |
+    |                       |                       |
+    | announce start        |                       |
+    |   numwant=x           |                       |
+    |   x SDP offers        |                       |
+    | >-------------------> |                       |
+    |                       |                       |
+    |                       |  SDP offer            |
+    |                       | >-------------------> |
+    |                       |                       |
+    |                       |  SDP answer           |
+    |                       | <-------------------< |
+    |                       |                       |
+    |  SDP answer           |                       |
+    | <-------------------< |                       |
+    |                       |                       | connection established
+   .|.. .. .. .. .. .. .. .. .. .. .. .. .. .. .. ..|.. .
+    |                                               |
+    V                                               V
+
+Websocket tracker protocol
+==========================
+
+credits
+-------
+The websocket tracker protocol has been designed for the Webtorrent [3] project
+by Feross Aboukhadijeh and hundreds of open source contributors.
+
+overview
+--------
+The websocket tracker uses JSON payloads reflecting the HTTP request parameters
+and an additionnal action property used to switch between announce and other
+actions (ex. scrape). If the announce URL of the torrent contains the ws or wss
+protocol, the client establishes a websocket connection with the tracker.
+
+WebRTC offers and answers can be provided as an extension of the announce
+message, the tracker will be responsible to forward them between the peers to
+act as a signalling service.
+
+After the connection is open, the client can begin announcing itself by using
+the JSON messages below.
+
+signalling related message format
+---------------------------------
+announce request::
+
+  {
+    "action":     "announce",
+    "info_hash":  "",
+    "peer_id":    ""
+    "numwant":    0,
+    "uploaded":   0,
+    "downloaded": 0,
+    "left":       0,
+    "event":      "",
+    "offers":     [
+      {
+        "offer_id": "",
+        "offer": ""
+      },
+      ...
+    ]
+  }
+
+A client can provide SDP offers that the tracker will forward to peers to
+establish a connection.
+"offers" is a JSON array containing a list of WebRTC SDP offers and a generated
+"offer_id" used to match peers and offers.
+
+
+announce response::
+
+  {
+    "action":     "announce",
+    "complete":   0,
+    "incomplete": 0,
+    "interval":   0,
+    "info_hash":  ""
+  }
+
+
+offer or answer message::
+
+  {
+    "action":     "announce",
+    "info_hash":  "",
+    "offer_id":   "",
+    "peer_id":    "",
+    "sdp":      ""
+  }
+
+The tracker forwards the offer or the answer to clients in an announce message
+with an "sdp" property.
+
+
+answer message::
+
+  {
+    "action":     "announce",
+    "info_hash":  "",
+    "offer_id":   "",
+    "peer_id":    "",
+    "to_peer_id":    "",
+    "answer":     ""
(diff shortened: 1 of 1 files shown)


bep_webrtc.rst
```diff
@@ -0,0 +1,241 @@
...
+    "action":     "scrape",
+    "info_hash":  ""
+  }
+
+
+scrape response::
+
+  {
+    "announce": "",
+    "info_hash": "",
```
yciabaud: I changed the scrape response format from infoHash to info_hash to be consistent with the other messages.

bep_webrtc.rst
```diff
@@ -0,0 +1,241 @@
...
+
+multi-scrape request::
+
+  {
+    "action":     "scrape",
+    "info_hash":  ["ih1", "ih2", ...]
+  }
+
+
+multi-scrape response::
```
yciabaud: Same thing for the multiscrape...

bep_webrtc.rst
```diff
@@ -0,0 +1,241 @@
...
+    }
+  }
+
+
+If the tracker encounters an error, it might send an error message.
+
+error response::
+
+  {
+    "error": ""
```
yciabaud: Here I changed from "error message" to "error" to be cleaner.

bep_webrtc.rst
```diff
@@ -0,0 +1,241 @@
...
+
+
+offer or answer message::
+
+  {
+    "action":     "announce",
+    "info_hash":  "",
+    "offer_id":   "",
+    "peer_id":    "",
+    "sdp":      ""
```
yciabaud: I changed from answer to sdp as this property can be either an offer or a response.

bep_webrtc.rst
```diff
@@ -0,0 +1,239 @@
...
+
+  {
+    "error": ""
+  }
+
+
+Existing implementations
+========================
+
+bittorrent-tracker [4] support this protocol, it is used in all WebTorrent [2]
```
alxhotel: Typo: `WebTorrent [3]`

lmatteis:
I'm a bit confused. So trackers supporting this would be able to track IP addresses of "normal" clients and also WebRTC contacts? That's cool but if WebRTC tabs can't connect to current BitTorrent client nodes (because most torrent clients don't support WebRTC), what's the point?

yciabaud:
The point of this BEP is to document the process of signalling webrtc peers on a tracker to connect each other.

Aside from the tracker implementation, the clients supporting this would become hybrid clients and could use webrtc to connect to browser clients.

This would help a lot to make bittorrent content available in webtorrent network.

Do you understand now? Do you have an idea to make it clearer for a reader to understand what we are aiming?

yciabaud:
The point is to make clients support webrtc by giving them a standard documentation ;)

DiegoRBaquero:
Just read it all. It's very technical to the point of a BEP. Is there any place where I can see the guideline for a BEP?

Looking good @yciabaud

yciabaud:
This page details the BEP process: http://www.bittorrent.org/beps/bep_0001.html, some other guidelines can be found in http://www.bittorrent.org/beps/bep_0002.html

austlane:
@yciabaud Thanks so much for writing this up. Looks detailed and well done!
I've been following this project for a number of years, hoping feross would do this sooner rather than later.

yciabaud:
Thank you very much @austlane, lets hope we will push that to bittorrent soon!

bep_webrtc.rst
alxhotel: Looking at the [UDP tracker BEP](http://www.bittorrent.org/beps/bep_0015.html), I've found it uses a type of `action` called `error` whenever an error is return. And then a `message` attribute for the error message.

It could make sense adding it (at least the `action` attribute), just to be consistent in the response.

yciabaud: I agree with you, it is definitely more consistent with the other messages. Pushing it right now!

bep_webrtc.rst
```diff
@@ -0,0 +1,242 @@
...
+
+
+If the tracker encounters an error, it might send an error message.
+
+error response::
+
+  {
+    "action":   "error"
+    "message":  ""
+  }
```
yciabaud: I changed the error response message for it to look like the other messages like in other beps.

wI2L:
@yciabaud Just to clarify, is this BEP based on the actual implementation ? I couldn't find the error message you describe in the source code of the bittorrent-tracker project, instead it send back a `failure reason` key.

yciabaud:
@wI2L I made some changes that I have commented inline in this PR. You are right, ATM the errors are labelled with `failure reason`.

I changed a bit some of the payloads too, all changes are inline commented.

bep_webrtc.rst
wI2L: Tbh, I think that both `failure reason` and an extra error field are necesseray. The first one wpuld stay for legacy reasons and would be used only for Bittorrent protocol related failures (either announce or scrape) while another extra field (or object, with code/message ?) could describe a signalling error.

bep_webrtc.rst
wI2L: I am thinking that a message like this one has nothing related to Bittorrent annoucing, and that maybe we could use two new `action` events `offer` and `answer`.

wI2L:
@yciabaud Ok. I have added comments on some.
While reading the source of bittorrent-tracker I have noticed the use of `common.hexToBinary` (for `info_hash`) and I am not sure of how to interpret that. Like I have mentionned it in our mail exchanges, the "type" of each field could be described.
Also, the websocket protocol describes binary and text messages. it may be necessary to define what kind of messages are supposed to be sent and read ?

bep_webrtc.rst
```diff
@@ -0,0 +1,242 @@
...
+
+answer message::
+
+  {
+    "action":     "announce",
+    "info_hash":  "",
+    "offer_id":   "",
+    "peer_id":    "",
+    "to_peer_id":    "",
+    "answer":     ""
```
wI2L: `answer` is an object with the fields `type` and `sdp`.

yciabaud:
Thanks @wI2L for these reviews, I agree that the types of the fields must be added to the BEP. 
I will try to find a moment to work again on this.

wI2L:
np @yciabaud 
Also, I was wondering if the BEP should indicates how a BitTorrent tracker answer to a Webclient before the connection is actually upgraded to the Websocket protocol, in case of errors.
A standard tracker send back a bencoded dict with a `failure reason` key, with status code 200.

bep_webrtc.rst
```diff
@@ -0,0 +1,239 @@
...
+credits
+-------
+The websocket tracker protocol has been designed for the Webtorrent [3] project
+by Feross Aboukhadijeh and hundreds of open source contributors.
+
+overview
+--------
+The websocket tracker uses JSON payloads reflecting the HTTP request parameters
+and an additionnal action property used to switch between announce and other
+actions (ex. scrape). If the announce URL of the torrent contains the ws or wss
```
Miserlou: I STRONGLY recommend that we formally drop support for insecure websockets. It's almost 2017, there is no reason to introduce a potential security vulnerability or vector for surveillance. Especially with Let's Encrypt available for free, there is simply no excuse for running an insecure public service. WebRTC itself already disallows insecure connections, we should incorporate that momentum into this BEP.

I suggest we change this explicitly disallow 'ws' and only allow 'wss' here.

bep_webrtc.rst
```diff
@@ -0,0 +1,242 @@
...
+    "action":     "scrape",
+    "info_hash":  ["ih1", "ih2", ...]
+  }
+
+
+multi-scrape response::
+
+  {
+    "ih1": {
+      "announce": "",
```
wI2L: According to https://github.com/feross/bittorrent-tracker/blob/master/server.js#L691 and https://github.com/feross/bittorrent-tracker/blob/master/server.js#L697, there is no shuch field `announce`.

bep_webrtc.rst
```diff
@@ -0,0 +1,242 @@
...
+  {
+    "action":     "scrape",
+    "info_hash":  ""
+  }
+
+
+scrape response::
+
+  {
+    "announce": "",
```
wI2L: According to https://github.com/feross/bittorrent-tracker/blob/master/server.js#L677-L680, there is no such field `announce`.

bep_webrtc.rst
wI2L: Actually I "think" (still not sure) that the response is created here : https://github.com/feross/bittorrent-tracker/blob/master/server.js#L698-L701, and that the "infoHash" field is used as a key in the forEachloop when creating a new entry in files list.

DiegoRBaquero:
Is there anything missing so this can be sent?

wI2L:
@DiegoRBaquero Depends, would you like with @feross  and @yciabaud  to address my proposals I have made in the PR here: https://github.com/yciabaud/webtorrent/pull/1
There is two ideas, publish a document that describes the actual state of the protocol, which could be done by finishing/polishing the original draft of @yciabaud or discuss furthermore about the changes I intoduced in favor of a more comprehensive protocol.

rom1504:
Why is this blocked ? It seems important for webtorrent and isn't most of the work already done ?

feross:
Unlocking and re-opening this.

PR #901, opened 2016-09-04
Bugfixes for server.close() and server.destroy()

dcposch:
@feross @mikeal

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -21,13 +21,13 @@ function Server (torrent, opts) {
   var _close = server.close
   server.close = function (cb) {
     closed = true
-    torrent = null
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -37,6 +37,7 @@ function Server (torrent, opts) {
     })
 
     // Only call `server.close` if user has not called it already
+    if (!cb) cb = function () {}
```
feross: It's best to set all defaults at the top of the function instead of in the body.

feross:
LGTM -- thanks for catching this.

feross:
0.96.5

PR #908, opened 2016-09-13
Add new FAQ: "Does WebTorrent leak your IP address?"

feross:
Inspired by the thread here: https://www.reddit.com/r/WebTorrent/comments/52f3ns/since_webtorrent_uses_webrtc_does_it_leak_your_ip/

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -351,6 +351,35 @@ would be around 10-20 seconds behind the live stream.
 This approach can definitely be improved, though! Why not give that a shot yourself
 and share the code?
 
+## Does WebTorrent leak your IP address when using a VPN? I heard that WebRTC leaks your IP address.
+
+No.
+
+WebRTC data channels do not allow a website to discover your public IP address when
+there is a VPN in use. The WebRTC discovery process will just find your VPN's IP
+address and the local network IP address.
+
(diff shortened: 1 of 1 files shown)


dcposch:
WebRTC doesn't leak your IP if you're using a VPN, but it **does** leak your IP if you're using the browser proxy settings, right?

feross:
> WebRTC doesn't leak your IP if you're using a VPN, but it does leak your IP if you're using a browser proxy, right?

I've never heard of this being the case. Link?

DiegoRBaquero:
@dcposch @feross It shouldn't. See https://bugs.chromium.org/p/chromium/issues/detail?id=497040

dcposch:
@DiegoRBaquero thanks. I remember seeing that bug. I think Firefox has the same issue:
https://bugzilla.mozilla.org/show_bug.cgi?id=959893

Since it's an open bug in both Chrome and Firefox, we should mention that browser proxies won't help for now. You need a full VPN if you want to location/IP privacy.

DiegoRBaquero:
@dcposch In Chrome it's actually fixed.

feross:
Great, thanks for the investigation and discussion. Merging this now!

PR #909, opened 2016-09-13
Cross-origin URL redirection workaround

jablko:
Browsers allow HTTP redirects for simple cross-origin requests but not for requests that require preflight.
Because they set the Range header, this includes all web seed requests (if they're cross-origin).
So currently, web seeds can be cross-origin or they can make redirects, but not both at the same time.
What do you think of the following workaround for this?

If a request fails, maybe it's because the web seed sent a redirect and the browser objected, so try a simple request.
Make a simple request to unwind any redirects and get the final URL. (I used a HEAD request for this.)
If the final URL doesn't match the original URL, retry our original request with the new URL.
I make a feeble attempt to check that this is a cross-origin request.
It catches all cross-origin cases, but matches a few same-origin cases too. I figure that's okay in the interest of simplicity, since the check is merely an optimization anyway.

### DIFF
--- a/lib/webconn.js
+++ b/lib/webconn.js
@@ -124,12 +124,7 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
         range: 'bytes=' + start + '-' + end
       }
     }
-    get.concat(opts, function (err, res, data) {
-      if (hasError) return
-      if (err) {
-        hasError = true
-        return cb(err)
-      }
+    function onResponse (res, data) {
       if (res.statusCode < 200 || res.statusCode >= 300) {
         hasError = true
         return cb(new Error('Unexpected HTTP status code ' + res.statusCode))
@@ -147,6 +142,51 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
           cb(null, ret)
         }
       }
+    }
+    get.concat(opts, function (err, res, data) {
+      if (hasError) return
+      if (err) {
+        // Browsers allow HTTP redirects for simple cross-origin
+        // requests but not for requests that require preflight.
+        // Use a simple request to unravel any redirects and get the
+        // final URL.  Retry the original request with the new URL if
+        // it's different.
+        //
+        // This test is imperfect but it's simple and good for common
+        // cases.  It catches all cross-origin cases but matches a few
+        // same-origin cases too.
+        if (typeof window === 'undefined' || url.startsWith(window.location.origin + '/')) {
+          hasError = true
+          return cb(err)
+        }
+
+        return get.head(url, function (errHead, res) {
+          if (hasError) return
+          if (errHead) {
+            hasError = true
+            return cb(errHead)
+          }
+          if (res.statusCode < 200 || res.statusCode >= 300) {
+            hasError = true
+            return cb(new Error('Unexpected HTTP status code ' + res.statusCode))
+          }
+          if (res.url === url) {
+            hasError = true
+            return cb(err)
+          }
+
+          opts.url = res.url
+          get.concat(opts, function (err, res, data) {
+            if (hasError) return
+            if (err) {
+              hasError = true
+              return cb(err)
+            }
+            onResponse(res, data)
+          })
+        })
+      }
+      onResponse(res, data)
     })
   })
 }


lib/webconn.js
```diff
@@ -147,6 +142,51 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
...
+        // Use a simple request to get the final URL after redirection
+        // and retry the original request with this URL if they're
+        // different.
+        //
+        // This same-origin test is imperfect, but it's simple,
+        // good for common cases, and is only ever falsely negative,
+        // never falsely positive.
+        //
+        // We can't use simple-get here because http-browserify doesn't
+        // expose the final URL.
```
josephfrazier: I came back to this, and I now think we actually can use `simple-get` here. `browserify` replaced `http-browserify` with `stream-http` (see the list of shims [here](https://github.com/substack/node-browserify#compatibility)), which _does_ expose `res.url` (see https://github.com/jhiesey/stream-http/pull/40).

Here's a proof of concept that you can browserify and run in the console at https://example.com:

```js
var get = require('simple-get')

function getPreflightRedirect(opts, cb) {
  return get(opts, function (err, res, data) {
    if (opts.url.startsWith(window.location.origin + '/')) {
      return cb(err)
    }

    return get({ url: opts.url, method: 'HEAD' }, function (err, res, data) {
      if (err) return cb(err)
      if (res.url === opts.url) return cb(Exception('url did not redirect'))

      opts.url = res.url
      return get(opts, cb)
    })
  })
}

var opts = {
  url: 'https://httpbin.org/redirect-to?url=https%3A%2F%2Fhttpbin.org%2F',
  headers: {
    Range: 'bytes=0-'
  }
}
getPreflightRedirect(opts, function (err, res, data) {
  if (err) {
    console.error(err)
  }
  else {
    console.log("final url: " + res.url)
  }
})
```

jablko: Oh, excellent! I switched to using simple-get.

lib/webconn.js
```diff
@@ -147,6 +142,51 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
...
+
+            hasError = true
+            return cb(err)
+          })
+
+          return
+        }
+
+        hasError = true
+        return cb(err)
```
jablko: Done.

lib/webconn.js
```diff
@@ -147,6 +142,51 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
...
+                  return cb(err)
+                }
+                onResponse(res, data)
+              })
+
+              return
+            }
+
+            hasError = true
+            return cb(err)
```
jablko: Done.

lib/webconn.js
```diff
@@ -147,6 +142,51 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
...
+        // good for common cases, and is only ever falsely negative,
+        // never falsely positive.
+        //
+        // We can't use simple-get here because http-browserify doesn't
+        // expose the final URL.
+        // fetch() or XMLHttpRequest.responseUrl are both candidates.
+        if (typeof window !== 'undefined' && typeof window.fetch === 'function' && !url.startsWith(window.location.origin + '/')) {
+          window.fetch(url, { method: 'HEAD' }).then(function (res) {
+            if (hasError) return
+            if (res.url !== url) {
```
jablko: Good point. Fixed.

lib/webconn.js
```diff
@@ -147,6 +142,51 @@ WebConn.prototype.httpRequest = function (pieceIndex, offset, length, cb) {
...
+        //
+        // This same-origin test is imperfect, but it's simple,
+        // good for common cases, and is only ever falsely negative,
+        // never falsely positive.
+        //
+        // We can't use simple-get here because http-browserify doesn't
+        // expose the final URL.
+        // fetch() or XMLHttpRequest.responseUrl are both candidates.
+        if (typeof window !== 'undefined' && typeof window.fetch === 'function' && !url.startsWith(window.location.origin + '/')) {
+          window.fetch(url, { method: 'HEAD' }).then(function (res) {
```
jablko: Oops. Yes. Fixed. Thanks!

jablko:
Thanks a lot for reviewing this!
I'm definitely not opposed to breaking this out into its own function, I just don't yet see how to do it in a way that improves clarity or simplicity.

josephfrazier:
You're welcome! I think it looks pretty good now. It could still be broken into its own function (like the one I posted [above](https://github.com/feross/webtorrent/pull/909#discussion_r82505024)), but that can happen later if needed.

We don't exactly have any tests for this part of the code, so can you confirm that it still fixes your use case?

jablko:
Excellent! I checked manually and confirm it still resolves the issue.

josephfrazier:
Merged, thanks!

jablko:
Thank you!

feross:
LGTM!

PR #911, opened 2016-09-15
Option to disable BEP19 web seeds

dcposch:
* For completeness. We have options to enable/disable WebRTC, the regular BitTorrent tracker client, and the DHT client. This lets us also enable/disable web seeds. 
* Useful for WebTorrent Desktop integration testing. See https://github.com/electron/electron/issues/7212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # WebTorrent Version History
 
+## v0.97.0 - 2016-09-17
+
+- Option to disable web seeds (BEP19)
+
 ## v0.95.2 - 2016-06-22
 
 - HEAD requests  to HTTP server should not send entire body
--- a/docs/api.md
+++ b/docs/api.md
@@ -51,11 +51,12 @@ If `opts` is specified, then the default options (shown below) will be overridde
 
 ```js
 {
-  dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
   maxConns: Number,        // Max number of connections per torrent (default=55)
   nodeId: String|Buffer,   // DHT protocol node ID (default=randomly generated)
   peerId: String|Buffer,   // Wire protocol peer ID (default=randomly generated)
-  tracker: Boolean|Object  // Enable trackers (default=true), or options object for Tracker
+  tracker: Boolean|Object, // Enable trackers (default=true), or options object for Tracker
+  dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
+  webSeeds: Boolean        // Enable BEP19 web seeds (default=true)
 }
 ```
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "webtorrent",
   "description": "Streaming torrent client",
-  "version": "0.96.5",
+  "version": "0.97.0",
   "author": {
     "name": "WebTorrent, LLC",
     "email": "feross@webtorrent.io",
--- a/test/node/download-webseed-torrent.js
+++ b/test/node/download-webseed-torrent.js
@@ -9,8 +9,12 @@ var serveStatic = require('serve-static')
 var test = require('tape')
 var WebTorrent = require('../../')
 
+// it should be fast to download a small torrent over local HTTP
+var WEB_SEED_TIMEOUT_MS = 500
+
 test('Download using webseed (via .torrent file)', function (t) {
   t.plan(6)
+  t.timeoutAfter(WEB_SEED_TIMEOUT_MS)
 
   var parsedTorrent = extend(fixtures.leaves.parsedTorrent)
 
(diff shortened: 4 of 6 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "webtorrent",
   "description": "Streaming torrent client",
-  "version": "0.96.5",
+  "version": "0.97.0",
```
feross: In the future, please don't change the `package.json` version in PRs. That's best done with the `npm version` command (i.e. `npm version patch`, `npm version minor`, `npm version major`) which creates a standalone commit where the version is changed. This makes it easier to spot releases in the changelog. It also takes care of creating a git tag on that same commit.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # WebTorrent Version History
 
+## v0.97.0 - 2016-09-17
+
+- Option to disable web seeds (BEP19)
+
 ## v0.95.2 - 2016-06-22
```
feross: Looks like v0.96.0 is missing. I'll add it.

docs/api.md
```diff
@@ -51,11 +51,12 @@ If `opts` is specified, then the default options (shown below) will be overridde
...
 ```js
 {
-  dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
   maxConns: Number,        // Max number of connections per torrent (default=55)
   nodeId: String|Buffer,   // DHT protocol node ID (default=randomly generated)
   peerId: String|Buffer,   // Wire protocol peer ID (default=randomly generated)
-  tracker: Boolean|Object  // Enable trackers (default=true), or options object for Tracker
+  tracker: Boolean|Object, // Enable trackers (default=true), or options object for Tracker
+  dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
+  webSeeds: Boolean        // Enable BEP19 web seeds (default=true)
```
feross: This would have been better as singular, `webSeed` IMO since we didn't name the other option `trackers`. No worries, I'll change it along with the other breaking API changes in v1.0.

test/node/download-webseed-torrent.js
```diff
@@ -9,8 +9,12 @@ var serveStatic = require('serve-static')
 var test = require('tape')
 var WebTorrent = require('../../')
 
+// it should be fast to download a small torrent over local HTTP
+var WEB_SEED_TIMEOUT_MS = 500
+
 test('Download using webseed (via .torrent file)', function (t) {
   t.plan(6)
+  t.timeoutAfter(WEB_SEED_TIMEOUT_MS)
```
feross: Won't this just add flakiness to the test suite? We don't do this in any of the other tests. What if it takes 600ms to complete on Travis CI servers?

test/node/download-webseed-torrent.js
```diff
@@ -72,3 +76,46 @@ test('Download using webseed (via .torrent file)', function (t) {
...
+
+      client = new WebTorrent({ dht: false, tracker: false, webSeeds: false })
+
+      client.on('error', function (err) { t.fail(err) })
+      client.on('warning', function (err) { t.fail(err) })
+
+      client.add(parsedTorrent, {store: MemoryChunkStore})
+
+      // The test above ensures that we can download the whole torrent over webseeds within a
+      // short time. Here, we wait the same amount of time and make sure no HTTP requests happen.
```
feross: I'm not sure what this comment means. Since the `webSeeds` option was set to `false`, no download progress will be made, right? This will just timeout after 500ms and move on to the cleanup code below.

That's fine, I think waiting 500ms is a reasonable way to ensure that no http requests get sent to the server, but the comment should be changed to be clearer.

feross: Also, another test we can add here is:

```js
var torrent = client.add(parsedTorrent, {store: MemoryChunkStore})

torrent.on('ready', function () {
  t.equal(torrent.urlList.length, 0) // ensure no web seeds were added
})
```

package.json
feross: This is best to do only when we're ready to do a release.

feross: If you're curious about the exact commands I use to do a release, they are here: https://github.com/feross/prezto/blob/master/runcoms/zshrc#L31-L33

feross: Sorry -- I see now that you did make a separate commit for the version bump. Still, it's not good to encourage this in PRs. When random people send you a PR and increment the version, it makes it harder to merge that PR.

CHANGELOG.md
feross: PR sent: https://github.com/feross/webtorrent/pull/915

package.json
dcposch: I did use `npm version minor`

I bumped the version because I need it for the desktop app: https://github.com/feross/webtorrent-desktop/pull/913

docs/api.md
dcposch: I wanted to match `webConns`. The other options, `tracker` and `dht`, both take an object, this one just takes true or false.

I think it's fine but have no strong opinion, if you want to change it in 1.0 that's cool too

test/node/download-webseed-torrent.js
dcposch: It completes very fast (~10ms) on my Macbook.

I added the timeout to match the timeout below. How do you verify that `webSeeds: false` is working? I thought: make sure that the webSeed download always completes in under X time, then set `webSeeds: false` and wait that same amount of time and verify there are no requests

test/node/download-webseed-torrent.js
dcposch: "The test above" refers to the test that has webSeeds enabled. There are two tests in the file, the first ensures that web seeds work, the second ensures that `webSeeds: false` works

dcposch: I don't think that extra test works. It doesn't prevent `urlList` from being populated (just as setting `tracker: false` won't make the `announce` list empty, right?). It just means that none of those URLs will be contacted.

CHANGELOG.md
dcposch: Left comments re some typos but otherwise looks good

PR #917, opened 2016-09-18
Add WebTorrent Clones List

DiegoRBaquero:
I'm guessing it's the right spot, I don't really know if the wording is the ideal. Open to suggestions.


dcposch:
I don't know if "clones" is the right word. It implies 1-for-1 copying.

I would say those are decentralized alternatives to centralized services. It's a great list and I think it's really cool that so many people are finding creative uses for WebTorrent.

Want to add WebTorrent Desktop to the list as an alternative to uTorrent and Transmission?

DiegoRBaquero:
@dcposch I put it that name because of the nomenclature I've seen throughout the internet like "YouTube Django Clone", etc.

WT Desktop is already in the list :)

PR #932, opened 2016-09-29
Add more peer ID entropy

feross:
This still keeps the peer ID printable as ascii, while adding lots more
entropy


feross:
16777216x as many possible peer ids 👍

feross:
Thanks for the review @DiegoRBaquero

PR #934, opened 2016-09-29
FAQ: Add Gaia 3D star map

feross:
http://charliehoey.com/threejs-demos/gaia_dr1.html


feross:
cc @flimshaw -- great work on this demo!

DiegoRBaquero:
When I first saw this in Chrome for iOS I was kind of lost as only some stars loaded. I just loaded it in my Mac and this is amazing! Nice work @filmshaw

PR #937, opened 2016-10-03
Initialize WRTC

DiegoRBaquero:
If the global WRTC variable is a function then call it to initialize it.

Related: 
- https://github.com/feross/webtorrent-hybrid/issues/46
- https://github.com/feross/webtorrent-hybrid/pull/47

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,9 +101,10 @@ function WebTorrent (opts) {
     if (opts.wrtc) {
       // TODO: remove in v1
       console.warn('WebTorrent: opts.wrtc is deprecated. Use opts.tracker.wrtc instead')
-      self.tracker.wrtc = opts.wrtc // to support `webtorrent-hybrid` package
+      self.tracker.wrtc = opts.wrtc
     }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -101,9 +101,10 @@ function WebTorrent (opts) {
     if (opts.wrtc) {
       // TODO: remove in v1
       console.warn('WebTorrent: opts.wrtc is deprecated. Use opts.tracker.wrtc instead')
-      self.tracker.wrtc = opts.wrtc // to support `webtorrent-hybrid` package
+      self.tracker.wrtc = opts.wrtc
```
feross: Good call removing this comment -- this option can be used by any package, not just `webtorrent-hybrid` :)

feross:
Actually, I think this belongs in `bittorrent-tracker` instead, since that's where the `wrtc` option gets used. Let's just support it there.

feross:
See: https://github.com/feross/bittorrent-tracker/commit/cb8c324ebe3e72f6ff82e332327881a19b7edc28

PR #947, opened 2016-10-12
Don't run Electron tests on Travis

josephfrazier:
This is a workaround for https://github.com/feross/webtorrent/issues/945

Note that pull requests from other repos will still run the Electron
tests (and remind us that it's broken), but at least master will be
green again.

### DIFF
--- a/bin/test.js
+++ b/bin/test.js
@@ -6,11 +6,11 @@ var runSauceLabs = !process.env.CI ||
   (process.env.SAUCE_USERNAME && process.env.SAUCE_ACCESS_KEY)
 
 npmRun('test-node', function () {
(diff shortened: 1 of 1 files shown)


josephfrazier:
Disregard, there's a better way to fix it. See the following:

* https://github.com/feross/webtorrent/commit/1013b5b8bc5ce581849aab1ad8608fdc94506fb7#commitcomment-19400884
* https://travis-ci.org/josephfrazier/webtorrent/builds/167166344#L1830

feross:
Thanks for doing this workaround, but I made a real fix here: https://github.com/feross/webtorrent/pull/949

PR #948, opened 2016-10-12
Fix xvfb on Travis

josephfrazier:
Fixes https://github.com/feross/webtorrent/issues/945

Adapted from https://github.com/feross/webtorrent/commit/1013b5b8bc5ce581849aab1ad8608fdc94506fb7


josephfrazier:
Hmm, now it seems like Travis is just flaky, given that master is currently equivalent to commit 6b4f073 above...

master: https://travis-ci.org/feross/webtorrent/builds/167193572#L1832
6b4f073: https://travis-ci.org/feross/webtorrent/builds/167173908#L1835

feross:
I just tried changing `electron` to `electron-prebuilt` and re-ran the build 5 times. All passed.

Looks like there was some change between electron pre-1.0 and latest 1.x that is causing a race condition. Maybe the timing of some event changed and `zuul` needs to be updated? :shrug:

PR #961, opened 2016-10-28
add 'user-agent' option

zunsthy:
#962


feross:
Let's set the user agent, but I don't think we want it to be configuration by the end-user since the implementation that's that making the requests ought to be identifiable IMO. This still fixes the issue of there being no user agent: https://github.com/feross/webtorrent/issues/962

feross:
This PR is superseded by https://github.com/feross/webtorrent/pull/1019

PR #968, opened 2016-11-02
Add localfiles.alhur.es

fiatjaf:
A project I had for a very long time, now made easy by WebTorrent.


DiegoRBaquero:
Nice idea! Thanks for sharing

feross:
Very interesting idea.

PR #972, opened 2016-11-06
Do not choke on web seeds

gpetrov:
Web seeds should be considered as pure uncheked seeds according to BEP19

So we should never choke on them. Otherwise when there are no other
seeds, the downloads will hang


lib/torrent.js
```diff
@@ -1093,8 +1093,10 @@ Torrent.prototype._onWireWithMetadata = function (wire) {
     wire.port(self.client.dht.address().port)
   }
 
-  timeoutId = setTimeout(onChokeTimeout, CHOKE_TIMEOUT)
-  if (timeoutId.unref) timeoutId.unref()
+  if (wire.type !== 'webSeed') { // do not choke on webseeds
```
josephfrazier: Given that `onChokeTimeout` will do nothing if `wire.type === 'webSeed'`, is this guard necessary?

DiegoRBaquero: I'd rather have this line and remove the change in onChokeTimeout. Never setting the timeout is better.

josephfrazier: Good point. I noticed there's one other place where the timeout could be set, but I'm assuming it won't actually happen for web seeds? https://github.com/gpetrov/webtorrent/blob/a6a52cf436dfaa078dfa58dabcc2d4ac01bbbf2a/lib/torrent.js#L1070

DiegoRBaquero: Nope, a webseed can't send a "choke" signal.

feross: I agree with this. Let's not set the timeout

feross:
After the PR is updated, feel free to land this PR

lib/torrent.js
gpetrov: After further diving into the webseed logic, I think it should be even better if the webseed kicks in as last, when for example there are no other peers. Currently if webseed is given it is directly used as first peer and when this is a fast website connection, most downloads will go through the webseed and other peers will be used very little.

So I would suggest that webseeds also get the lowest priority (maybe as option).

I solved this currently in my case by adding the webseed dynamically 3 seconds after the torrent has being activated, so all client peers are exhausted first, but it is not an optimal solution.

What do you think?

josephfrazier: I'm not too familiar with the logic/code that handles webseed priority, but it sounds like a technically separate issue. I think we should go ahead and land this PR (once the required change has been made), then discuss webseed priority in a new issue/PR :-)

gpetrov:
are we good now? :)

josephfrazier:
Looks great, thanks!

DiegoRBaquero:
Great! Something we needed. @feross Can we get this published?

feross:
@DiegoRBaquero Sorry for the delay. Just published this as 0.98.1.

PR #974, opened 2016-11-10
Adding property for downloaded bytes per file.

mikeal:
Needed this in an app I've been building. It's almost the same as the property on the full torrent.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -38,6 +38,24 @@ function File (torrent, file) {
   }
 }
 
+Object.defineProperty(File.prototype, 'downloaded', {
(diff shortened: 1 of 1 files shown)


vankasteelj:
I second this, I need it too. `torrent.files[<index>].downloaded` is an API call I spent too much time trying to find x)

mikeal:
Is this going to get merged soon or should I deploy with a fork for a while until it is?

feross:
Released as 0.98.0. Sorry for the delay.

vankasteelj:
Awesome, thanks

PR #989, opened 2016-12-04
Add WTGC

FluorescentHallucinogen:
Add WebTorrent Google Cast (WTGC) app to "Who is using WebTorrent today?" list.


feross:
Wow, nice! I didn't realize that Google Cast was available via a JS API in Chrome!

PR #991, opened 2016-12-06
Don't send 'completed' event to tracker on client.seed

sidd:
Follow-on to [bittorrent-tracker/#185](https://github.com/feross/bittorrent-tracker/pull/185)

Completed event now only emits when a previously started torrent download completes

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1532,7 +1532,7 @@ Torrent.prototype._request = function (wire, index, hotswap) {
           wire.have(index)
         })
 
-        self._checkDone(onDone)
+        if (self._checkDone()) self.discovery.complete()
       } else {
         self.pieces[index] = new Piece(piece.length)
         self.emit('warning', new Error('Piece ' + index + ' failed verification'))
@@ -1541,10 +1541,6 @@ Torrent.prototype._request = function (wire, index, hotswap) {
     })
   })
 
-  function onDone () {
-    self.discovery.complete()
-  }
-
   function onUpdateTick () {
     process.nextTick(function () { self._update() })
   }
@@ -1552,10 +1548,9 @@ Torrent.prototype._request = function (wire, index, hotswap) {
   return true
 }
 
-Torrent.prototype._checkDone = function (cb) {
+Torrent.prototype._checkDone = function () {
   var self = this
   if (self.destroyed) return
-  if (!cb) cb = noop
 
   // are any new files done?
   self.files.forEach(function (file) {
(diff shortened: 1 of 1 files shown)


sidd:
@DiegoRBaquero From a glance it looks like `_checkDone` is eventually called via `_onTorrentId`, but this is from client.seed. So we could remove `_checkDone` from torrent.load, but now `torrent.js`'s proper operation is dependent on properly using it through client.seed

I'm not very close to this code, so let me know if I'm missing something obvious!

feross:
@DiegoRBaquero `_checkDone` needs to get called or else `file.on('done')` and `torrent.on('done')` won't fire on torrents that are added when they are already complete on disk.

I think this PR is correct to just change the message that is sent to the tracker server.

However, I have style comments. See inline.

lib/torrent.js
```diff
@@ -1541,16 +1541,21 @@ Torrent.prototype._request = function (wire, index, hotswap) {
...
+
   function onUpdateTick () {
     process.nextTick(function () { self._update() })
   }
 
   return true
 }
 
-Torrent.prototype._checkDone = function () {
+Torrent.prototype._checkDone = function (cb) {
```
feross: No need for a callback. This method is entirely sync. Let's just return `true` if the entire torrent is done, `false` otherwise.

lib/torrent.js
```diff
@@ -1579,7 +1584,7 @@ Torrent.prototype._checkDone = function () {
   if (!self.done && done) {
     self.done = true
     self._debug('torrent done: ' + self.infoHash)
-    self.discovery.complete()
+    cb(null)
     self.emit('done')
```
feross: Let's return `true` at the very end of this `if` statement.

lib/torrent.js
```diff
@@ -1579,7 +1584,7 @@ Torrent.prototype._checkDone = function () {
...
   }
```
feross: Otherwise, return `false`.

lib/torrent.js
```diff
@@ -1532,7 +1532,7 @@ Torrent.prototype._request = function (wire, index, hotswap) {
           wire.have(index)
         })
 
-        self._checkDone()
+        self._checkDone(onDone)
```
feross: If this returns `true`, then we can call `self.discovery.complete()`.

lib/torrent.js
sidd: Returning true before `_gcSelections` seems to be causing tests to hang. getBuffer never calls its callback: https://github.com/feross/webtorrent/blob/master/test/node/download-dht-magnet.js#L77.  `return self.done` after `_gcSelections` the function resolves this. Thoughts?

feross: Yeah, good point

feross:
0.98.2

PR #1009, opened 2017-01-03
added BitChute to the app list

rayvahey:
Hi, thanks for your amazing project. I've been using it in my own project over the Christmas Holidays and would like to submit it to the webtorrent app list.


feross:
Looks nice!

PR #1012, opened 2017-01-11
Until it is fixed, let notify webtorrent user

kocoten1992:
https://github.com/feross/webtorrent/issues/1000


feross:
Thanks for the PR, @kocoten1992! Once tests pass, I'll merge.

PR #1018, opened 2017-01-17
Add Plaktos to "Projects" section

feross:
I noticed in an issue comment that @georgeaf99 built this project using
WebTorrent :tada:


georgeaf99:
Awesome, thanks!

PR #1039, opened 2017-02-10
Refactor http server; support content-disposition

feross:
Refactored the server into many smaller functions to make it easier to
understand all the different code paths.

- added a Content-Disposition header, which tells the browser the
file's name, since we use urls like http://localhost:port/0 <-- no
human-readable file name
- Server returns valid HTML documents (with all the required tags) now.
- Return 204 status for OPTIONS request
- reduce access-control-max-age to chromium max of 600s
- respond to OPTIONS requests that lack
'access-control-request-headers' (before they were treated as GET)
- return '405 invalid verb' for all other verbs

For: https://github.com/brave/browser-laptop/issues/6737

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -1,7 +1,6 @@
 module.exports = Server
 
 var arrayRemove = require('unordered-array-remove')
-var debug = require('debug')('webtorrent:server')
 var http = require('http')
 var mime = require('mime')
 var pump = require('pump')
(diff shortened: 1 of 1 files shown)


feross:
Post-merge review welcome 🖐 Merging and releasing because I need this to fix a bug in Brave.

feross:
0.98.6

feross:
Thanks @DiegoRBaquero!

Issue #163, opened 2014-10-24
Limit download/upload speeds (throttling)


green13:
It would be cool to be able to limit download/upload speeds.

servermeta:
As a noobish system administrator it feels a bit strange to me to use Javascript to limit bandwidth. Maybe it's just a prejudice, but node doesn't strike me as being a good candidate.

If it's ok to limit the solution to only linux systems, we could use Ip/Nftables and tc. We could open the stream as a child process, and then the linux kernel could keep track of the process, which is very efficient at doing it. 


```
PID=`ps aux |grep inetd |head -n 1 |cut -b 10-14`
```

First we get the PID of the process

```
/usr/local/sbin/iptables -A OUTPUT -p TCP -m owner --pid-owner $PID -j ACCEPT
````

we could use iptables to mark the packet, then Traffic control could shaper on a per PID basis

http://lartc.org/howto/lartc.qdisc.filters.html

Svenskunganka:
@servermeta It's bad habit to rely on binaries outside the project, especially those who require root privileges. Node is definitely more than qualified for this task, it is known for its I/O and networking capabilities, we should utilize that.

servermeta:
@Svenskunganka the only thing we would have to rely on is a unix kernel :P

astro:
No, iptables is specific to Linux.

servermeta:
yes you are right. I'm studying the source to try to come up with a good solution, since I need this, and to better learn this project. @feross why you say 
`(Btw, it doesn't make much sense for the browser version of webtorrent).`

astro:
While dropping packets is perfectly valid, this shaping is best done in the actual code.

FWIW here's my approach from Bitford: https://github.com/astro/bitford/blob/master/src/shaper.js
I bet there are similar modules on npmjs.org.

bradisbell:
Once implemented, will stream backpressure on the read side slow the download speed?

astro:
Backpressure should go into https://github.com/feross/bittorrent-protocol/ where it doesn't seem to be implemented at this time. That will only become a problem if download speeds exceed CPU power & disk throughput.

bradisbell:
@astro In the download case yes, but when streaming media it is very common that the download speed greatly exceed the playback speed.  For example, if you pipe media into VLC, only a couple seconds of media data will be buffered prior to decoding and playback.

feross:
https://github.com/substack/node-brake

[...]

PR #1040, opened 2017-02-11
Feature/163 global download upload limit

chtrinh:
throttleDownload() and throttleUpload() helper methods for rate adjustments

#163


BenmasterTM:
HI, I try this changes in the webtorrent client in the webtorrent-desktop client, and when I try to set the upload/download limit for Web Torrent, this changes dont work, the only way I found to slowdown the upload files from the webtorrent client in webtorrent-desktop is changing the webtorrent/node_modules/bittorrent-protocol/index.js changing the Wire.prototype.piece function, and encapsulating the code into a timeout:

`    var self = this;
    
    (function(self, index, offset, buffer) {
    setTimeout(function() {

    self._debug('piece index=%d offset=%d', index, offset)
    self.uploaded += buffer.length
    self.uploadSpeed(buffer.length)
    self.emit('upload', buffer.length)
    self._message(7, [index, offset], buffer)
    
    }, 800)    
    })(self, index, offset, buffer)
`
altering the timeout time change the upload speed to the web seeds, but cause many messages "start backpressure" from the simple-peer module

chtrinh:
How did you set upload/download limits for webtorrent-desktop client? 
it should be something like this: 
```
var torrent = new WebTorrent({downloadLimit: 10240, uploadLimit: 10240}); // (downstream/upsttream) bandwidth to 10 KB/s
```
This PR relies on https://www.npmjs.com/package/stream-throttle to throttle your bandwidth.

BenmasterTM:
Yes I try that:

```let client = window.client = new WebTorrent({downloadLimit: 10240, uploadLimit: 10240})```

but the clients connected to the webtorrent client get 100% of my internet connection, i think is because im talking always about WebRTC clients (I dont test with regular uTPC clients, but in the code I see self.type === 'tcpOutgoing' and i think the web clients go for another point)

**Comment by chtrinh** [comment]
My initial use case was for TCP not WebRTC. IIRC, I encountered an edge case where you couldn't always set throttle pipes. You don't know what "tcpIncoming" are going to be use for. 
https://github.com/chtrinh/webtorrent/blob/462a1916fc1736fa9ede2ce0240e1500efad51de/lib/peer.js#L32-L39

I would try something like so:
```js
 if ((self.type === 'tcpOutgoing') || (self.type === 'webSeed')) {
    self.setThrottlePipes()
 } else {
    conn.pipe(wire).pipe(conn)
  }
```

BenmasterTM:
The correct value in Web Peers is self.type === 'webrtc', but works correctly, thanks!

BenmasterTM:
Hi, I found an error in specific case: If you change the upload or download speed while torrents are downloading/uploading, and change the speed limit (with the helper function or directly in webtorrent.uploadSpeed / webtorrent.downloadSpeed, the webtorrent crash:

Error: Uncaught, unspecified "error" event. (Requested tokens 2984254 exceeds bucket size 1049600)
    at Throttle.emit (events.js:163)
    at Throttle.onerror (D:\Escritorio\electro\node_modules\webtorrent\node_modules\readable-stream\lib\_stream_readable.js:562)
    at emitOne (events.js:96)
    at Throttle.emit (events.js:188)
    at onwriteError (_stream_writable.js:344)
    at onwrite (_stream_writable.js:362)
    at afterTransform (_stream_transform.js:79)
    at TransformState.afterTransform (_stream_transform.js:54)
    at D:\Escritorio\electro\node_modules\webtorrent\node_modules\stream-throttle\src\throttle.js:34
    at _combinedTickCallback (internal/process/next_tick.js:67)
    at process._tickCallback (internal/process/next_tick.js:98)

rigelk:
@chtrinh do you still plan on finishing this ?

quicoto:
Any updates on this? 🙏

TwoAbove:
@BenmasterTM unfortunately, no news on that, but I can confirm that this fork works

craigharman:
Any chance that someone is going to look at this again? @chtrinh seems like is a few formatting issues and deletion of .min.js file away?

ultimate-tester:
Subscribed. This pull request should be accepted. On my server the current implementation will actually at times max out on bandwidth dropping other non-torrent connections. Therefore a limit would be really appropriate.

ultimate-tester:
Even better, I've just opened another pull request #1852 to deal with the outdated situation of this pull request. 

The essential code is 100% identical to the pull request of @chtrinh, I've merely adjusted it to fit with the ECMAScript version used in webtorrent. It will merge succesfully with the current version in master branch.

feross:
Superceded by #1852

PR #1044, opened 2017-02-14
wait to notify() or updateInterest() at end of GC

feross:
To prevent it from possibly being called twice, which will do extra work


lib/torrent.js
```diff
@@ -893,7 +893,7 @@ Torrent.prototype.deselect = function (start, end, priority) {
   for (var i = 0; i < self._selections.length; ++i) {
     var s = self._selections[i]
     if (s.from === start && s.to === end && s.priority === priority) {
-      self._selections.splice(i--, 1)
+      self._selections.splice(i, 1)
```
feross: the `--` is not needed to compensate for the `splice()`, since we call `break` right afterwards.

PR #1045, opened 2017-02-15
Updating electron-prebuilt to 1.4.13

DiegoRBaquero:
Please update [electron-prebuilt](https://npmjs.org/package/electron-prebuilt) to version 1.4.13.

If this passes your tests you can merge it in.  If not please use this branch for changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "brfs": "^1.4.3",
     "browserify": "^14.0.0",
     "cross-spawn": "^5.0.1",
-    "electron-prebuilt": "^0.37.8",
+    "electron-prebuilt": "^1.4.13",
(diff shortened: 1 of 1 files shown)


feross:
This causes the Electron tests to freeze on Travis. I tried to debug this before, but I think it's a subtle bug in the zuul test runner. You can see how it froze here: https://travis-ci.org/feross/webtorrent/builds/201759142

I think we should keep this pinned for now.

feross:
@DiegoRBaquero Btw, what is bithound? I see that in the branch name

DiegoRBaquero:
@feross bithound does code analysis on repos, looks for security issues, dep issues, etc. It was it that suggested the upgrade.

PR #1049, opened 2017-02-17
Create .codeclimate.yml

larkinscott:
Scott here from Code Climate! I created a Code Climate config with ESLint and our duplication engine enabled. Happy to answer any questions you might have!


feross:
Sorry, no thanks. I don't know what your "duplication engine" does but it doesn't sound useful. We already have a linter too.

Issue #195, opened 2014-11-29
UPnP and NAT-PMP port forwarding


feross:
> Universal Plug and Play (UPnP) is a set of networking protocols that permits networked devices, such as personal computers, printers, Internet gateways, Wi-Fi access points and mobile devices to seamlessly discover each other's presence on the network and establish functional network services for data sharing, communications, and entertainment. UPnP is intended primarily for residential networks without enterprise-class devices.

Note: UPnP only makes sense in the node.js (server) use case. In the browser, WebTorrent uses WebRTC and the browser handles NAT traversal using the ICE, STUN, and TURN protocols. Lots of fun acronyms :)

This issue is for adding UPnP and/or NAT-PMP support to WebTorrent. There appear to be [quite a few implementations](https://www.npmjs.org/search?q=upnp) on npm, so a bit of research will be required to find the best implementation to use.

jman6495:
Second this, I'm trying to make connections and it simply won't work ..

astro:
@jman6495 In node, right? Making outgoing connections doesn't need an open port or even a listening socket.

feross:
@jman6495 Yeah, you should be able to download just fine without UPnP and NAT-PMP port forwarding. They're just useful if you want people to be able to make *incoming connections* to you.

drwasho:
Any progress on this?

mnaamani:
There is a nice library that I use in my own projects, [node-nat-upnp](https://github.com/indutny/node-nat-upnp) but it needs a couple of small [fixes](https://github.com/indutny/node-nat-upnp/pulls)

retrohacker:
Leaving this here for future reference

RFC5128: State of Peer-to-Peer (P2P) Communication across Network Address Translators (NATs)
https://tools.ietf.org/html/rfc5128

roccomuso:
Any update on this issue?
Is it just a matter of finding the right upnp module to use?

feross:
@roccomuso Yeah, and someone needs to do the work of integrating it.

roccomuso:
@feross any idea on the upnp module to use?

feross:
No, sorry. I haven't done an exhaustive search of npm. Ideally, something with no or few dependencies. And definitely no native dependencies.

roccomuso:
Ok, there's no wide choice around, I came out with this, https://github.com/indutny/node-nat-upnp
Take a look and let me know.

feross:
[...]

PR #1051, opened 2017-02-18
Add support for NAT traversal techniques

yciabaud:
Here is an implementation of NAT traversal as discussed in #195 using https://github.com/indutny/node-nat-upnp and https://github.com/TooTallNate/node-nat-pmp.

I have also ecountered https://github.com/level451/pmp that claims to make both UPnP ad NAT-PMP but I don't have tested it.

This feature is implemented in a new module for node env only and provided by a single instance shared by the webtorrent clients because NAT-PMP needs to open a specific port to work.

I checked the UPnP part with my router by it does not support NAT-PMP protocol.

The mappings are sent both with UPnP and NAT-PMP whe available and refreshed every 30 min.

Reviews are welcome!


yciabaud:
Looks like TooTallNate/node-nat-pmp#13 prevents Travis from running the tests

yciabaud:
Using my patch TooTallNate/node-nat-pmp#13 does the trick, we should wait for it to be merged but I have no feedback ATM.

heroboy:
What protocol port you used to map? As I know dht uses UDP. `self._natTraversal.portMapping` seems mapping TCP port default.

oleiba:
That sounds interesting!
What's left for it to be merged?

PR #1061, opened 2017-02-25
Implement correct interested behaviour

### LINKED ISSUE #1059: Implementing correct Interested/Uninterested behaviour
Hi, I'm going to try and add (more) correct interested/uninterested behaviour i.e. only marking a peer as interested if they actually have a piece that we want. Here was my first idea:
In `Torrent.prototype._updateInterest`, for each wire iterate through their `peerPieces` bitfield and mark them as interested if they have any piece which we do not currently have.

Is this too inefficient? It would mean #peers * #pieces complexity but from what I can see `_updateInterest` isn't called very often so this might not be too much of a problem?

bradleyjkemp:
Start peers as uninterested and only move to interested if/once they have a piece that we need

Fixes #1059

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1181,7 +1181,7 @@ Torrent.prototype._updateInterest = function () {
         break
       }
     }
-    
+
     if (interested) wire.interested()
     else wire.uninterested()
   })


lib/torrent.js
```diff
@@ -1174,8 +1174,15 @@ Torrent.prototype._updateInterest = function () {
   self._amInterested = !!self._selections.length
 
   self.wires.forEach(function (wire) {
-    // TODO: only call wire.interested if the wire has at least one piece we need
-    if (self._amInterested) wire.interested()
+    var interested = false
+    for (var index = 0; index < self.pieces.length; ++index) {
+      if (self.pieces[index] && wire.peerPieces.get(index)) {
```
DiegoRBaquero: Shouldn't this `!self.pieces[index]`? Indicating that we don't yet have that piece?

KayleePop: `self.pieces[index]` is deleted when that piece is downloaded and verified, so `self.pieces[index]` would be truthy when the piece still needs to be downloaded. The actual piece data is in the store. `self.pieces[]` is used to facilitate downloading.

`!self.bitfield.get(index)` is clearer.

PR #1076, opened 2017-03-12
Fix detection of seeding peers

malern:
`wire.peerPieces.length` is always undefined, which meant the `updateSeedStatus` function was always returning early, and peers were never being detected as seeds.


feross:
Released as 0.98.11

PR #1078, opened 2017-03-16
add filename to path

pahwaranger:
This allows you to add the filename to the path with produces more readable and understandable URLs. It doesn't actually do anything with or require the additional URL params so it shouldn't break any existing deployments.

Current URL:
`localhost:3000/0`

New URL:
`localhost:3000/0/FILE_NAME.mp4`

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -107,7 +107,7 @@ function Server (torrent, opts) {
         return serveIndexPage()
       }
 
-      var index = Number(pathname.slice(1))
+      var index = Number(pathname.split('/')[1])
       if (Number.isNaN(index) || index >= torrent.files.length) {
         return serve404Page()
       }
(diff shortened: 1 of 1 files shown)


DiegoRBaquero:
Sure, why not.

lib/server.js
```diff
@@ -107,7 +107,7 @@ function Server (torrent, opts) {
         return serveIndexPage()
       }
 
-      var index = Number(pathname.slice(1))
+      var index = Number(pathname.split('/')[0])
```
feross: Did you actually test this out? This looks wrong. Shouldn't it be `pathname.split('/')[1]`?

feross:
```js
'/0/file.mp4'.split('/')   // [ '', '0', 'file.mp4' ]
```

pahwaranger: I am running this on my own deployment and it is working. Not sure how to explain this though, your logic is also correct. I'll do some more testing and update later.

pahwaranger: Wow, so it happened to be that I was always testing with `0` as the file index. `Number('')` resolves to 0. Will update the PR

pahwaranger:
@feross good catch, I have updated the PR. PTAL

feross:
Released as 0.98.14.

PR #1086, opened 2017-03-28
Add keen.tv live video player 

boiankeen:
live video section from keen.tv appended to faq.md


feross:
Thanks for the PR but it's not obvious what the copyright status is of the videos on your site. We don't link to, promote, or aid copyright infringement in the WebTorrent project, so out of an abundance of caution I'm going to reject this change.

Thanks.

PR #1091, opened 2017-03-31
Add free-torrents.md

feross:
This markdown file provides a list of the sample torrents built into
WebTorrent Desktop, but with the following improvements:

- working trackers
- all web seeds now hosted on webtorrent.io
- cleaned up files (no useless archive.org metadata)
- added multi-language subtitle files, where available
- magnet links include &xs= property linking to torrent file, so they
can work even when there are no peers

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,11 +18,13 @@ npm install webtorrent
 ```js
 var client = new WebTorrent()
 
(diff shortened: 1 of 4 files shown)


DiegoRBaquero:
@feross Are you running a tracking server in webtorrent.io again? I see a non secure ws listed. Maybe it got added because of the xs too?

PR #1096, opened 2017-04-08
Add `origin` option for torrent.createServer()

feross:
When the origin option is specified, only requests from the given
origin will be allowed.

This is useful to add additional security to any app that is starting a
WebTorrent server but doesn't want it to be exposed to the entire Web.


yciabaud:
Nice addition, this will be useful!
@Thorbjorn needs something like this to ensure its torrents are consumed only on its platform.

PR #1102, opened 2017-04-13
Add local file deletion on `torrent.destroy`

Rowern:
When we call `torrent.destroy` and we are not running in a browser, we use
`rimraf` to delete the folder containing the torrent files.

This is a work in progress because I need feedback on a good way to implement it.
For instance, `client.destroy` calls `torrent.destroy` on all its torrents, this might be problematic as destroying the client looks like the legitimate way of stopping every torrent transactions but keeping your progress.

### DIFF
--- a/test/torrent-destroy.js
+++ b/test/torrent-destroy.js
@@ -1,8 +1,10 @@
+var fs = require('fs')
+var path = require('path')
 var fixtures = require('webtorrent-fixtures')
 var test = require('tape')
 var WebTorrent = require('../')
 
-test('torrent.destroy: destroy and remove torrent', function (t) {
+test('torrent.destroy: destroy torrent', function (t) {
   t.plan(5)
 
   var client = new WebTorrent({ dht: false, tracker: false })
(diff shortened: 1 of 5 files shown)


feross:
Thanks for the PR, @Rowern!

I think the right way to handle this is to call the `destroy()` method on the store, which already uses `rimraf` internally. See here: https://github.com/feross/fs-chunk-store/blob/8d7a57b56572625346024fca8cf3e2839ba2d49d/index.js#L222-L232

I haven't looked, but I suspect that we're only calling `close()` to close the store, but not destroy it.

You're right that we also want to keep the option to destroy the torrent without destroying the files on disk, so that probably should be an option to `torrent.destroy()`, like `torrent.destroy({ remove: true })` which defaults to `false`.

Rowern:
You are right, I will continue to work on this PR tomorrow.

test/torrent-destroy.js
```diff
@@ -22,3 +24,34 @@ test('torrent.destroy: destroy and remove torrent', function (t) {
...
+  }, function (torrent) {
+    t.equal(client.torrents.length, 1)
+    t.equal(torrent.infoHash, fixtures.leaves.parsedTorrent.infoHash)
+    t.equal(torrent.magnetURI, fixtures.leaves.magnetURI)
+
+    var completeFileName = path.join(torrent.path, torrent.files[0].name)
+
+    client.remove(torrent, {'remove': true}, function (err) {
+      t.error(err, 'torrent removed')
+      fs.stat(completeFileName, function (err) {
```
Rowern: I know this might not work well with browser API, I might add a check on the type of `store` used.

Rowern:
Sorry for closin/re-opening but it was the only way to re-trigger a travis build.

PR #1103, opened 2017-04-13
Fix uncaught exception

feross:
Fixes: https://github.com/feross/webtorrent/issues/1100
Fixes: https://github.com/feross/webtorrent/issues/1101


feross:
0.98.17

PR #1126, opened 2017-05-14
Update faq.md

manicphase:
added super quick share


DiegoRBaquero:
Hello @manicphase ! Thank you for sharing what you've created.

I had trouble getting to use it, however. I could not get to share an image.

It would be best if you could make the necessary changes to improve the user experience and make it more user friendly before we add the link.

manicphase:
Thanks for the feedback.

Could you tell me what you did and what went wrong? Like what devices/browsers you used. A console output if possible. 

I know there's a couple of issues like iPhone compatibility, webtorrent trackers are blocked on corporate WiFi and that it's unable to transfer between ipv4 and ipv6, but I haven't had a problem otherwise. I should probably implement some error messages for that and some more obvious feedback for when it's working on something before requesting a pull again.

Thanks.

manicphase:
I think i might have fixed the bug. There was an issue with chrome where it would try to use scrollTo on an element that chrome doesn't allow but firefox does. Is this working for you now?

PR #1161, opened 2017-07-24
Only show json-loader requirement for webpack 1.x

alxlu:
json-loader is included in webpack 2 and up: https://webpack.js.org/guides/migrating/#json-loader-is-not-required-anymore


DiegoRBaquero:
I have no idea about this, but do we really want to force users to use webpack 2+?

josephfrazier:
I don't see it as forcing users to use webpack 2+, just informing them that they can simplify their `webpack.config.js` if they do.

PR #1183, opened 2017-09-08
fix debug function undefined bug

magicxiao:
fix _debug function throw undefined bug when call torrent.destory() function.


DiegoRBaquero:
Hi @magicxiao, did you run into this problem?

PR #1221, opened 2017-11-04
Add Cinetimes.org project to FAQ

MiKatre:
Added [Cinetimes](http://cinetimes.org/) link to the project list.
It si not in english yet but it is just a matter of time.


DiegoRBaquero:
Are these movies copyright free?

MiKatre:
I do my best to make sure they are! According to french law though.

PR #1260, opened 2018-01-12
Add hostname option to mitigate DNS rebinding

diracdeltas:
This adds the `hostname` option to allow the server to validate the `Host` header of incoming requests to prevent DNS rebinding attacks. Needed for https://github.com/brave/browser-laptop/issues/12616.

For more context, see https://bugs.chromium.org/p/project-zero/issues/detail?id=1447.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -51,6 +51,15 @@ function Server (torrent, opts) {
     // deny them
     if (req.headers.origin == null) return false
 
+    // If a 'hostname' string is specified, deny requests with a 'Host'
+    // header that does not match the origin of the torrent server to prevent
+    // DNS rebinding attacks.
+    if (opts.hostname) {
+      if (req.headers.host !== `${opts.hostname}:${server.address().port}`) {
+        return false
+      }
+    }
+
     // The user allowed all origins
     if (opts.origin === '*') return true
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -54,10 +54,8 @@ function Server (torrent, opts) {
     // If a 'hostname' string is specified, deny requests with a 'Host'
     // header that does not match the origin of the torrent server to prevent
     // DNS rebinding attacks.
-    if (opts.hostname) {
-      if (req.headers.host !== `${opts.hostname}:${server.address().port}`) {
-        return false
-      }
+    if (opts.hostname && req.headers.host !== `${opts.hostname}:${server.address().port}`) {
+      return false
     }
 
     // The user allowed all origins
--- a/docs/api.md
+++ b/docs/api.md
@@ -327,6 +327,7 @@ Returns an `http.Server` instance (got from calling `http.createServer`). If
 ```js
 {
   origin: String // Allow requests from specific origin. `false` for same-origin. [default: '*']
+  hostname: String // If specified, only allow requests whose `Host` header matches this hostname. Note that you should not specify the port since this is automatically determined by the server. Ex: `localhost` [default: `undefined`]
 }
 ```
 


diracdeltas:
lazy manual test:

1. in Brave, or another webtorrent client, open `https://webtorrent.io/torrents/sintel.torrent`. in the code that calls `createServer`, specify `origin: *` to pretend that same-origin policy doesn't exist. (this is done to simulate DNS rebinding.)
2. click the 'start download' button and inspect the page to figure out the origin of the torrent server, ex `http://localhost:$PORT`
3. serve the following HTML file on a localhost server and then load it in your browser: 
```
<div id='result'></div>
<script>
    const PORT = $PORT // replace this with $PORT number from step 2
    const serverUrl = `http://localhost:${PORT}/0`
    fetch(serverUrl).then((response) => {
      response.text().then((text) => result.innerText = text)
    })
</script>
```
4. the page should display a bunch of text from the torrent file
5. now go back to the source code that calls `createServer` and add `hostname: 'localhost'` to the `createServer` options. 
6. repeat steps 1-3
7. this time the page should *not* load a bunch of text from the torrent file

diracdeltas:
credit goes to @taviso for finding and posting https://github.com/transmission/transmission/pull/468

diracdeltas:
@feross any thoughts on this approach?

lib/server.js
```diff
@@ -51,6 +51,15 @@ function Server (torrent, opts) {
     // deny them
     if (req.headers.origin == null) return false
 
+    // If a 'hostname' string is specified, deny requests with a 'Host'
+    // header that does not match the origin of the torrent server to prevent
+    // DNS rebinding attacks.
+    if (opts.hostname) {
+      if (req.headers.host !== `${opts.hostname}:${server.address().port}`) {
```
feross: What would Brave set the `hostname` option to? If it would be set to `chrome-extension://xxx`, then I fear this won't work since it's going to append the torrent server's port to that and reject anything that doesn't match that.

I think we can remove the code that adds the port. Correct me if I'm wrong.

feross:
One last thought -- the `origin` option currently has three possible values:

1. `'*'` - allow all cross-origin requests
2. `'example.com'` - only allow cross-origin requests from `example.com`
3. `false` - no cross-origin requests allowed

(Same origin requests are always allowed, with all the options above)

Given the presence of DNS rebinding, it doesn't make sense to offer the third option, since cross-origin requests are effectively allowed, even when that option is set.

## Proposal

Let's remove the special handling when the `origin` option is `false`.

If we merge the `hostname` option into the `origin` option as I proposed above, that leaves the following possible values for `origin`:

1. `'*'` - allow all requests from any origin
2. `'example.com'` - allow requests from example.com only

### When { origin: '*' } is set:

[This is the default value when the user does not set an origin.]

- Server will not check the `Host` header.

- When the `Origin` header is present (indicating a cross-origin request) and set to **ANY value** the server will add CORS headers to the response. This tells the user's browser to allow the request.

### When { origin: 'example.com' } is set:

- Server refuses to reply to any request where the `Host` header is not set to `chrome-extension://xxx`.

- When the `Origin` header is present (indicating a cross-origin request) and the value of that header is `chrome-extension://xxx` the server will add CORS headers to the response. This tells the user's browser to allow the request.

## Thoughts?

@diracdeltas LMK your thoughts.

lib/server.js
diracdeltas: @feross actually it would set it to `localhost`. The expected host is `localhost:$RANDOM_PORT`, as assigned in https://github.com/dcposch/webtorrent-remote/blob/3e99afb9a58787afb68f780da0a8f3174aa6cf0e/server.js#L203

note that the Host header is the origin of the server that the client is trying to contact

diracdeltas:
@feross It actually can't be merged with `origin` AFAICT since:

* the `origin` parameter specifies the `Access-Control-Allow-Origin` HTTP *response* header returned by the server on `localhost:$RANDOM_PORT`. this is used for the server to whitelist what other origins can make requests to it.
* the `host` parameter specifies the `Host` HTTP *request* header which the server will accept from a client. This is somewhat confusing because you'd think that if a browser is making a request to localhost:44494 (for instance), then the Host header would be always set to localhost:44494. However the key trick of DNS rebinding is forcing the client to visit localhost:44494 by binding it to attacker.com:44494 - in that case, the Host header seen by the torrent server will be attacker.com:44494 instead of localhost:44494. This PR lets the server detect when this is happening and block those requests.

feross:
Sorry – I lost track of this PR.

Merged and released as 0.98.24.

PR #1289, opened 2018-02-09
Fix cleanup DHT tables cache on torrent removal

DEgITx:
Fix bug on second adding of same torrent. After removing torrent and adding again to webtorrent it didn't search it in DHT again.


feross:
Thanks for the PR! ✨Some feedback:

- This uses a private API in `bittorrent-dht`. We should probably expose this functionality in the DHT API. @mafintosh - does this seem like a reasonable API addition? Is there some better way to do this?

- Can you add a test that fails without this change, but passes with the change?

DiegoRBaquero:
@DEgITx willing to do the changes?

feross:
Even though this PR needs work, I'm re-opening so it's here as a TODO since this seems like it fixes a real issue.

PR #1294, opened 2018-02-16
Add jsdelivr download count badge

feross:
[jsdelivr download count](https://cdn.jsdelivr.net/npm/webtorrent@latest/webtorrent.min.js)


feross:
5M now!

PR #1305, opened 2018-02-20
Add live-torrent to FAQ

pldubouilh:
Simple HLS based live streaming implementation using webtorrent whenever possible


pldubouilh:
Thanks :+1: Will do that soon yep !

PR #1309, opened 2018-02-21
Replace zuul with airtap

feross:
Testing out a replacement package for zuul called airtap.

https://www.npmjs.com/package/airtap

Much simpler than zuul in that it only supports tap/tape, actively maintained, doesn't rely on the unreliable localtunnel service.

It's still super early software, so I want to test it on a few repos first. Currently added it to these repos:

- `simple-peer`
- `magnet-uri`
- `buffer`
- `webtorrent` (this PR)

It may have big breaking changes before V1, but it's already so much more reliable that I think it's worth it.

### DIFF
--- a/.airtap.yml
+++ b/.airtap.yml
@@ -1,4 +1,5 @@
-ui: tape
+sauce_connect: true
+loopback: airtap.local
 browsers:
   - name: chrome
     version: -1..latest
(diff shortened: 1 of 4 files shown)


feross:
Looks good!

Issue #68, opened 2014-05-29
Implement µTP extension (BEP29)


transitive-bullshit:
Micro Transport Protocol or µTP is a UDP-based variant of the standard bittorrent protocol intended to mitigate poor latency and other congestion control issues found in conventional bittorrent over TCP, while providing reliable, ordered delivery.

Resources:
- Official [BEP29](http://www.bittorrent.org/beps/bep_0029.html)
- [Micro Transport Protocol](http://en.wikipedia.org/wiki/Micro_Transport_Protocol) on Wikipedia
- µTorrent [page](http://www.utorrent.com/intl/fr/help/guides/utp) on uTP
- [libutp](https://github.com/bittorrent/libutp) standard, open source C++ implementation used by many bittorrent clients including µTorrent
- Solid [paper](http://norswap.eu/papers/utp.pdf) describing differences between standard bittorrent protocol and µTP

mvayngrib:
i was experimenting with a combination of utp and bittorrent dht's implied_port option to see if I could get peers behind a NAT to work well. It wasn't quite enough. To take advantage of implied_port, you need the dht and the peer wire on the same port, which requires socket multiplexing, something vanilla dgram sockets don't support. They allow you to bind the same port, but only one socket will receive messages. 

To reuse the dht's port for the peer wire, I hacked together [sock-plex](https://www.npmjs.com/package/sock-plex), and it seemed to work well - bittorrent-dht & bittorrent-swarm (with utp) were sharing a port, and the wire was able to take advantage of the hole punching done by the dht. I was using it in a p2p otr chat module [zlorp](https://www.npmjs.com/package/zlorp) and a torrent-based replication module [bitkeeper-js](https://github.com/tradle/bitkeeper-js) with pretty good results. 

Right now sock-plex is not very subtle (it rewires dgram), but if this is an interesting direction, I can teach it some manners. Is this something interesting for webtorrent?

heri16:
Surprised that this feature hasn't been added. Likely the cause of this project's decline.

feross:
This landed and was released in v0.109.0

PR #1328, opened 2018-03-07
Add uTP support 

pldubouilh:
Work in progress implementation of uTP (issue #68). It uses [utp-native](https://www.npmjs.com/package/utp-native).

Also seems to greatly help seeding - after running it for an hour on a public distro torrent, it seems to seed much more than with plain TCP BT :ok_hand:

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,7 +58,8 @@ If `opts` is specified, then the default options (shown below) will be overridde
   peerId: String|Buffer,   // Wire protocol peer ID (default=randomly generated)
   tracker: Boolean|Object, // Enable trackers (default=true), or options object for Tracker
   dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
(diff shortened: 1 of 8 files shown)


pldubouilh:
Thanks - now I think it's good to be reviewed.

Just wondering about that appveyor test issue though, might be related to mafintosh/utp-native#20 but I'm not too sure, the test kind of fails in a cryptic manner

G-Ray:
Utp support would be great! Any update on this @DiegoRBaquero @feross ?

pldubouilh:
Happy to resolve the conflict if utp is still of interest, otherwise I'll just close the PR.

DiegoRBaquero:
I like this one. @feross ?

DiegoRBaquero:
Could you update so we can move this forward?

pldubouilh:
closing for #1486

PR #1345, opened 2018-04-01
[docs/api] Removed wild semicolons for consistency

now-raymond:
Removed semicolons to be consistent with JS Standard Style.


DiegoRBaquero:
Thank you!

Issue #889, opened 2016-08-24
warning: possible EventEmitter memory leak detected


roccomuso:
* WebTorrent version: 1.2.3 (webtorrent 0.96.0)
* Node.js version: 4.4.7

This the warning i'm getting:

```
(node) warning: possible EventEmitter memory leak detected. 11 ready listeners added. Use emitter.setMaxListeners() to increase limit.
Trace
    at Torrent.addListener (events.js:239:17)
    at Torrent.once (events.js:265:8)
    at Request.onChunk [as callback] (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/lib/torrent.js:1488:34)
    at Wire._callback (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/bittorrent-protocol/index.js:604:11)
    at Wire._onPiece (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/bittorrent-protocol/index.js:508:8)
    at Wire._onMessage (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/bittorrent-protocol/index.js:674:19)
    at Wire._write (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/bittorrent-protocol/index.js:592:10)
    at doWrite (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/readable-stream/lib/_stream_writable.js:319:64)
    at writeOrBuffer (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/readable-stream/lib/_stream_writable.js:308:5)
    at Wire.Writable.write (/usr/local/lib/node_modules/webtorrent-cli/node_modules/webtorrent/node_modules/readable-stream/lib/_stream_writable.js:246:11)

```

DiegoRBaquero:
So requests are being delayed to the ready event.

If we are already requesting, doesn't that mean we have metadata? @feross

feross:
We actually start the discovery process right away, even before the `'ready'` event fires, which is why we need https://github.com/feross/webtorrent/blob/c5eda255ab89efe8c1d62589a5df4ee7c146397b/lib/torrent.js#L1487-L1488.

Maybe we should wait for the metadata?

feross:
[...]

PR #1350, opened 2018-04-12
Defer Store Verification

KayleePop:
This PR is to fix the problem demonstrated here: https://jsfiddle.net/1g80d0aj/18/
The entire store needs to be verified before any files in the torrent can be accessed or streamed. This leads to a massive delay in appending the video when using a store with a significant read time instead of the default in-memory store. ([this is the indexeddb chunk store used in the demo](https://github.com/xuset/indexeddb-chunk-store)).

To fix this, I made the 'ready' event fire before store verification finishes, and any calls to _request() are cancelled if that piece hasn't yet been checked for in the store.

Critical pieces are verified immediately instead of being waited for in the parallelLimit() queue, ~~and requests for critical pieces aren't cancelled before verification like normal pieces~~. This allows file streams to work as quickly as possible.

(EDIT: Critical pieces are now also cancelled before verification)

Here's the same demo but using a minified build from this PR instead of the latest webtorrent build. https://jsfiddle.net/6r2hgxf5/11/

&nbsp;

Viewing this one commit at a time might make the changes clearer.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -100,6 +100,8 @@ function Torrent (torrentId, client, opts) {
   this._amInterested = false
   this._selections = []
   this._critical = []
+  this._verified = []
+  this._verifying = []
 
   this.wires = [] // open wires (added *after* handshake)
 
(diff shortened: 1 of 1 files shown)


whothisbeme:
LGTM, and it does resolve the issues with #889 as well.

I've tested the changes in this PR with the latest master branch and it seems to work.

Was there a reason this wasn't merged? Other than the conflicts? @KayleePop @feross @DiegoRBaquero

Thank you for this PR and your time spent 👍

Issue #1354, opened 2018-04-20
Changelog is massively out of date


feross:
The changelog is massively out of date. This is mostly my fault 🙈

PR incoming to fix this up!

DiegoRBaquero:
Never used, always wanted to. Willing to explore options and find the answer to the questions 

Diego

PR #1355, opened 2018-04-20
Update changelog for versions back to 2016-09-17

feross:
Fixes: https://github.com/webtorrent/webtorrent/issues/1354


feross:
Took longer than I would have liked ;)

PR #1356, opened 2018-04-20
Update list of WebTorrent talks

feross:
If anyone knows of other talks that include WebTorrent in some way, please send PRs!


DiegoRBaquero:
Haven't seen any other, I was supposed to give another last year but was too busy in the second semester.

PR #1358, opened 2018-04-21
expose webtorrent version

KayleePop:
``` javascript
let version = require('webtorrent').VERSION
```


feross:
This was released in v0.99.1 on 2018-04-24

PR #1363, opened 2018-04-25
Add .npmignore; Move internal scripts from bin/ to scripts/

feross:
By convention the bin/ folder is for user-facing command line scripts. Let's use scripts/ for internal scripts that don't need to be published.


feross:
Released as 0.99.2

feross:
FYI, I made the same change in https://github.com/webtorrent/webtorrent-cli but forgot to send a PR

DiegoRBaquero:
Gotcha

Issue #1000, opened 2016-12-17
client.remove(torrentId, [function callback (err) {}]) not removing temp folder


davidgatti:
The documentation states:

> Remove a torrent from the client. Destroy all connections to peers and delete all saved file data. If callback is specified, it will be called when file data is removed.

The issue that I'm having is with the following statement: `and delete all saved file data`, this seams not to be true. If I check the content of the following folder `/tmp/webtorrent` after firing this method, the folder of the selected torrent is not being delete. 

Is this a bug, feature or I'm doing something wrong? :)

* WebTorrent version: 0.98.0
* Node.js version: v6.4.0
* Browser name/version (if using WebTorrent in the browser): Using NodeJS module.

feross:
You're right, this is not behaving as documented. The removal of files on the filesystem is unimplemented. In the meantime, I suggest using the `rimraf` module to delete the entire `/tmp/webtorrent` folder, or just the `/tmp/webtorrent/<infohash>` folder for a single torrent.

davidgatti:
OK, good to know :) it be nice if in the documentation there as a sentence like the following:

`for not to delete the folder related to the torrent, please implement your own way` - or somethign like that to give an idea what to expect :)

austin-rausch:
The wording seems kind of confusing, if this were implemented would this remove both the tmp folder as well as the path if specified? Or just clean up the tmp folder. I'm assuming it is the latter but I just wanted clarification.

PR #1364, opened 2018-04-25
Option to destroy store on torrent removal (delete files from disk)

KayleePop:
Intended to replace https://github.com/webtorrent/webtorrent/pull/1102

Should work with custom stores like indexeddb as well, but I didn't test that.

API is this:

``` javascript
client.remove(torrentId, {destroyStore: true}, () => console.log('files deleted'))

torrent.destroy({destroyStore: true}, () => console.log('files deleted'))
```

fixes https://github.com/webtorrent/webtorrent/issues/1000

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -163,12 +163,14 @@ destroyed and all torrents are removed and cleaned up when this occurs.
 
 Always listen for the 'error' event.
 
-## `client.remove(torrentId, [function callback (err) {}])`
+## `client.remove(torrentId, [opts], [function callback (err) {}])`
 
-Remove a torrent from the client. Destroy all connections to peers and delete all saved
-file data. If `callback` is specified, it will be called when file data is removed.
+Remove a torrent from the client. Destroy all connections to peers and delete all saved file metadata.
 
-*Note: This method does not currently delete torrent data (in e.g. `/tmp/webtorrent/...`, see the `path` option to `client.add`). Until this is fixed, please implement it yourself (consider using the `rimraf` npm package).
+If `opts.destroyStore` is truthy, `store.destroy()` will be called, which will delete the torrent's files from the disk.
+
+If `callback` is provided, it will be called when the torrent is fully destroyed,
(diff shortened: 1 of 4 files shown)


DiegoRBaquero:
As opts is optional, if you don't included but you have the callback function, this would break the API. The way to fix it is this: https://github.com/webtorrent/webtorrent/pull/1102/files#diff-168726dbe96b3ce427e7fedce31bb0bcR383

Why would you like to replace #1102?

KayleePop:
Both client.remove() and torrent.destroy() pass both opts and the callback into torrent._destroy before they're used, and I cover that scenario there https://github.com/webtorrent/webtorrent/pull/1364/files#diff-65b0a569f99e647ae33e8c31b9bf72baR647

If the API was broken like that, then these tests would fail https://github.com/webtorrent/webtorrent/blob/master/test/torrent-destroy.js#L19-L22

Should I handle it in all three functions for clarity?

&nbsp;

Yeah, looking back at https://github.com/webtorrent/webtorrent/pull/1102 it seems to almost the same as this with minor differences.

I don't remember why I wanted to replace it. I think I didn't look closely enough at the changes and thought it was abandoned. I should've asked there first.

Should I close this? It's no big deal. 

The only real differences between the two are that this one tests for both API functions and works with all custom stores. I could submit a review on the other one for that.

guanzo:
Any progress on this? I'm using WebTorrent in the browser and would like to delete file data with `client.remove()/torrent.destroy()`

KayleePop:
You can do this without this PR:

```js
torrent.destroy((err) => {
  if (err) throw err

  torrent.store.destroy((err) => {
    if (err) throw err

    console.log('destroyed and files removed')
  })
})
```

You might have to do it in the opposite order (destroy its store, then the torrent).

You also only need to delete it if you use a custom chunk store like [idbkv-chunk-store](https://github.com/KayleePop/idbkv-chunk-store). Otherwise, in the browser, the data is never written to the disk and only temporarily stored in memory.

alxhotel (approved):
LGTM :+1:

PR #1376, opened 2018-05-05
ProBot: Disable the lock comment

feross:
Sorry for all the spam! Let's disable the comment when the bot locks the issue. It's not really necessary to notify all repo watchers whenever an issue is locked (even once it finishes locking all old issues and the flood slows down)


transitive-bullshit:
Thank you! 💯

DiegoRBaquero:
Thank youuuuu

Diego

Issue #1267, opened 2018-01-18
Access-Control-Allow-Methods should be GET,HEAD


feross:
In the webtorrent server, the `Access-Control-Allow-Methods` header should not set it's value to `GET,HEAD,PUT,PATCH,POST,DELETE` since only `GET,HEAD` are actually handled. The rest return an error when requested.

Let's remove it for defense in depth reasons.

diracdeltas:
++

feross:
Just sent a PR for this: https://github.com/webtorrent/webtorrent/pull/1383

PR #1383, opened 2018-05-10
Access-Control-Allow-Methods should be GET,HEAD

feross:
Fixes: https://github.com/webtorrent/webtorrent/issues/1267

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -111,7 +111,7 @@ function Server (torrent, opts) {
     function serveOptionsRequest () {
       res.statusCode = 204 // no content
       res.setHeader('Access-Control-Max-Age', '600')
-      res.setHeader('Access-Control-Allow-Methods', 'GET,HEAD,PUT,PATCH,POST,DELETE')
+      res.setHeader('Access-Control-Allow-Methods', 'GET,HEAD')
 
       if (req.headers['access-control-request-headers']) {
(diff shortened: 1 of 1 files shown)


feross:
@diracdeltas - does this look good to you?

feross:
Sauce Labs outage is causing tests to fail.

lib/server.js
```diff
@@ -111,7 +111,7 @@ function Server (torrent, opts) {
     function serveOptionsRequest () {
       res.statusCode = 204 // no content
       res.setHeader('Access-Control-Max-Age', '600')
-      res.setHeader('Access-Control-Allow-Methods', 'GET,HEAD,PUT,PATCH,POST,DELETE')
+      res.setHeader('Access-Control-Allow-Methods', 'GET,HEAD')
```
diracdeltas: lgtm - do you actually need HEAD though?

feross: We do currently support HEAD, so it needs to stay in this header if we want it to keep working. You can see how we handle it here: https://github.com/webtorrent/webtorrent/blob/8c8613d6bcb2a48a7aa996ffb08d379c3e20bea0/lib/server.js#L209

I don't think it hurts to keep it.

feross:
Nice, thanks for the reviews :)

Issue #1395, opened 2018-05-17
Implement support for 'select only' (so) magnet link parameter


feross:
*Originally proposed by @Daniel-Abrecht in https://github.com/webtorrent/webtorrent-desktop/issues/83#issuecomment-389965535*

There's a BEP that we should probably support: http://www.bittorrent.org/beps/bep_0053.html

> This magnet extension introduces the concept of an optional select-only file index array in the magnet URI, so that torrent clients can automatically select specific file(s) for download after the metadata has been downloaded.

This could probably replace the `ix` parameter that we invented for this exact use case. Since this is an actual BEP, it's probably better to support this instead (though it didn't exist when we invented `ix`)

SilentBot1:
This has been implemented as of the commit https://github.com/webtorrent/webtorrent/commit/925bb881cfe90cc055f67e747df693f32d265733.

Daniel-Abrecht:
Wow, this was really fast. Thanks everyone.

feross:
Released as 0.100.0.

PR #1396, opened 2018-05-18
Implements BEP53 to allow file selection using &so in magnetURIs

SilentBot1:
This pull request implements [BEP53](http://www.bittorrent.org/beps/bep_0053.html) as requested in #1395.

If `&so=` is specified, all pieces are deselected and only the pieces which contain parts of the selected files are downloaded, if no pieces are specified, the entire file is left deselected.

The use of [parse-numeric-range](https://www.npmjs.com/package/parse-numeric-range) allows for ranges to be specified in three different formats which include using hyphens, which is what the BEP calls for, e.g. `1-5` which is equal to `[1, 2, 3, 4, 5]`, using two dots e.g. `1..5` which is also equal to `[1, 2, 3, 4, 5]` or by using three dots e.g. `1...5` which would equal `[1, 2, 3, 4]`.

Any feedback or suggestions are welcome.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -30,6 +30,7 @@ var speedometer = require('speedometer')
 var uniq = require('uniq')
 var utMetadata = require('ut_metadata')
 var utPex = require('ut_pex') // browser exclude
+var parseRange = require('parse-numeric-range')
 
 var File = require('./file')
 var Peer = require('./peer')
@@ -482,11 +483,6 @@ Torrent.prototype._onMetadata = function (metadata) {
     })
   }
 
-  // start off selecting the entire torrent with low priority
-  if (self.pieces.length !== 0) {
-    self.select(0, self.pieces.length - 1, false)
-  }
-
   self._rarityMap = new RarityMap(self)
 
   self.store = new ImmediateChunkStore(
@@ -508,6 +504,22 @@ Torrent.prototype._onMetadata = function (metadata) {
   self.files = self.files.map(function (file) {
     return new File(self, file)
   })
+
+  // Only select specified files (BEP53)
+  if (self.so) {
+    var specificFiles = parseRange.parse(self.so)
+    // https://github.com/webtorrent/webtorrent/issues/164
+    self.deselect(0, self.pieces.length - 1, false)
+
+    self.files.forEach(function (v, i) {
+      specificFiles.indexOf(i) === -1 ? self.files[i].deselect() : self.files[i].select(true)
+    })
+  } else {
+    // start off selecting the entire torrent with low priority
+    if (self.pieces.length !== 0) {
+      self.select(0, self.pieces.length - 1, false)
+    }
+  }
 
   self._hashes = self.pieces
 
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -505,14 +505,12 @@ Torrent.prototype._onMetadata = function (metadata) {
     return new File(self, file)
   })
 
-  // Only select specified files (BEP53)
+  // Select only specified files (BEP53) http://www.bittorrent.org/beps/bep_0053.html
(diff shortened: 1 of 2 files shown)


lib/torrent.js
```diff
@@ -509,6 +505,22 @@ Torrent.prototype._onMetadata = function (metadata) {
     return new File(self, file)
   })
 
+  // Only select specified files (BEP53)
+  if (self.so) {
+    var specificFiles = parseRange.parse(self.so)
+    // https://github.com/webtorrent/webtorrent/issues/164
+    self.deselect(0, self.pieces.length - 1, false)
```
feross: I believe that this line should not be necessary since you moved the `self.select(0, self.pieces.length - 1, false)` call to an `else` block below. So, what is this actually deselecting?

lib/torrent.js
```diff
@@ -509,6 +505,22 @@ Torrent.prototype._onMetadata = function (metadata) {
     return new File(self, file)
   })
 
+  // Only select specified files (BEP53)
+  if (self.so) {
+    var specificFiles = parseRange.parse(self.so)
```
feross: Can we name this variable `selectOnlyFiles`, since "so" stands for "select only"?

lib/torrent.js
```diff
@@ -509,6 +505,22 @@ Torrent.prototype._onMetadata = function (metadata) {
     return new File(self, file)
   })
 
+  // Only select specified files (BEP53)
```
feross: Can you add an actual link to the BEP in this comment?

lib/torrent.js
```diff
@@ -509,6 +505,22 @@ Torrent.prototype._onMetadata = function (metadata) {
...
+  if (self.so) {
+    var specificFiles = parseRange.parse(self.so)
+    // https://github.com/webtorrent/webtorrent/issues/164
+    self.deselect(0, self.pieces.length - 1, false)
+
+    self.files.forEach(function (v, i) {
+      specificFiles.indexOf(i) === -1 ? self.files[i].deselect() : self.files[i].select(true)
```
feross: Unless I'm missing something, it shouldn't be necessary to call `deselect()` since all files start out deselected. Can we change this to just call `select()`?

Also, can you using `array.includes()` instead of `array.indexOf()` since it's clearer?

lib/torrent.js
SilentBot1: Can do, included in next commit.

lib/torrent.js
SilentBot1: Of course, that would also make it more clear for people searching the code.

lib/torrent.js
SilentBot1: Initially I had the `if (self.so)` statement under `self.select` all without the `else` statement, this is leftover code from that which can be removed, removed in next commit.

lib/torrent.js
SilentBot1: You're correct, the `self.files[i].deselect()` serves no purpose as pieces start out deselected, this will be removed in the next commit.

Replacing `specificFiles.indexOf(i)` with `selectOnlyFiles.includes(i)` for clarity will also be in the next commit.

Thank you for the review :)

lib/torrent.js
```diff
@@ -505,14 +505,12 @@ Torrent.prototype._onMetadata = function (metadata) {
     return new File(self, file)
   })
 
-  // Only select specified files (BEP53)
+  // Select only specified files (BEP53) http://www.bittorrent.org/beps/bep_0053.html
```
SilentBot1: Addresses https://github.com/webtorrent/webtorrent/pull/1396#discussion_r189165523.

feross:
@SilentBot1 Thanks for making the changes and merging. Released as 0.100.0.

Issue #524, opened 2015-12-11
Seeding re-stores chunks even if they are already stored


DiegoRBaquero:
When you seed a file, it doesn't matter if the pieces are already on the store, it will create a ChunkStoreWriteStream and re-store them.

The verifying happens in Torrent._onMetadata, called on Torrent._onSwarnListening
The re-storing happens in Torrent.load, called in WebTorrent.seed

```log
webtorrent:torrent got metadata +5ms
webtorrent:torrent verifying existing torrent data +3ms
webtorrent:torrent piece verified 0 +34ms
webtorrent:torrent piece verified 1 +1ms
webtorrent:torrent piece verified 2 +1ms
webtorrent:torrent done verifying +2ms
webtorrent:torrent on store +1ms
webtorrent:torrent select 0-2 (priority 0) +1ms
webtorrent:torrent file done: PMC-InformelaMuestra.pdf +3ms
webtorrent:torrent torrent done: 307f97a2e9264a5c17d04dd83aae6069d249a287 +1ms
bittorrent-tracker send `complete` +0ms
function onData(chunk) {
    if (self.destroyed) return
    store.put(index, chunk)
    index += 1
  }
put 0 from function (index, buf, cb) {
  console.info(arguments.callee.caller.toString())
  var self = this
  self.mem[index] = buf
  self.store.put(index, buf, function (err) {
    self.mem[index] = null
    if (cb) cb(err)
  })
}
function onData(chunk) {
    if (self.destroyed) return
    store.put(index, chunk)
    index += 1
  }
put 1 from function (index, buf, cb) {
  console.info(arguments.callee.caller.toString())
  var self = this
  self.mem[index] = buf
  self.store.put(index, buf, function (err) {
    self.mem[index] = null
    if (cb) cb(err)
  })
}
function onData(chunk) {
    if (self.destroyed) return
    store.put(index, chunk)
    index += 1
  }
put 2 from function (index, buf, cb) {
  console.info(arguments.callee.caller.toString())
  var self = this
  self.mem[index] = buf
  self.store.put(index, buf, function (err) {
    self.mem[index] = null
    if (cb) cb(err)
  })
}
webtorrent:torrent torrent done: 307f97a2e9264a5c17d04dd83aae6069d249a287 +54ms
bittorrent-tracker send `complete` +1ms
````

feross:
You're right. We can: 

1. Eliminate verifying for all `input` types to `client.seed`, since it'll obviously be correct.

[...]

Issue #1237, opened 2017-12-09
Should local file seeds pieces be verified


Ciao121:
Hi,
I'm not understanding why, when I seed a local file, pieces verification occurs. It this needed?
Thank you!

feross:
@Ciao121 This is not really needed, but we're probably just going through that code path no matter what. PR welcome to fix this.

PR #1398, opened 2018-05-18
Optimize file seeding

KayleePop:
- Stop verifying the hashes of files passed into client.seed()  _(we just calculated the hashes that are being checked)_
- Don't call torrent.load() when seeding an FS filepath. _(fs-chunk-store is initialized using the same paths, so the files already exist in the store)_

Fixes https://github.com/webtorrent/webtorrent/issues/524 and https://github.com/webtorrent/webtorrent/issues/1237


feross:
Thanks for the PR. I'll review it as soon as I get the chance. In the meantime, reviews from other collaborators are welcome and appreciated.

DiegoRBaquero:
Hey @KayleePop , willing to update to the new code base? taking into account that 
 https://github.com/webtorrent/webtorrent/pull/1434 was merged

KayleePop:
Rebased

PR #1419, opened 2018-06-17
Add support for NAT traversal techniques

oleiba:
Extending the PR #1051, by allowing to specify specific protocol (TCP/UDP) for a NAT traversal technique.
This is important in particular for the case of registering a DHT port as UDP.

I here add the original author's comments for clarity:

> Here is an implementation of NAT traversal as discussed in #195 using https://github.com/indutny/node-nat-upnp and https://github.com/TooTallNate/node-nat-pmp.
> 
> I have also ecountered https://github.com/level451/pmp that claims to make both UPnP ad NAT-PMP but I don't have tested it.
> 
> This feature is implemented in a new module for node env only and provided by a single instance shared by the webtorrent clients because NAT-PMP needs to open a specific port to work.
> 
> I checked the UPnP part with my router by it does not support NAT-PMP protocol.
> 
> The mappings are sent both with UPnP and NAT-PMP whe available and refreshed every 30 min.
> 
> Reviews are welcome!


oleiba:
Closing in favor of #1437

PR #1431, opened 2018-07-14
Persist DHT nodes

### LINKED ISSUE #72: Persist the DHT to disk so it can be used between restarts
This should be really simple using dht.toArray(), as described here: https://github.com/feross/bittorrent-dht#arr--dhttoarray

bookmoons:
Persists DHT nodes to disk every 15 minutes.
Loads persisted nodes on construction.
Enabled by default.

Default save file is `dht.json`, under crossplatform app data folder provided by `app-data-folder`.

Adds option `dhtState` to configure:
* false disables
* true enables with default path
* String specifies path to custom save file

Closes #72.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,7 +58,8 @@ If `opts` is specified, then the default options (shown below) will be overridde
   peerId: String|Buffer,    // Wire protocol peer ID (default=randomly generated)
   tracker: Boolean|Object,  // Enable trackers (default=true), or options object for Tracker
   dht: Boolean|Object,      // Enable DHT (default=true), or options object for DHT
-  dhtState: Boolean|String, // Persist DHT nodes (default=true), or save file path
+  persistDht: Boolean,      // Persist DHT nodes (default=true)
+  persistDhtPath: String,   // DHT persist save file (default=dht.json under OS appdata dir)
   webSeeds: Boolean         // Enable BEP19 web seeds (default=true)
 }
 ```
--- a/docs/api.md
+++ b/docs/api.md
@@ -53,12 +53,13 @@ If `opts` is specified, then the default options (shown below) will be overridde
 
 ```js
 {
-  maxConns: Number,        // Max number of connections per torrent (default=55)
-  nodeId: String|Buffer,   // DHT protocol node ID (default=randomly generated)
-  peerId: String|Buffer,   // Wire protocol peer ID (default=randomly generated)
-  tracker: Boolean|Object, // Enable trackers (default=true), or options object for Tracker
-  dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
-  webSeeds: Boolean        // Enable BEP19 web seeds (default=true)
+  maxConns: Number,         // Max number of connections per torrent (default=55)
+  nodeId: String|Buffer,    // DHT protocol node ID (default=randomly generated)
+  peerId: String|Buffer,    // Wire protocol peer ID (default=randomly generated)
+  tracker: Boolean|Object,  // Enable trackers (default=true), or options object for Tracker
+  dht: Boolean|Object,      // Enable DHT (default=true), or options object for DHT
+  dhtState: Boolean|String, // Persist DHT nodes (default=true), or save file path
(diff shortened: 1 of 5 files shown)


bookmoons:
Hi guys. Aiming to implement #72 here, persisting DHT nodes. All local tests succeed.

It uses [app-data-folder](https://github.com/payne8/app-data) to get a good crossplatform default location for variable app data.

Loading is unfortunately synchronous. Making the DHT initialization async breaks the client for some reason unclear to me. Usage expects to call methods immediately, but they start failing if DHT loading is not finished. Even with that ready event delayed until everything is ready. If there's an easy way to fix that I can update.

bookmoons:
I'm proposing a new option `dhtState` here. Default is true, which enables persistence with the default path. A string specifies a custom path. False disables.

DiegoRBaquero:
I like this, but we need some tests. Would you add them?

bookmoons:
Thanks for looking it over. I've added 2 tests, 1 for saving, 1 for loading. They're in [`persist-dht-nodes.js`](https://github.com/webtorrent/webtorrent/pull/1431/files#diff-66984283e6d63f8338a135559a0484d1). They test with temp files. All tests are passing, including the new ones.

Started splitting things out into `lib/dhtpersist.js` and ended up not needing it. I left it in; think it makes `index.js` cleaner.

I added an API method `saveDhtstate`, to make it possible to get at it in the test. Not sure if that's a good idea. It starts here:
https://github.com/webtorrent/webtorrent/pull/1431/files#diff-168726dbe96b3ce427e7fedce31bb0bcR439

index.js
```diff
@@ -19,6 +20,7 @@ var randombytes = require('randombytes')
 var speedometer = require('speedometer')
 var zeroFill = require('zero-fill')
 
+var dhtPersist = require('./lib/dhtpersist') // browser exclude
```
KayleePop: It says browser exclude, but it's not actually excluded in the browser field of package.json?

like this 

``` json
"browser": {
  "./lib/dhtpersist": false
}
```

docs/api.md
```diff
@@ -53,12 +53,13 @@ If `opts` is specified, then the default options (shown below) will be overridde
...
-  peerId: String|Buffer,   // Wire protocol peer ID (default=randomly generated)
-  tracker: Boolean|Object, // Enable trackers (default=true), or options object for Tracker
-  dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
-  webSeeds: Boolean        // Enable BEP19 web seeds (default=true)
+  maxConns: Number,         // Max number of connections per torrent (default=55)
+  nodeId: String|Buffer,    // DHT protocol node ID (default=randomly generated)
+  peerId: String|Buffer,    // Wire protocol peer ID (default=randomly generated)
+  tracker: Boolean|Object,  // Enable trackers (default=true), or options object for Tracker
+  dht: Boolean|Object,      // Enable DHT (default=true), or options object for DHT
+  dhtState: Boolean|String, // Persist DHT nodes (default=true), or save file path
```
KayleePop: I think this would be clearer as two separate variables. A boolean for whether to persist the state, and a file path for persisting the state

Maybe named (doesn't matter)
``` js
{
  persistDht: Boolean,
  persistDhtPath: String
}
```

index.js
bookmoons: Misunderstood what that was about. I was thinking it was some control message to the bundler. I've updated.

docs/api.md
bookmoons: I've updated to split them that way.

bookmoons:
Thanks for the review. Didn't even show up in my activity list. Will have to find a better way to watch for that. I've addressed both of them.

bookmoons:
I guess one of the automated builds failed. It looks like it failed on `npm install`. Is it possible that was a temporary network error?

DiegoRBaquero:
I don't like 3 new dependencies, any way these could be integrated as libs?

Also, should we enable this by default? @feross ?

bookmoons:
>I don't like 3 new dependencies, any way these could be integrated as libs?

I'm dug into something else for a while here, but will look at this when I have a chance.

Hanoosh:
any news on this

PR #1436, opened 2018-07-20
Torrent to pass discovery interval to torrent-discovery

oleiba:
`torrent-discovery` supports accepting `intervalMs` in its constructor, for changing the interval for discovering new peers (e.g. querying the DHT).
However, passing this argument is not yet supported in the main WebTorrent client which uses this package. This PR enables it.

Sometimes it makes a lot of sense to adjust the interval of querying the DHT for updates, for example if you want shorter discovery intervals than the 15 minutes default.


DiegoRBaquero:
I'll take this with added documentation and tests

PR #1455, opened 2018-08-01
Add package-lock.json to gitignore

KayleePop:
Seems like a good idea since npm automatically creates it.

At the very least it would remove a minor annoyance


feross:
Yep, this makes sense.

PR #1456, opened 2018-08-02
Indexeddb chunk store in browser

KayleePop:
~~Here's the chunk store used https://github.com/KayleePop/idbkv-chunk-store
which is built on this IndexedDB wrapper https://github.com/KayleePop/idb-kv~~

~~`npm run size` shows this branch's bundle to be `85,873 bytes` which compared to the `85,252 bytes` of master is a .7% increase or 621 bytes larger~~

&nbsp;

Here's a deployment of instant.io using this PR's code for demo and smoke testing https://instant-io-idbkv.glitch.me/ (glitch.com is awesome)

Things to test (please share results :smile: ):
- Very large files
- Concurrent downloads of different torrents
- Reload tab to load persisted data from disk
  - After download has finished
  - While download is still going
- Multiple browser tabs (they use the same indexeddb database)
- RAM usage
- Speed compared to https://instant.io
  - Speed of completing download
  - downloading the file to disk
- Streaming
- Different browsers

EDIT: The default-chunk-store is no longer changed in this PR

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -45,7 +45,8 @@ const PIPELINE_MAX_DURATION = 1
 const RECHOKE_INTERVAL = 10000 // 10 seconds
 const RECHOKE_OPTIMISTIC_DURATION = 2 // 30 seconds
 
-const FILESYSTEM_CONCURRENCY = 2
+// IndexedDB chunk stores used in the browser benefit from maximum concurrency
+const FILESYSTEM_CONCURRENCY = process.browser ? Infinity : 2
 
 const RECONNECT_WAIT = [ 1000, 5000, 15000 ]
 
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -459,7 +459,8 @@ class Torrent extends EventEmitter {
           length: file.length,
           offset: file.offset
         })),
-        length: this.length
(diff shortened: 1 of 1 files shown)


KayleePop:
One thing I noticed is that everything breaks in Firefox's private browsing mode (ctrl+shift+P). This is because Firefox simply disables indexeddb in that mode.

Maybe this could be fixed by allowing immediate-chunk-store to act like memory-chunk-store if there are errors returned from put()

Right now it's deleting the data if the put() callback is called with an error https://github.com/feross/immediate-chunk-store/blob/master/index.js#L20

KayleePop:
A potential optimization is to increase or remove the concurrency limit on verification here https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L594 (FILESYSTEM_CONCURRENCY is set to 2)

[idb-kv](https://github.com/KayleePop/idb-kv) uses batching to improve the performance of indexeddb as each transaction adds a lot of overhead. Increasing the limit would allow more reads to make it into each batch, and it might speed up store verification.

KayleePop:
What about just the name option and concurrency changes?

KayleePop:
I removed the commit that changed the default store.

PR #1461, opened 2018-08-07
Added reference to P2P Media Loader project

mrlika:
Hi. We just open-sourced our P2P live and VOD streaming project. It uses WebTorrent tracker for signaling as one of the key components of P2P network.


mrlika:
Hi. We just opensourced our P2P live and VOD streaming project. It uses WebTorrent tracker for signaling as one of the key components of P2P network.

mrlika:
Any issues with this pull request?

PR #1464, opened 2018-08-09
Small debugging improvements

jhiesey:
These were useful when I was doing some debugging and probably will help others too.


jhiesey:
@KayleePop yup that's exactly the issue I hit

PR #1476, opened 2018-08-18
modernize lib/file.js and calculate downloaded correctly

jimmywarting:
this fixes the calculation of `.files[i].downloaded`

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -36,18 +36,30 @@ class File extends EventEmitter {
 
   get downloaded () {
     if (!this._torrent.bitfield) return 0
-    let downloaded = 0
-    for (let index = this._startPiece; index <= this._endPiece; ++index) {
-      if (this._torrent.bitfield.get(index)) {
+
+    const { pieces, bitfield, pieceLength } = this._torrent
+    const { _startPiece:start, _endPiece:end } = this
+    const piece = pieces[start]
+
+    // Calculate first piece diffrently, it sometimes have a offset
+    let downloaded = bitfield.get(start)
+      ? pieceLength - this.offset
+      : Math.max(piece.length - piece.missing - this.offset, 0)
+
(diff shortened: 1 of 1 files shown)


lib/file.js
```diff
@@ -1,127 +1,123 @@
...
+  get downloaded () {
     if (!this._torrent.bitfield) return 0
-    var downloaded = 0
-    for (var index = this._startPiece; index <= this._endPiece; ++index) {
+    let downloaded = 0
+    for (let index = this._startPiece; index <= this._endPiece; ++index) {
       if (this._torrent.bitfield.get(index)) {
         // verified data
-        downloaded += (index === this._endPiece) ? this._torrent.lastPieceLength : this._torrent.pieceLength
+        downloaded += index === this._endPiece ? this._torrent.lastPieceLength : this._torrent.pieceLength
```
KayleePop: I think these parentheses make the code more readable.

DiegoRBaquero: I'd rather keep them

jimmywarting:
I also notice `file.download` is calculated wrong.

Using the Sintel torrent as an example...
https://instant.io/#08ada5a7a6183aae1e09d831df6748d566095a10
(this can be tested in the console=

It has 6 srt files and 1 small piece of the mp4 file in the first piece
after the first piece is downloaded you can see that the first files are complete but the `progress` and `downloaded` are incorrect

```js
const file = client.get('08ada5a7a6183aae1e09d831df6748d566095a10').files[0]
console.log(file.done) // true
console.log(file.progress) // 39.58777239709443 <- incorrect b/c of download
console.log(file.downloaded) // 65399
console.log(file.length) // 1652
```

I'm working on a fix for that now also

DiegoRBaquero:
Correct, please separate into two PRs, let's keep this one about modernizing the code.

jimmywarting:
ok, closing this then, new pr: #1478

PR #1477, opened 2018-08-18
added install size badge

jimmywarting:


DiegoRBaquero:
I feel like this is not needed. @feross ?

PR #1484, opened 2018-08-26
Add Cheff to demo apps in FAQ.

vasco3:
Adding [Cheff](https://github.com/vasco3/cheff) a web app for meal planning which is client-only so it uses webtorrent to sync recipes between user's devices.


vasco3:
Added more info in the description

vasco3:
@DiegoRBaquero that's correct. 

1. I store in localStorage as stringified JSON. 
2. On export I open a torrent seed and display the magnetURI as a QR Code. 
3. Then with another device I scan the QR Code to get the magnetURI and connect to the torrent.
4. Through the torrent I buffer the stringified JSON 
5. Then store in localStorage.

DiegoRBaquero:
Ah, nice ! cool project

vasco3:
Thanks man

feross:
@vasco3 Looks like this domain is taken over by a malware site now. Removing your site from the list. Please resend a PR if you fix the issue.

Issue #1443, opened 2018-07-27
Meta: ES Next


feross:
Let's convert all WebTorrent packages to ES6. (Continued from https://github.com/webtorrent/webtorrent/issues/444)

Note that this is a separate task from async-await (Promises) support.

- [x] webtorrent
- [x] bittorrent-dht
- [x] bittorrent-peerid
- [x] bittorrent-protocol
- [x] bittorrent-tracker
- [x] create-torrent
- [x] magnet-uri
- [x] parse-torrent
- [x] render-media
- [x] torrent-discovery
- [x] ut_metadata
- [x] ut_pex
- [x] webtorrent-fixtures
- [x] addr-to-ip-port
- [x] webtorrent-cli
- [x] webtorrent-hybrid
- [x] load-ip-set
- [x] ip-set

hgghyxo:
Do you plan to add support for ESModule like imports? 

```javascript
import { WebTorrent } from 'https://unpkg.com/webtorrent'
```

>Uncaught SyntaxError: The requested module 'https://unpkg.com/webtorrent' does not provide an export named 'WebTorrent'

DiegoRBaquero:
@hgghyxo I think you have to do `import * as WebTorrent from ...`

DiegoRBaquero:
Are you using webpack, babel?

hgghyxo:
neither, ESModule: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import
the `<script type="module" src="...">` thingy

DiegoRBaquero:
oh, then just use the normal import and it'll be available as a global.

There's no plan on supporting ES Modules atm

hgghyxo:
Webtorrent-hybrid was't a big project: https://github.com/webtorrent/webtorrent-hybrid/pull/86  
:)

jimmywarting:
PR's waiting to be merged

- [x] webtorrent
- [ ] bittorrent-dht
- [x] bittorrent-peerid
- [x] bittorrent-protocol
- [ ] bittorrent-tracker [#295](https://github.com/webtorrent/bittorrent-tracker/pull/295), [#296](https://github.com/webtorrent/bittorrent-tracker/pull/296), [#297](https://github.com/webtorrent/bittorrent-tracker/pull/297)
- [x] create-torrent
- [x] magnet-uri
- [x] parse-torrent
- [ ] render-media [#51](https://github.com/feross/render-media/pull/51)
[...]

PR #1485, opened 2018-08-26
modernize lib/rarity-map.js

jimmywarting:
#1443

### DIFF
--- a/lib/rarity-map.js
+++ b/lib/rarity-map.js
@@ -7,7 +7,7 @@ class RarityMap {
   constructor (torrent) {
     this._torrent = torrent
     this._numPieces = torrent.pieces.length
-    this._pieces = new Array(this._numPieces).fill(0)
+    this._pieces = new Array(this._numPieces)
 
     this._onWire = wire => {
       this.recalculate()
--- a/lib/rarity-map.js
+++ b/lib/rarity-map.js
@@ -1,121 +1,113 @@
-module.exports = RarityMap
 
 /**
  * Mapping of torrent pieces to their respective availability in the torrent swarm. Used
  * by the torrent manager for implementing the rarest piece first selection strategy.
  */
-function RarityMap (torrent) {
-  var self = this
+class RarityMap {
+  constructor (torrent) {
+    const self = this
 
-  self._torrent = torrent
-  self._numPieces = torrent.pieces.length
-  self._pieces = []
+    self._torrent = torrent
+    self._numPieces = torrent.pieces.length
+    self._pieces = new Array(self._numPieces).fill(0)
 
-  self._onWire = function (wire) {
+    self._onWire = wire => {
+      self.recalculate()
+      self._initWire(wire)
+    }
+    self._onWireHave = index => {
+      self._pieces[index] += 1
+    }
+    self._onWireBitfield = () => {
+      self.recalculate()
+    }
+
+    self._torrent.wires.forEach(wire => {
+      self._initWire(wire)
+    })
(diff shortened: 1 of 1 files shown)


lib/rarity-map.js
```diff
@@ -1,121 +1,113 @@
...
+class RarityMap {
+  constructor (torrent) {
+    const self = this
 
-  self._torrent = torrent
-  self._numPieces = torrent.pieces.length
-  self._pieces = []
+    self._torrent = torrent
+    self._numPieces = torrent.pieces.length
+    self._pieces = new Array(self._numPieces).fill(0)
```
DiegoRBaquero: Any benefits in doing this?

lib/rarity-map.js
```diff
@@ -1,121 +1,113 @@
...
 
 /**
  * Mapping of torrent pieces to their respective availability in the torrent swarm. Used
  * by the torrent manager for implementing the rarest piece first selection strategy.
  */
-function RarityMap (torrent) {
-  var self = this
+class RarityMap {
+  constructor (torrent) {
+    const self = this
```
DiegoRBaquero: We should stop using self

lib/rarity-map.js
```diff
@@ -1,121 +1,113 @@
...
     }
   }
 
-  if (candidates.length > 0) {
-    // if there are multiple pieces with the same availability, choose one randomly
-    return candidates[Math.random() * candidates.length | 0]
-  } else {
-    return -1
+  destroy () {
+    const self = this
```
DiegoRBaquero: Same here, no more self

lib/rarity-map.js
```diff
@@ -1,121 +1,113 @@
...
+   *
+   * @param {function} pieceFilterFunc
+   * @return {number} index of rarest piece, or -1
+   */
+  getRarestPiece (pieceFilterFunc) {
+    let candidates = []
+    let min = Infinity
+
+    for (let i = 0; i < this._numPieces; ++i) {
+      if (pieceFilterFunc && !pieceFilterFunc(i)) continue
```
DiegoRBaquero: Much better removing that trueFn

lib/rarity-map.js
jimmywarting: You probably don't need the initial `.fill(0)` in `new Array(self._numPieces).fill(0)` maybe just `new Array(self._numPieces)` will do

Reason why i did this was b/c `new Array(n)` is faster then writing to a literal array `[]`<br>And using `.fill(0)` is faster then writing to each index

skarmavbild 2018-08-26 kl 22 01 14

also b/c this looks cleaner then
```js
recalculate () {	
  this._pieces.fill(0)
}
```
then
```js
recalculate () {	
  for (let i = 0; i < this._numPieces; ++i) {	
    this._pieces[i] = 0	
  }
}
```

lib/rarity-map.js
jimmywarting: thanks

lib/rarity-map.js
```diff
@@ -1,121 +1,108 @@
...
-  var self = this
-
-  self._torrent = torrent
-  self._numPieces = torrent.pieces.length
-  self._pieces = []
+class RarityMap {
+  constructor (torrent) {
+    this._torrent = torrent
+    this._numPieces = torrent.pieces.length
+    this._pieces = new Array(this._numPieces).fill(0)
```
DiegoRBaquero: What is the fill here for? If it's not needed I think it will just consume O(n) to fill it and can be removed.

PR #1487, opened 2018-08-26
remove `self = this` in torrent.js

KayleePop:


lib/torrent.js
```diff
@@ -281,83 +278,78 @@ class Torrent extends EventEmitter {
...
-    } else if (self.xs) {
-      self._getMetadataFromServer()
+      this._onMetadata(this)
+    } else if (this.xs) {
+      this._getMetadataFromServer()
     }
   }
 
   _getMetadataFromServer () {
+    // to allow function hoisting
```
KayleePop: Hoisting was used here for readability, and hoisting is impossible with arrow functions

lib/torrent.js
```diff
@@ -1151,30 +1123,30 @@ class Torrent extends EventEmitter {
...
-      self._updateWire(wire)
+      this._updateWire(wire)
     }
   }
 
   /**
    * Attempts to update a peer's requests
    */
   _updateWire (wire) {
+    // to allow function hoisting
```
KayleePop: Same here. Arrow functions don't allow hoisting

lib/torrent.js
```diff
@@ -1433,62 +1402,65 @@ class Torrent extends EventEmitter {
...
+    wire.request(index, chunkOffset, chunkLength, async (err, chunk) => {
+      if (this.destroyed) return
+
+      const onUpdateTick = () => {
+        process.nextTick(() => { this._update() })
+      }
 
       // TODO: what is this for?
-      if (!self.ready) return self.once('ready', () => { onChunk(err, chunk) })
+      if (!this.ready) await new Promise(resolve => this.once('ready', resolve))
```
KayleePop: I used await and an async function here to allow this callback to be an anonymous arrow function. I think that's OK because async functions are supported in every browser that supports webRTC, and in node 8 which is LTS. (the tests wouldn't pass otherwise as they're set to the latest LTS in travis)

lib/torrent.js
```diff
@@ -1662,26 +1627,26 @@ class Torrent extends EventEmitter {
...
-      self._debug(
+      this._debug(
         'conn %s closed: will re-add to queue in %sms (attempt %s)',
         peer.addr, ms, peer.retries + 1
       )
 
-      const reconnectTimeout = setTimeout(function reconnectTimeout () {
-        const newPeer = self._addPeer(peer.addr)
+      const reconnectTimeout = setTimeout(() => {
+        const newPeer = this._addPeer(peer.addr)
```
KayleePop: Did this timeout callback need to be named? I don't think so.

DiegoRBaquero:
We are not implementing async/await just yet FYI. Everything for now should
work in node 6

lib/torrent.js
```diff
@@ -281,83 +278,78 @@ class Torrent extends EventEmitter {
...
-      port: self.client.torrentPort,
+      port: this.client.torrentPort,
       userAgent: USER_AGENT
     })
 
-    self.discovery.on('error', onError)
-    self.discovery.on('peer', onPeer)
-    self.discovery.on('trackerAnnounce', onTrackerAnnounce)
-    self.discovery.on('dhtAnnounce', onDHTAnnounce)
-    self.discovery.on('warning', onWarning)
```
KayleePop: Changed these to in-line arrow functions instead of being hoisted. IMO it's actually more readable that way, but it's a change in style.

KayleePop:
reverted changes to _request()

It just uses self still.

PR #1489, opened 2018-08-29
modernize

jimmywarting:
index.js


jimmywarting:
I thought that would be my next task

PR #1513, opened 2018-09-21
Try to reduce dl impact on slow computers (web)

Chocobozzz:
On slow computers with a fast network, web browsers may no longer respond because webtorrent is downloading chunks too fast.

The main issue is that you cannot stream a video torrent: the web browser does not have time to play the video, it is too busy downloading the video chunks.

On web browser supporting [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback#See_also), we download chunks with a lower priority allowing the web browser to execute other tasks (like playing the video).

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1138,7 +1138,17 @@ class Torrent extends EventEmitter {
     const ite = randomIterate(this.wires)
     let wire
     while ((wire = ite())) {
-      this._updateWire(wire)
+      this._updateWireWrapper(wire)
+    }
+  }
+
+  _updateWireWrapper (wire) {
+    const self = this
+
+    if (typeof window !== 'undefined' && typeof window.requestIdleCallback === 'function') {
+      window.requestIdleCallback(function () { self._updateWire(wire) })
+    } else {
+      self._updateWire(wire)
     }
   }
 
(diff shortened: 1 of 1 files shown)


ThaUnknown:
this kind of annoys me, because it's a REALLY good change, that benefits slow computers, but on the other hand it's one of the heaviest functions to run, so for fast computers some 20% of the CPU time goes to scheduling this
image

ThaUnknown:
the old code for this was:
```js
    const ite = randomIterate(this.wires)
    let wire
    while ((wire = ite())) {
      this._updateWireWrapper(wire)
    }
  }

  _updateWireWrapper (wire) {
    const self = this

    if (typeof window !== 'undefined' && typeof window.requestIdleCallback === 'function') {
      window.requestIdleCallback(function () { self._updateWire(wire) })
    } else {
      self._updateWire(wire)
    }
  }
  ```
  
I'm suggesting we flip it, that requestIdleCallback is called before random iterate, then after the callback all the peers are updated, this should still help performance while not dwindling ultra-high data transfer speeds:
```js
  /**
   * Heartbeat to update all peers and their requests.
   */
  _update () {
    if (IDLE_CALLBACK) {
      IDLE_CALLBACK(() => this._updateWireWrapper(), { timeout: 250 })
    } else {
      this._updateWireWrapper()
    }
  }

  _updateWireWrapper () {
    if (this.destroyed) return
    // update wires in random order for better request distribution
    const ite = randomIterate(this.wires)
    let wire
    while ((wire = ite())) {
      this._updateWire(wire)
    }
  }
  ```
  
@Chocobozzz do you think this will still accomplish its original goal? There aren't really many reproduction steps provided, and I haven't ran into the issue you're describing with this disabled, so this is kind of a blind change

PR #1577, opened 2019-02-04
chore: pin addr-to-ip-port to 1.5.0

piresten:
webtorrent


DiegoRBaquero:
Latest is not breaking. Should be safe to keep.

PR #1578, opened 2019-02-04
chore(package): update debug to version 4.1.1

piresten:
webtorrent


DiegoRBaquero:
Hi @piresten ! Thanks for the PR, but this doesn't need to be merged as minor and patch version are covered. Greenkeeper will only advise if an specific version breaks the build.

Issue #1552, opened 2018-11-23
incorrect value of file progress. file progress less than 0


De-Zol:
**What version of WebTorrent?**
0.102.4

**What operating system and Node.js version?**
Window 10, Nodejs 10.11.0

After stopping and resuming torrent download progress of files is calculated incorrect.

I am stopping torrent according to documentation (https://github.com/webtorrent/webtorrent/issues/164) by deselecting whole torrent (torrent.deselect(0, torrent.pieces.length - 1, 0)) and then deselecting each file (file.deselect()). When I want to resume torrent download, I am selecting all files (file.select()). and after that i am receiving incorrect progress of files
![image](https://user-images.githubusercontent.com/45289945/48945564-792a8380-ef33-11e8-9509-55310e347e16.png)

How can I solve this problem?

note* I am using electron v3.0.6

KayleePop:
Is this a duplicate of https://github.com/webtorrent/webtorrent/issues/1515?

De-Zol:
> Is this a duplicate of #1515?

You are right, it`s the same problem

PR #1592, opened 2019-03-11
Fixes #1515, #1552.

### LINKED ISSUE #1515: Regression: File.downloaded becomes negative once the first piece is fully downloaded
**What version of WebTorrent?**
0.102.4

**What operating system and Node.js version?**
macOS 10.13.6
Node.js 10.0.0

**What browser and version? (if using WebTorrent in the browser)**

`File.download` property should always be a nonnegative integer.

**What actually happened?**
After the first piece (`_startPiece`) of the file has finished downloaded, the file's `offset` is subtracted from the downloaded size, causing it to be a (sometimes quite large) negative.
This occurs in `file.js`, line 46 (https://github.com/webtorrent/webtorrent/blob/master/lib/file.js#L46), as of revision 6fbef8e.

I noticed the negative values after upgrading from version 0.98.24, which did not have this issue.

corwin-of-amber:
At least, no negative values have been encountered in my app since I applied this patch.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -43,7 +43,7 @@ class File extends EventEmitter {
 
     // Calculate first piece diffrently, it sometimes have a offset
     let downloaded = bitfield.get(start)
-      ? pieceLength - this.offset
+      ? pieceLength - (this.offset % pieceLength)
       : Math.max(piece.length - piece.missing - this.offset, 0)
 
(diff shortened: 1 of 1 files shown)


DiegoRBaquero:
Released as 0.103.1 :)

corwin-of-amber:
Wow, it really payed off to reply to stale-bot 🍭

PR #1598, opened 2019-03-21
Generate relative urls in server.js

last-g:
This change will allow using server behind a reverse proxy.


Fenny:
This is exactly what I need

feross:
Thanks for the PR @last-g. Thanks for reviewing and testing it @alxhotel.

The code in this file changed, so there's now a merge conflict. I just manually made this change and released 0.107.10. ✨

PR #1601, opened 2019-03-25
add vgo desktop app.

vg0vg0:
P2P exhange resources on blockchain.


DiegoRBaquero:
Your app's screenshots show that copyrighted content is being downloaded and we don't support or like that. If you'd like to be included, please remove anything that links to piracy.

Issue #1530, opened 2018-10-24
Webtorrent setting a custom User-Agent in requests


C0nw0nk:
I see webtorrent sends a custom user agent in all its request to my server checking nginx logs.

`WebTorrent/0.102.4 (https://webtorrent.io)`

Isit possible to include a feature to set my own custom user agent in javascript. The benefits of that could be awesome for my server expecting a specific useragent instead in requests and also for `sha1` url's etc.

```
torrent.setUserAgent('WebTorrent/0.102.4 (https://webtorrent.io)') //can change modify and set a custom user agent for requests from webtorrent to your server
```

SilentBot1:
Hey @C0nw0nk,
 
Currently the user agent is hard coded for all [webseeds](https://github.com/webtorrent/webtorrent/blob/master/lib/webconn.js#L117), [tracker](https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L311), and [metadata](https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L368) requests, though it would be easy to implement a changeable user agent. Would you be looking for changing all instances of the user agent including requests made to trackers or only the user agents included with WebSeed / Metadata requests?

All the best.

C0nw0nk:
I only had interest in changing the webseed requests that my servers would
receive.

C0nw0nk:
Ontop of what i previously wrote does webtorrent when requesting the `.torrent` file from my server can that send a custom header or modify the user-agent too ?

jimmywarting:
Borrowing this thread a lite, But i'm wondering how you can get the user-agent of another peer...

SilentBot1:
@C0nw0nk, the implementation I wrote which can be seen [here](https://github.com/SilentBot1/webtorrent/commit/2dd13948acd3f6da0a57025d6ee109eed83a8b3e), I'll not send a PR yet as it seems not all your requirements are there, it sets one blanket user-agent for all requests, meaning it is presented to the tracker, .torrent server (metadata server) and webseed.

If you are wanting to be able to specify all three separately, that could also be done though for your use case I would recommend using a combination of a user-agent and request origin or even a custom middleware with tokens.

Hey @jimmywarting,

[...]

Issue #1507, opened 2018-09-17
Allow transformation of webseed http request


guanzo:
**What version of WebTorrent?**
"webtorrent": "^0.102.4"

**What operating system and Node.js version?**
OS: Windows
Node.js version: v8.9.1

**What browser and version? (if using WebTorrent in the browser)**
Browser: Chrome
Version: 69.0.3497.92 (Official Build) (64-bit)

Feature request: Allow some way for the user to transform the webseed http request.

Tangentially related issue: https://github.com/webtorrent/webtorrent/issues/1193

I'm running into an issue regarding Chrome caching range requests.

Full context here: https://stackoverflow.com/questions/52349984/chrome-cors-requests-are-faster-when-cache-is-disabled

Another explanation of the problem here: https://stackoverflow.com/questions/27513994/chrome-stalls-when-making-multiple-requests-to-same-resource

TL;DR: Chrome reuses the same TCP connection and explicitly waits for each range request to finish before executing the next one because the URL is the same per request, resulting in extremely poor network performance. Enable "Disable cache" to see the performance drastically improve.

One solution that brought the network performance from snail-speed (1s) back to "normal" (250ms) was to add `cache-control: no-cache, no-store` to webconn.js, like so

```
var opts = {
      url: url,
      method: 'GET',
      headers: {
        'user-agent': 'WebTorrent/' + VERSION + ' (https://webtorrent.io)',
        range: 'bytes=' + start + '-' + end,
        'cache-control': 'no-cache, no-store' // Added by me
      }
    }
```
To do this I have to modify code inside node_modules, so it'd be nice if webtorrent provided an API to transform http requests before they're sent.

jimmywarting:
Would be nicer if you could set the cache control on Request instead of sending a extra header to the server

jimmywarting:
Using fetch in browser will lower the dependency. Not so much for us since we already use some of the sub-dependencies. 

node-fetch and browsers fetch are compatible of doing the same things on both sides 

What you don't see in your dependency tree is Nodes core modules that becomes bundled with browserify like `require('stream')` actually becomes `require('readable-stream')` but readable-stream is not included in your package.json so you don't exactly know what you will get. and readable-stream requires buffer but that is also not something you will see in the package.json dependency tree b/c browserify takes care of that.

When requiring "simple-get" i think you will include all of this modules

```js
stream-http: "^3.0.0"
builtin-status-codes: "^3.0.0",
inherits: "^2.0.1",
[...]

PR #1613, opened 2019-04-09
Allows customization of webseed http requests through an options call…

guanzo:
…back function.

This pull request allows users to modify the http options passed to `simple-get`, such as adding headers, query parameters, etc. `torrentOpts` now accepts a `getWebseedOpts` callback function, similar to `getAnnounceOpts`.

Relevant issues: 
https://github.com/webtorrent/webtorrent/issues/1507
https://github.com/webtorrent/webtorrent/issues/1530

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -72,6 +72,7 @@ class Torrent extends EventEmitter {
     this.skipVerify = !!opts.skipVerify
     this._store = opts.store || FSChunkStore
     this._getAnnounceOpts = opts.getAnnounceOpts
+    this._getWebseedOpts = opts.getWebseedOpts
 
     this.strategy = opts.strategy || 'sequential'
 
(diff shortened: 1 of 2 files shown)


gruns:
:+1:

jimmywarting:
Is this something that we would want to have if we changed to using fetch instead of simple-get?
Is there anyone interested in using fetch instead?

guanzo:
> Is this something that we would want to have if we changed to using fetch instead of simple-get?

Sure why not. Regardless of which API handles http requests, I think allowing custom http options is important.

DiegoRBaquero:
Where is the issue where fetch was discussed?

jimmywarting:
Think there is none
But I brought it up in #1507

Fenny:
Any updates?

feross:
The spirit of this PR is good, but I think it would be nicer if this option supported configuring the HTTP request options for all the places where we send HTTP requests, not just Web Seeds. Is there any interest to send a PR that does this?

Issue #1618, opened 2019-04-24
Cannot Handle Invalid Windows Characters


Jaliborc:
**What version of WebTorrent?**

**What operating system and Node.js version?**
Windows 10, Node 10

**What did you expect to happen?**
When you download a torrent that contains files with names with invalid characters (colons for example, "The Banana King: Requiem - Episode 4.mp4"), either windows system invalid characters should be translated automatically like most torrent clients do, or there should be
an way in the API to assign each file a different write name individually.

**What actually happened?**
WebTorrent tries to download the file with the original name. Which fails.

Jaliborc:
Bump

feross:
This seems reasonable to support. Is there a list of invalid windows characters and the preferred characters to translate them into?

Jaliborc:
I made an [npm module](https://github.com/Jaliborc/safe-filename) for our internal use that converts all the invalid characters to the most similar we could find. 

[This file](https://github.com/Jaliborc/safe-filename/blob/master/index.js) lists all the invalid characters and our suggested translations.

ThaUnknown:
fixed in latest versions

PR #1629, opened 2019-05-16
Handle reserved characters in filepaths

Jaliborc:
These two lines of code additions show how you can easily make WebTorrent be able to handle torrents and files having paths using UNIX or POSIX reserved characters, like the majority of torrenting clients do.

It is using a tiny package I created for the purpose of replacing reserved characters with visually similar unicode characters.

See issue #1618

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -26,7 +26,7 @@ const uniq = require('uniq')
 const utMetadata = require('ut_metadata')
 const utPex = require('ut_pex') // browser exclude
 const parseRange = require('parse-numeric-range')
-const filenameReservedRegex = require('filename-reserved-regex')
+const safeFilename = require('safe-filename')
 
 const File = require('./file')
 const Peer = require('./peer')
@@ -448,7 +448,7 @@ class Torrent extends EventEmitter {
     }
 
     this._rarityMap = new RarityMap(this)
-    this.files.each(file => file.path = file.path.replace(filenameReservedRegex(), '_'))
+    this.files.forEach(file => file.path = safeFilename.path(file.path))
 
     this.store = new ImmediateChunkStore(
       new this._store(this.pieceLength, {
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -26,6 +26,7 @@ const uniq = require('uniq')
 const utMetadata = require('ut_metadata')
 const utPex = require('ut_pex') // browser exclude
 const parseRange = require('parse-numeric-range')
+const filenameReservedRegex = require('filename-reserved-regex')
 
 const File = require('./file')
 const Peer = require('./peer')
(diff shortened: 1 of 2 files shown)


jimmywarting:
Hmm, I might think that we shouldn't touch/edit the filename or path like that
browser/node can handle whatever. A unix system should be able to save unix characters as posix should be able to save posix characters. 

If I want to transfer a file from Mac to Mac I would want to reserve the characters in the path/name.

So it should be handled in the place where you actually write the file to the disc?

Jaliborc:
You **need** to do it. Currently, webtorrent tries to write the files with the reserved characters, which fails. For example:
"cache_directory_given/someFolder/This is some file: with a colo.mp4"

This will actually create a 0kb "cache_directory_given/someFolder/This is some file", which webtorrent won't be able to write in. So the file is never downloaded.

Jaliborc:
For reference, BitTorrent replaces every single reserved character with an underscore. This implementation is smarter, as most of the time the change isn't even visible to the eye.

jimmywarting:
I'm not saying we shouldn't do it. I just said it could be done in another way.
Also have to think about what would happen if the torrent had more similar files.

what if a torrent had this 4 files?

- This is some file: with a colon.mp4
- This is some file_ with a colon.mp4
- This is some file with a colon.mp4
- This is some file

( maybe could be worth to have a original and a renamed path/name )

Jaliborc:
I do see your point, but I think you have misunderstood this implementation. While most clients simply replace with an underscore, as you have written in the comment, and to no apparent issue, this one goes one step further: `This is some file: with a colon.mp4` becomes `This is some file꞉ with a colon.mp4`. The difference is not visible, but the colon was replaced with a native canadian character used by the Inuit tribe.

I have a server currently processing over 400.000 torrents and have not found a single one using any of the replacement characters. But yes, in theory it is still possible to have a match, even though ridiculously unlikely. My suggestion for this case would be:

1. Define an escape character
2. Replace the reserved characters as in this pull request
3. Check if replacement creates duplicates
4. If so, append the replaced character with the escape character
5. Repeat from step 3

Jaliborc:
@jimmywarting Found an additional issue, and for this one I don't have a good suggestion. It's very rare, but some torrent file paths go over the maximum length allowed by some operative systems (cough Windows 10 again).  It results in weird undefined behavior, where the file is written but cannot be renamed, copies cannot be created under most circumstances and it cannot be found by many applications/processes (Windows explorer does though).

jimmywarting:
eh, what is the maximum length allowed?

ultimate-tester:
Maximum length on Windows depends on the application. There's the constant MAX_PATH which is set to 260 (drive letter:/<filepath of 256 characters><0 terminator>) characters.

Then, we also have the unicode extended-length path which is 32,767 characters where each individual part (in between slashes) is limited to 255 characters on most implementations.
It's a little messy, agreed...

As for this implementation, I like the creativity but I don't agree with replacing the characters with a hacky character instead, that will throw off everyone and completely pollute the internet with weird file names (personally I like it to be pure). 

If not supported, don't work around but remove instead is my personal belief (or replace with underscore or hyphen). As far as it goes for webtorrent, I don't belief it carries this responsibility at all.

PR #1547, opened 2018-11-22
ability to close and restore streaming server

untitledlt:
Added ability to restore steaming server by calling `.listen(port)` on existing server which was closed before with `.close()`

registered issue: https://github.com/webtorrent/webtorrent/issues/1545

PR #1641, opened 2019-06-17
Return server from server.listen for method chaining to work

tryzniak:
Hi! This should fix https://github.com/webtorrent/webtorrent-cli/issues/101 and https://github.com/webtorrent/webtorrent-cli/issues/100. There was some breaking change in https://github.com/webtorrent/webtorrent/pull/1547. We used to return server for method chaining, to do stuff like `server.listen(...).on('error...')`, and it broke. This PR fixes everything back to normal.


DiegoRBaquero:
I had fixed returning server, but I prefer your version :). Thank you!

feross:
0.107.15

PR #1650, opened 2019-06-24
Add torrent.rescanFiles() to allow manual verify

jhiesey:
Useful if files are modified externally to webtorrent. Needed for Internet Archive use case.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -366,6 +366,14 @@ connections, nor does it pause the streams of existing connections or their wire
 
 Resume connecting to new peers.
 
+## `torrent.rescanFiles([function callback (err) {}])`
+
+Verify the hashes of all pieces in the store and update the bitfield for any new valid
+pieces. Useful if data has been added to the store outside WebTorrent, e.g. if another
+process puts a valid file in the right place. Once the scan is complete,
+`callback(null)` will be called (if provided), unless the torrent was destroyed during
+the scan, in which case `callback` will be called with an error.
+
 ## `torrent.on('infoHash', function () {})`
 
 Emitted when the info hash of the torrent has been determined.
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -505,6 +505,12 @@ class Torrent extends EventEmitter {
       this._markAllVerified()
       this._onStore()
     } else {
+      const onPiecesVerified = (err) => {
+        if (err) return this._destroy(err)
+        this._debug('done verifying')
+        this._onStore()
+      }
+
       this._debug('verifying existing torrent data')
       if (this._fileModtimes && this._store === FSChunkStore) {
         // don't verify if the files haven't been modified since we last checked
@@ -517,11 +523,11 @@ class Torrent extends EventEmitter {
             this._markAllVerified()
             this._onStore()
           } else {
-            this._verifyPieces()
+            this._verifyPieces(onPiecesVerified)
           }
         })
       } else {
-        this._verifyPieces()
+        this._verifyPieces(onPiecesVerified)
       }
     }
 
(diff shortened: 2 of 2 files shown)


lib/torrent.js
```diff
@@ -500,6 +500,12 @@ class Torrent extends EventEmitter {
       this._onWireWithMetadata(wire)
     })
 
+    const onPiecesVerified = (err) => {
```
feross: This function is only used in the else case below. Can you move it into that else block?

lib/torrent.js
```diff
@@ -547,9 +553,10 @@ class Torrent extends EventEmitter {
     })
   }
 
-  _verifyPieces () {
-    parallelLimit(this.pieces.map((_, index) => cb => {
+  _verifyPieces (cb) {
+    parallelLimit(this.pieces.map((piece, index) => cb => {
       if (this.destroyed) return cb(new Error('torrent is destroyed'))
+      if (!piece) return cb(null)
```
feross: Why was this line not necessary before yet is necessary now?

lib/torrent.js
```diff
@@ -568,10 +575,18 @@ class Torrent extends EventEmitter {
...
-      if (err) return this._destroy(err)
-      this._debug('done verifying')
-      this._onStore()
+    }), FILESYSTEM_CONCURRENCY, cb)
+  }
+
+  rescanFiles (cb) {
+    if (!cb) cb = noop
+
+    // already checks for torrent destroyed
```
feross: This still needs a check for torrent destroyed that throws an error synchronously when the torrent is destroyed, i.e. `if (this.destroyed) throw new Error('torrent is destroyed')`. We do this on all public methods. This makes it easier to catch errors where the called doesn't realize the torrent is destroyed.

lib/torrent.js
```diff
@@ -568,10 +575,18 @@ class Torrent extends EventEmitter {
...
+    this._verifyPieces((err) => {
+      if (err) return cb(err)
```
feross: In addition to calling the user's callback, shouldn't this also call `this._destroy(err)`? That is what you do in the other case where `_verifyPieces` returns an error.

lib/torrent.js
jhiesey: Sure

lib/torrent.js
jhiesey: It's not necessary, just a slight optimization to avoid `store.get` returning an error. I'll take it out.

lib/torrent.js
jhiesey: Good to know. Fixed.

lib/torrent.js
jhiesey: Yes it should.

docs/api.md
```diff
@@ -366,6 +366,13 @@ connections, nor does it pause the streams of existing connections or their wire
 
 Resume connecting to new peers.
 
+## `torrent.verifyPieces([function callback (err) {}])`
```
feross: The method is supposed to be called `rescanFiles`, right?

docs/api.md
```diff
@@ -366,6 +366,13 @@ connections, nor does it pause the streams of existing connections or their wire
...
+
+Verify the hashes of all pieces in the store. Useful if the store has been modified
+outside WebTorrent. If a piece hash is incorrect, the torrent will be destroyed, and
```
feross: If a piece hash is incorrect, nothing happens. The underlying `_verifyPieces` method just marks the valid pieces as verified and ignores the invalid pieces. So I think this should say that the callback is called when verification is complete, and an error is returned if the torrent is destroyed during verification for any reason. But, `null` will be returned if verification completes successfully, whether or not all the pieces hash correctly or not.

jhiesey: Yes, you're right. I made this change a while ago and didn't look at it closely enough again.

docs/api.md
jhiesey: D'oh! yes.

feross:
Released as 0.105.0.

PR #1675, opened 2019-07-23
Fix error in Chrome extension environment

feross:
Video streaming is broken in Brave since they upgraded to WebTorrent v0.105.0. The root cause is described in this issue: https://github.com/brave/brave-browser/issues/5358

This is a quick fix. Once we drop support for Node, we can purge `require('url')` from across the codebase and just use the global `URL` everywhere. So, this fix can be removed then without affecting Brave.

Fixes https://github.com/brave/brave-browser/issues/5358


feross:
Released as 0.105.1.

PR #1678, opened 2019-07-30
Fix server 'hostname' option to mitigate DNS rebinding attack

feross:
It appears that this feature, originally added in https://github.com/webtorrent/webtorrent/pull/1260, may have never worked correctly.

When the request hostname does not match the user-provided opts.hostname value, we should stop processing the request and return nothing. Instead, what was happening was that we'd simply omit the Access-Control-Allow-Origin header, which is not sufficient since the whole point of DNS rebinding attacks is that they appear to be same origin and therefore don't require a CORS header.

cc @diracdeltas @yrliou

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -54,13 +54,6 @@ function Server (torrent, opts = {}) {
     // deny them
     if (req.headers.origin == null) return false
 
-    // If a 'hostname' string is specified, deny requests with a 'Host'
-    // header that does not match the origin of the torrent server to prevent
-    // DNS rebinding attacks.
-    if (opts.hostname && req.headers.host !== `${opts.hostname}:${server.address().port}`) {
-      return false
-    }
-
     // The user allowed all origins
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -218,6 +218,11 @@ function Server (torrent, opts = {}) {
       const html = getPageHTML('405 - Method Not Allowed', '<h1>405 - Method Not Allowed</h1>')
       res.end(html)
     }
+
+    function serveEmptyResponse () {
+      res.statusCode = 204
```
feross: Ideally, we'd just kill the whole TCP connection without responding at all but I'm not sure if there's a clean way to do this from the http layer. Looking into it.

diracdeltas:
lgtm

feross:
Nice, figured out how to kill the connection without sending a response. Merging now.

diracdeltas:
lgtm

feross:
0.105.2

PR #1682, opened 2019-08-04
docs: Update the webtorrent JS url

feross:
We were using the old URL pattern, which linked to 0.98.x instead of 0.105.x.


KayleePop:
weird. These exact changes were merged here https://github.com/webtorrent/webtorrent/pull/1716, but github merged this one too without creating any new commits? IDK. Seems like everything is fine though

PR #1691, opened 2019-08-08
App "chromeapp" field to `package.json`

feross:
For: https://github.com/brave/brave-browser/issues/5587

I'm attempting to make a defacto standard for specifying Chrome App dependency substitutions using the `"chromeapp"` field in `package.json`.

The `"chromeapp"` field is just like the [`"browser"` field in `package.json`](https://github.com/defunctzombie/package-browser-field-spec) except it's intended for Chrome Apps instead of a generic browser environment. Bundler tools like `browserify` or `webpack` can be configured to look for the `"chromeapp"` field instead of the `"browser"` field when doing a build for a Chrome App.

In this specific package, since Chrome Apps can use raw sockets we want to replace e.g. `require('net')` with `require('chrome-net')`.

I'm also sending PRs to all dependent projects that need substitutions to work in Chrome Apps.

- https://github.com/jscissr/http-node/pull/2
- https://github.com/webtorrent/torrent-discovery/pull/45
- https://github.com/mafintosh/k-rpc-socket/pull/17
- https://github.com/webtorrent/bittorrent-tracker/pull/328

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,12 @@
     "os": false,
     "ut_pex": false
   },
+  "chromeapp": {
+    "fs-chunk-store": "memory-chunk-store",
+    "http": "http-node",
+    "net": "chrome-net",
+    "os": false
+  },
   "browserify": {
     "transform": [
       "package-json-versionify"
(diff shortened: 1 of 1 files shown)


feross:
This will be part of 0.108.0

PR #1694, opened 2019-08-08
Adds missing callback.

guanzo:
`torrent._verifyPieces()` is missing a callback invocation. Currently it hangs forever. I reached the conditional after calling `torrent.rescanFiles()`, which I believe is buggy and will file an issue for.


feross:
0.107.1

PR #1699, opened 2019-08-09
Allow multiple remove/add calls for a torrent object

ddumont:
torrent object were created each time, even if torrentId was already a torrent.

### DIFF
--- a/index.js
+++ b/index.js
@@ -234,7 +234,7 @@ class WebTorrent extends EventEmitter {
     this._debug('add')
     opts = opts ? Object.assign({}, opts) : {}
 
-    const torrent = new Torrent(torrentId, this, opts)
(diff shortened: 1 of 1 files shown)


ddumont:
ugh...   looks like the state of the torrent after remove is borked... 
I think i need to do this another way

ddumont:
removing the torrent and then adding back the old torrent.torrentFile gets the workflow working, however it loses the data from the previously downloaded torrent...   any ideas?

feross:
You can pass a `path` option to the same location and WebTorrent will reuse the files that already exist there. You can look at `webtorrent-desktop` for an example of how to do this.

ddumont:
@feross any thoughts on a method like pause that actually halts all traffic?  I wouldn't have to remove, re-add then

feross:
Not a bad idea. I don't have bandwidth to write it, but would like to get to this eventually.

PR #1701, opened 2019-08-11
Fix 'left' parameter on seed

### LINKED ISSUE #1105: When you call .seed method, the tracker receives a START event with a "left" parameter greater than zero
Version: webtorrent-hybrid 1.0.5
OS: Archlinux
Node: v.6.10.0

I expect the tracker receives the "left" parameter with 0, because I'm seeding directly a file I have.
 
**What actually happened?**
The tracker got a "left" with the actual size of the file. 

I'm trying to detect when a user is going to download a file, and I think this is a false positive.

alxhotel:
Fixes #1105

As mentioned by @feross, this PR just verifies the torrent before trying to discover peers.

I have tested it by seeding and downloading and it works fine.

If anyone has better solutions, let me know.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -277,6 +277,19 @@ class Torrent extends EventEmitter {
   }
 
   _onListening () {
+    if (this.destroyed) return
+
+    if (this.info) {
+      // if full metadata was included in initial torrent id, use it immediately. Otherwise,
+      // wait for torrent-discovery to find peers and ut_metadata to get the metadata.
+      this._onMetadata(this)
+    } else {
+      if (this.xs) this._getMetadataFromServer()
+      this._startDiscovery()
+    }
+  }
+
+  _startDiscovery () {
     if (this.discovery || this.destroyed) return
 
     let trackerOpts = this.client.tracker
@@ -298,14 +311,6 @@ class Torrent extends EventEmitter {
           return opts
         }
       })
-    }
-
-    if (this.info) {
-      // if full metadata was included in initial torrent id, use it immediately. Otherwise,
-      // wait for torrent-discovery to find peers and ut_metadata to get the metadata.
-      this._onMetadata(this)
-    } else if (this.xs) {
-      this._getMetadataFromServer()
     }
 
     // begin discovering peers via DHT and trackers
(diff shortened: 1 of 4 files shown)


feross:
This doesn't seem to fix the issue for me. Piece verification is asynchronous so it is not guaranteed to complete before the `Discovery` object is initialized and sends messages to the tracker servers.

I used `webtorrent-cli` pointing to a version of `webtorrent` with this PR on it and got this:

```
DEBUG='bittorrent-tracker*' ./bin/cmd.js seed ~/Desktop/blah.png
  bittorrent-tracker:client new client c97eedde957e0689871a450f537c2cf19af36276 +0ms
  bittorrent-tracker:udp-tracker new udp tracker udp://explodie.org:6969 +0ms
  bittorrent-tracker:udp-tracker new udp tracker udp://tracker.coppersurfer.tk:6969 +0ms
  bittorrent-tracker:udp-tracker new udp tracker udp://tracker.empire-js.us:1337 +0ms
  bittorrent-tracker:udp-tracker new udp tracker udp://tracker.leechers-paradise.org:6969 +0ms
  bittorrent-tracker:udp-tracker new udp tracker udp://tracker.opentrackr.org:1337 +0ms
  bittorrent-tracker:client setInterval 900000 +2ms
  bittorrent-tracker:client send `start` { numwant: 50, uploaded: 0, downloaded: 0, left: 119515, event: 'started' } +1ms
```

Note that `left` is not `0`. I think we would need to wait for the verification to actually complete.

alxhotel:
Ah yeah, of course. Thanks for the feedback!

I've made some changes to take the asynchronous behaviour into account. (Still don't know if it's the best approach)

I also had to change the tests `blocklist-dht` and `blocklist-tracker` because the clients were sharing the download path. This meant that `client2` always had the file downloaded, so it wasn't blocking any peers because [it didn't need to](https://github.com/webtorrent/webtorrent/blob/72a2fa9477ab4be51bfbda5ed4dce90766ac57d5/lib/torrent.js#L320). By separating the download paths, now the behaviour is as expected. (It was working previously because the discovery was always trigerring before verifying the pieces)

feross:
> the clients were sharing the download path

Thanks for catching this!

feross:
0.107.4

PR #1704, opened 2019-08-12
destroy videostream (if any) when file is destroyed.

ddumont:
for use with https://github.com/feross/render-media/pull/70 to clean up errors

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -119,7 +119,7 @@ class File extends EventEmitter {
 
   renderTo (elem, opts, cb) {
     if (typeof window === 'undefined') throw new Error('browser-only method')
-    render.render(this, elem, opts, cb)
+    this._stream = render.render(this, elem, opts, cb)
   }
 
   _getMimeType () {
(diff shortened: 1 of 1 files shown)


lib/file.js
```diff
@@ -119,7 +119,7 @@ class File extends EventEmitter {
 
   renderTo (elem, opts, cb) {
     if (typeof window === 'undefined') throw new Error('browser-only method')
-    render.render(this, elem, opts, cb)
+    this._stream = render.render(this, elem, opts, cb)
```
feross: See feedback on your PR to `render-media`, but this won't work if `render` doesn't return a `Videostream` instance and instead uses one of the other strategies.

feross:
I think the spirit of this PR is good, but the current PR against `render-media` has issues that need be resolved first: https://github.com/feross/render-media/pull/70

ddumont:
Since I've moved away from render-media, I don't know if this pull request makes sense.   I don't have an environment any more to test the changes or respond to questions.

PR #1713, opened 2019-08-26
adding add funding.yaml option through Unlock for donations with Ethe…

julien51:
Hello!

First of all, thanks for your (hard) work on WebTorrent @feross. 
It is fantastic to see open source developers contribute immense value through their work and it is also very exciting to see all your work on the "funding OSS" movement. You're leading by example and this is why I am opening this Pull Request!

At [Unlock](https://unlock-protocol.com), we're building a _protocol for memberships_. This means a thousand things, but one example that we think of a lot, is around making "donations" or "sponsorships". A sponsor, donor is a member of a very unique club: the club of people/organizations which help someone do their work "for free" for the rest of the world.

One key aspect of protocols is that they need to be "permission-less" which means that, like Torrent, HTTP or git, they do not require someone else's permission to be used. Memberships should be permission-less too. 

Unlock uses the Ethereum blockchain which means that it is "unstoppable" (even, most importantly by us). When a creator uses Unlock to create their membership(s), no one can stop people from becoming members.

This Pull Requests adds the ability to contribute to your work using a Lock. This Lock is a smart contract deployed at [`0x7567Ce8bBEb821AA39883f4E80f75DA605FdD394`](https://etherscan.io/address/0x7567ce8bbeb821aa39883f4e80f75da605fdd394). For now, it is owned by me, but if you decide to move forward I want to transfer it to you (just send me an Ethereum address!). After that, it is yours only and nobody else can alter or change its behavior.

It is currently set to sell keys using the following characteristics [we mimicked what you had on your _WebTorrent Supporter_ level]:
* they are each "valid" for 30 days
* there can be an unlimited number being sold
* they are priced at 10 DAI each. DAI is a USD _stable_ coin, which means that it is really $10.

One someone purchases a key, they will receive a *unique* token to represent their donation/membership. This token is what people call a Non Fungible Token.

Before going on, I think it would be important to check that your existing sponsor button would not conflict with the `FUNDING.yml` file that this PR adds.

I am more than happy to answer question and listen to feedback. Please feel free to reach out.

Finally, here is what it looks like:

image

PS: I have purchased a key, so there are 10DAI on the lock already ;) and you can see my NFT using any NFT store, [such as OpenSea](https://opensea.io/assets/0x7567ce8bbeb821aa39883f4e80f75da605fdd394/1).

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+custom: https://donate.unlock-protocol.com/?r=webtorrent/webtorrent
--- a/.unlock-protocol.config.json
+++ b/.unlock-protocol.config.json
@@ -1,0 +1,14 @@
+{
+   "persistentCheckout": true,
+   "icon": "https://assets.unlock-protocol.com/tmp/webtorrent.png",
+   "callToAction": {
+     "default": "👋 Hey there, I'm Feross! ✨\n\nI'm an open source author, maintainer, and mad scientist. I maintain 100+ packages on npm which are downloaded 100+ million times per month 🤯It's quite mind-blowing, tbh! All my code is freely accessible on GitHub.\n\nPlease, help me continue my work on WebTorrent by showing your support!",
+     "pending": "Thank you for your support. It means a lot to me!",
+     "confirmed": "Thank you for your support. It means a lot to me!"
+   },
+   "locks": {
+     "0x7567Ce8bBEb821AA39883f4E80f75DA605FdD394": {
+       "name": "WebTorrent Supporters"
+     }
+   }
+ }


feross:
@julien51 Thanks for the PR. I think this sounds really cool, but I'm skeptical that enough folks will be able to figure this out that it's worth it. And I don't really have a problem of folks trying to censor my donations. It's already hard to get anyone to donate using "normal" payment methods. Adding in the extra complexity of blockchain doesn't seem like it will improve the situation here. Who is your product for?

I support the cause of trying out new things. But honestly, it seems like a lot of work. How would I easily convert these funds into dollars or at least something like bitcoin or ether?

julien51:
Hey @feross , thanks for the feedback.
There are a ton of questions here ;)

1. About censoring donation: agreed that this is not that big of deal, BUT, for people who already have a wallet, it is actually much easier than going through their CC. 
[I am of course happy to help you set up a wallet so you can try things by yourself! Just reach out julien@unlock-protocol.com]

2. About converting funds, you can convert DAI to dollars directly on Coinbase for example. Yoiu can also convert to ether using a decentralized exchange (like uniswap) veyr easily too.

Now, the bigger question about "why". The large answer is "programability" (is this a word?). By decoupling  the "donation" from the platform itself, this means that anyone could easily show donation happening here on their 3rd party site. For example you can see I made a donation because [I have a key to this specific lock](https://opensea.io/assets/0x7567ce8bbeb821aa39883f4e80f75da605fdd394/1).

Another idea that we have is that by using a lock it really becomes easy to "pool" resources from multiple repos, even through different users. I am not sure if you're familiar with the idea of DAO, but [we wrote](https://www.reddit.com/r/unlock_protocol/comments/d1w8ly/a_lock_owned_by_a_dao_to_fund_open_source_in/) about how "pooling" donations on a single "foundation-like" organization could look like.

PR #1737, opened 2019-09-06
Emit `metadata` before `ready` and `done`

alxhotel:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

Regarding events, I think it is important the order in which they are emitted.

If I'm correct, the usual order in which `torrent` events are emitted should be:

```
infoHash
metadata
ready
done
```

But right now, the `metadata` event is emitted the last one. This PR fixes the order of this event. And maybe fixes [webtorrent-desktop#1429](https://github.com/webtorrent/webtorrent-desktop/issues/1429). I'm not sure since I was not able to reproduce the bug with the *exact* same steps. But, at least, the order is now (as I believe) correct.

They should focus on staying awesome ;)

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -505,6 +505,9 @@ class Torrent extends EventEmitter {
       this._onWireWithMetadata(wire)
     })
 
+    // Emit 'metadata' before 'ready' and 'done'
+    this.emit('metadata')
+
     if (this.skipVerify) {
       // Skip verifying exisitng data and just assume it's correct
       this._markAllVerified()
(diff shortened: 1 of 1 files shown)


feross:
Released as 0.107.8

feross:
This PR broke `webtorrent-cli` because it destroys the torrent in response to the `'metadata'` event, so I just added an extra check in https://github.com/webtorrent/webtorrent/commit/ef6879a8daca46edc5e2de2febc09d4c87e47988

Released as 0.107.9

PR #1738, opened 2019-09-07
Fix how the first piece's irrelevant bytes are calculated

feross:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Noticed that the fix to use modulo on the offset from https://github.com/webtorrent/webtorrent/pull/1592 wasn't applied to the second case.

Nope


feross:
0.107.12

Issue #1574, opened 2019-01-27
How do you manually add webrtc peer?


jimmywarting:
**What version of WebTorrent?**
0.103.0

I'm trying to connect two torrent clients within the same LAN.
To achieve this I'm using my own communication layer so no tracker, ip:port or DHT should be used.
So I want to use the WebRTC instead of TCP

In the source i found this:
https://github.com/webtorrent/webtorrent/blob/5b7c5eebf41c9e1e1fe8e9e1d4a6ad06e93e7430/lib/torrent.js#L747-L750

I guess that is what i want to use. So to start simple I'm just trying to figure out how to do this using the same browser tab.

So here something along the line at what I'm trying to do:

```js
const opts = { webSeeds: false, tracker: false, dht: false }
const clientA = new WebTorrent(opts)
const clientB = new WebTorrent(opts)
const file = new Blob(['abc'])

clientA.seed(file, torrentA => {
  clientB.add(torrentA.torrentFile, torrentB => {
    torrentB.addPeer(/* webrtc peer */)
  })
})
```

I know i also have to do the signaling myself. I think i need a bit of help with that also

It certainly would have help if there was something like `torrent.createWebRTCPeer(opts)` that took care of other things like adding the peer to the torrent setting a timeout, setting peerID...

jimmywarting:
Figure it out by myself.
Still doe, `torrent.createWebRTCPeer(opts)` would be useful maybe

PR #1739, opened 2019-09-07
Add tests to check the order of torrent events

alxhotel:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[X] Other, please explain: Tests

This PR adds tests for #1737

### DIFF
--- a/test/torrent-events.js
+++ b/test/torrent-events.js
@@ -1,0 +1,83 @@
+const fixtures = require('webtorrent-fixtures')
+const test = require('tape')
+const WebTorrent = require('../')
+const MemoryChunkStore = require('memory-chunk-store')
+
+test('client.add: emit torrent events in order', function (t) {
+  t.plan(6)
+
+  var client1 = new WebTorrent({ dht: false, tracker: false })
+  var client2 = new WebTorrent({ dht: false, tracker: false })
+
+  client1.on('error', function (err) { t.fail(err) })
+  client1.on('warning', function (err) { t.fail(err) })
+
+  client2.on('error', function (err) { t.fail(err) })
+  client2.on('warning', function (err) { t.fail(err) })
+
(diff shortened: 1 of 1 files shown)


alxhotel:
I just realised that the build fails because the tests don't pass in the browser, although I'm not sure why the browser tests weren't executed in the PR checks...

Anyhow, this line does not work in the browser (obviously):
https://github.com/webtorrent/webtorrent/blob/84fb338b9b4bc1dc6d376be34a21828da26e8739/test/torrent-events.js#L29

I guess there are two options:
1. Move the test file to the `node` folder. And only check the order of events in Node.
2. Create a way to add a WebRTC peer to the torrent with: `client.addPeer(webrtcPeer)`. There's an issue (#1574) about how to do this manually. But I'm not sure if the implementation shown there is the proper way to do it.

This brings me to another question: why aren't there any tests for downloading and seeding in the browser? Or am I missing something?

feross:
A lot of the tests are node-only. I think we can solve this by moving the test to the node folder for now. This is still a lot better than not having this test. We can figure out how to solve the webtrc peer issue separately.

We don't run the browser tests in PRs since there is a secret API keys required to interact with Saucelabs and travis doesn't expose those environment variables to PRs since then anyone could send a PR and just log out the secret API key. It's not ideal, but it makes some sense.

alxhotel:
> We don't run the browser tests in PRs since there is a secret API keys required to interact with Saucelabs and travis doesn't expose those environment variables to PRs since then anyone could send a PR and just log out the secret API key. It's not ideal, but it makes some sense.

Ah got it :+1: 

> We can figure out how to solve the webtrc peer issue separately.

Great!

> I think we can solve this by moving the test to the node folder for now

Thanks for fixing this :blush:

PR #1770, opened 2019-10-28
Fix parsed torrent

alxhotel:
Fixes issue [webtorrent-cli#64](https://github.com/webtorrent/webtorrent-cli/issues/64).

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

This PR checks that the `parsedTorrent` has an `infoHash`. If so, this means it is a valid Bittorrent magnet. Thus, other magnets, like the ones used by Gnutella (e.g. `magnet:?xt=urn:sha1:PD...`), are not valid.

Also this PR adds a test to check this behaviour.

[webtorrent-cli#64](https://github.com/webtorrent/webtorrent-cli/issues/64)

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -196,7 +196,7 @@ class Torrent extends EventEmitter {
 
     let parsedTorrent
     try { parsedTorrent = parseTorrent(torrentId) } catch (err) {}
-    if (parsedTorrent) {
+    if (parsedTorrent && parsedTorrent.infoHash) {
       // Attempt to set infoHash property synchronously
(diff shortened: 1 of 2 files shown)


alxhotel:
Good point.

The check that this PR adds is also in the code of `parse-torrent` here: 

https://github.com/webtorrent/parse-torrent/blob/8add26b94698ede08564b16881ea24b4dede4053/index.js#L61

So I guess there is a reason why the function `parseTorrent` does (should?) not throw an error.

alxhotel:
Closing this PR in favor of webtorrent/parse-torrent#80

PR #1786, opened 2019-12-06
Fix infinite loop bug

koush:
The wire request may fail due to choking or wire closed. This results in an infinite loop.

Here is where "false" is expected, otherwise it loops infinitely:
https://github.com/koush/webtorrent/blob/master/lib/torrent.js#L1316

The failure is reported in the callback, but not as a return value:
https://github.com/webtorrent/bittorrent-protocol/blob/master/index.js#L316

Alternatively, the peer choking check can be done in the while loop.

[ ] Documentation update
[ X] Bug fix
[ ] New feature
[ ] Other, please explain:

Watch for a failed wire request (number of requests does not change), and return false.

### DIFF
--- a/lib/rarity-map.js
+++ b/lib/rarity-map.js
@@ -10,21 +10,21 @@ class RarityMap {
     this._pieces = new Array(this._numPieces)
 
     this._onWire = wire => {
-      this.recalculate()
+      this._dirty = true
       this._initWire(wire)
     }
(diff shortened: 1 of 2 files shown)


feross:
Is this still an issue?

DiegoRBaquero:
Closing as it's stale, if you encounter this issue, please re-open

PR #1788, opened 2019-12-10
allow for progress on write

ftreesmilo:
I use this to show progress when dropping large files into my site to create a torrent...
This lets me show the progress of the torrent as the file is moved from File handles into the store.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -113,6 +113,8 @@ class Torrent extends EventEmitter {
     this._servers = []
     this._xsRequests = []
 
+    this._writeCb = opts.writeCb
+
     // TODO: remove this and expose a hook instead
     // optimization: don't recheck every file if it hasn't changed
     this._fileModtimes = opts.fileModtimes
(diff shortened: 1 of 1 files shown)


alxhotel:
Yep, as you've said, this seems like a nice feature for large files.

To merge this PR, I think we need a couple of things:
- Some tests
- Some documentation of the new feature

In addition, I would even consider changing the callback into an event. And be called like `on('write', () => console.log('writing to store'))`.

Also accessing a private attribute (`_blockstream`) is usually not recommended. An alternative would be to add a new "transform", in the `pump` call, between `readable` and `writable`.

ftreesmilo:
will look into it

ftreesmilo:
google's native fs apis have been giving me a headache...    i finally got to something I'm happy with, and I'll come back to this soon and clean things up.  I'd really like to get rid of my diffs

ftreesmilo:
fyi i changed this to happen in my store...   it ended up being much easier, as adding the torrent with files and relying on the internal stream readers to copy them was killing performance.  It was much faster to pipe the streams in the store using the native fs.

ftreesmilo:
please review this as well

Issue #1757, opened 2019-10-10
end-of-stream >=1.4.2 update creates a "premature close" error on seed() when using a W3C File input


Evoks:
**What version of this package are you using?**
0.107.16

**What operating system, Node.js, and npm version?**
Mac OS 10.14.6 / Chrome 77.0.3865.90 or Safari 13 / Node 12.0.0 / npm 6.11.3

**What happened?**
On browser, when I call the `seed` method using a W3C File input, "end-of-stream" is throwing an error :
```
error Error: premature close
    at onclosenexttick (webtorrent.min.js:12485)
    at Item.run (webtorrent.min.js:3160)
    at drainQueue (webtorrent.min.js:3130)
```
It adds the torrent in the WebTorrent client but the `torrent.files === []` and `torrent.done === false`.

If I put the 1.4.1 end-of-stream package, there is no problem.

**What did you expect to happen?**
And the `torrent.files` should contain the input data and `torrent.done` should be `true`.

alxhotel:
Yep. Thanks for notifying this @Evoks.

Looks like `end-of-stream` has added some incompatible changes in the last versions. Related [end-of-stream#11](https://github.com/mafintosh/end-of-stream/issues/11)

umbra-scientia:
Is there any plan for this issue to be solved? Is there a known workaround to make the most minimal upload functionality work at least? All examples and documentation for uploading a file are broken out of the box using the latest release versions. Always "premature close" in the console after a short delay, for almost all file types except PNG files for some odd reason.

lovvtide:
@umbra-scientia Webtorrent just requires ^1.1.0 so what I ended up doing was explicitly requiring end-of-stream@1.4.1 my dependencies, and that solved the problem for me. Hope it helps

belvederef:
> @umbra-scientia Webtorrent just requires ^1.1.0 so what I ended up doing was explicitly requiring end-of-stream@1.4.1 in my dependencies, and that solved the problem for me. Hope it helps

This workaround does not work anymore. Have a look at my solution here https://github.com/webtorrent/instant.io/issues/210#issuecomment-609687048

DeltaF1:
@alxhotel @KayleePop This issue is blocking uploads on instant.io and has crippled usability for the past several months. Is there a way that this can be assigned some sort of higher priority label?

belvederef:
[...]

PR #1804, opened 2020-01-03
downgrade end-of-stream to v1.4.1

KayleePop:
The minor update in our version range was breaking. We can update it later when it's fixed. 

even without an NPM bump, this patch is useful to just let the CI tests work

issue: https://github.com/webtorrent/webtorrent/issues/1757 (should maybe keep that open until we can re-enable updates for the package though)


DeltaF1:
It appears that the version is still not properly downgraded, see https://github.com/webtorrent/instant.io/issues/210#issuecomment-609687048

PR #1818, opened 2020-03-14
use native Set instead of uniq library

KayleePop:


jimmywarting:
Another way of doing it, instead of `Array.from`
```js
[...new Set()]
```

heck why not even use Set instead of an array? Would it be best if it could stay uniq all the time? guess it's more problematic to just change something from Array to a Set

(Just Code Golf - either way LGTM)

KayleePop:
That'd be a breaking change because torrent.announce is exposed and documented, but seems like a good idea to me

PR #1819, opened 2020-03-14
Fix and document "private" option

### LINKED ISSUE #1816: "private" option when seeding a torrent?
according to this https://github.com/webtorrent/webtorrent/issues/386, there is a "private" option for the seeded torrent; however, it is not documented. I can clearly see it in the source code also, but Webtorrent API documentation does not mention it.

alxhotel:

[X] Documentation update
[X] Bug fix
[ ] New feature
[X] Other, please explain:
- Add tests

- Add the `private` option to the docs of the API.
- Add tests for the `private` option
- Add missing functionality of the `private` option

Fixes #1816 

Let me know if the changes made in `lib/torrent.js` can be improved :)

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -69,10 +69,11 @@ class Torrent extends EventEmitter {
     this.urlList = opts.urlList
 
     this.path = opts.path
-    this.private = !!opts.private
     this.skipVerify = !!opts.skipVerify
     this._store = opts.store || FSChunkStore
     this._getAnnounceOpts = opts.getAnnounceOpts
+
+    if (typeof opts.private === 'boolean') this.private = opts.private
 
     this.strategy = opts.strategy || 'sequential'
 
@@ -252,6 +253,14 @@ class Torrent extends EventEmitter {
 
   _processParsedTorrent (parsedTorrent) {
     this._debugId = parsedTorrent.infoHash.toString('hex').substring(0, 7)
+
+    if (typeof this.private !== 'undefined') {
+      // `private` option overrides default, only if it's defined
+      parsedTorrent.private = this.private
+    } else {
+      // use default value
+      this.private = !!parsedTorrent.private
+    }
 
     if (this.announce) {
       // Allow specifying trackers via `opts` parameter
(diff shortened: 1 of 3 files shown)


lib/torrent.js
```diff
@@ -252,6 +254,14 @@ class Torrent extends EventEmitter {
   _processParsedTorrent (parsedTorrent) {
     this._debugId = parsedTorrent.infoHash.toString('hex').substring(0, 7)
 
+    if (typeof this.private !== 'undefined') {
+      // `private` option overrides default, only if it's defined
+      parsedTorrent.private = this.private
+    } else {
+      // use default value
+      this.private = !!parsedTorrent.private
```
KayleePop: The Object.assign below will assign `this.private` to `parsedTorrent.private`, so the else block isn't necessary? Or maybe there's a bug here?

alxhotel: Good one! Well, if the parsed torrent is public then `parsedTorrent.private` would be `undefined`. Thus, the variable `this.private` would not be updated in the Object.assign. This line basically forces this update.

lib/torrent.js
```diff
@@ -73,6 +73,8 @@ class Torrent extends EventEmitter {
     this._store = opts.store || FSChunkStore
     this._getAnnounceOpts = opts.getAnnounceOpts
 
+    if (typeof opts.private === 'boolean') this.private = opts.private
```
KayleePop: could use a comment saying that `this.private === undefined` when it wasn't passed as an option and otherwise it will overwrite the private field in the torrent file

Otherwise it just looks like you're making sure the input is a boolean instead of relying on it being undefined later

lib/torrent.js
KayleePop: It would be falsy either way, but I guess having it defined is good? seems fine

lib/torrent.js
alxhotel: Agreed

lib/torrent.js
alxhotel: Ok, done. I think your change even improves readability. Thanks :)

Issue #1671, opened 2019-07-20
Webtorrent v0.105.0 won't download from webseed if no peer online


ziro-official:
**What version of WebTorrent?**
   v0.105.0 and v0.99.0
**What operating system and Node.js version?**
  Ubuntu 18.04
**What browser and version? (if using WebTorrent in the browser)**
  No
**What did you expect to happen?**
 v0.105.0 won't start download from webseed but v0.99.0 did.
**What actually happened?**
  When I create a client from a .torrent file, Webtorrent won't start download from webseed if no peer online (fresh torrent, no one seed it, no metadata on any tracker). I tried to uninstall v0.105.0 and install v0.99.0 and it work, it start download from webseed right away.

SilentBot1:
Hi @ziro-official 

Please could you provide a code example of how you are attempting to achieve this in v0.105.0?

Kind regards,
Brad

feross:
We need a code example to attempt to reproduce this. Thanks!

jimmywarting:
Think i have stumble upon this to before.
think it's when you only have a magnet link and not the full torrent metadata?
the magnet link may contain one `ws` (webseed) without a `xs` (exact source to torrent file)

if there where nobody downloading this and xs where not attached it would not download anything.
```
magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10
&ws=https://webtorrent.io/torrents/Sintel.torrent # folder to all the files
&xs=https://webtorrent.io/torrents/sintel.torrent # actual torrent file
```

jimmywarting:
While i was looking at the sintel.torrent source i could not find any information regarding a `xs` in the torrent file?
That would result in a magnet link not having any `xs`

SilentBot1:
@jimmywarting that's what I was thinking.

I just wanted to double check with @ziro-official how they were adding the torrent, as if they were adding it via the MagnetURI without `&xs` or peers, this would be causing the torrent to not load as no metadata is available.

The `sintel.torrent` downloaded from [WebTorrent.io](https://webtorrent.io) doesn't include a `xs` location as the full torrent buffer is embedded into the JavaScript itself, meaning if a MagnetURI was generated from this `.torrent`, it would contain no `xs` location.

If @ziro-official is adding the torrent with the `.torrent` file, this would be an unrelated issue than what we're discussing, which is why I asked for more clarification.

PaTiToMaSteR:
[...]

PR #1823, opened 2020-03-29
Mark webseeds interested at the start, ensuring they download anything

ianmcorvidae:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

I've made it so there is a call to `wire.interested()` for webseed peers since it was discovered on the issue linked below that the `wire.uninterested()` further up in the file was preventing webseeds from working correctly.

I tested this manually with webseeds using a similar bit of code to the one in the issue below.

#1671 
possibly others

I'd like to be sure of the correctness of this, particularly if I might have affected anything other than webseeds.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1094,6 +1094,8 @@ class Torrent extends EventEmitter {
     if (wire.type !== 'webSeed') { // do not choke on webseeds
       timeoutId = setTimeout(onChokeTimeout, CHOKE_TIMEOUT)
       if (timeoutId.unref) timeoutId.unref()
+    } else { // mark webseeds interested
+      wire.interested()
     }
 
     wire.isSeeder = false


DiegoRBaquero:
Could you come up with a test scenario so that this fixes and makes that test passing?

TemaSM:
Tested, webseeds now works just fine. Thanks for fix @ianmcorvidae 👍

hicom150:
I've been reviewing this PR and I found out that this issue seems already fixed in https://github.com/webtorrent/webtorrent/commit/a07188f87a4acb7f8b2338c3d41da1b57cb9878c and released in [v0.108.4](https://github.com/webtorrent/webtorrent/releases/tag/v0.108.4) 😅

It would be nice if @ianmcorvidae you could recheck if the issue has been solved, so we can decide if we need to merge or close this PR ;)

ianmcorvidae:
@hicom150 I haven't looked at this in ages (obviously, I guess!) but that looks like it _should_ fix this and is a better solution than mine. To be honest, I don't even remember what I was using before to test this, so if others are having webseeds work correctly, I'm content with the released code.

PR #1832, opened 2020-04-15
use for-await in torrent.load

KayleePop:
this allows the removal of `chunk-store-stream` and `multistream` dependencies

pump will also no longer be included in the browser bundle because it's only required in `server.js` after this PR. (actually, it's probably still in the bundle from a dependency :shrug:)

I think this also means sidestepping the premature close issue with multistream? https://github.com/webtorrent/webtorrent/issues/1748

Compatibility shouldn't be an issue https://caniuse.com/#search=for%20await (and node >=10)


Arman92:
@alxhotel This PR helps a lot of people! Please review it

PR #1762, opened 2019-10-20
Add global down/up speed limit

JohnRey9:
**What is the purpose of this pull request? (put an "X" next to item)**

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

**What changes did you make? (Give an overview)**
Add functional for control download and upload speed

**Is there anything you'd like reviewers to focus on?**
This is update for #1040

PR #1852, opened 2020-05-05
Feature/163 global download upload limit Renewed

ultimate-tester:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

This provides a way to throttle download/upload speed. I've implemented the new code fully provided by @chtrinh and adjusted it to fit with the latest commit of the master branch (also involves converting the code to the current version of ECMAScript that webtorrent ). Dependencies were also updated (merged with greenkeeper and manually corrected where needed).


ultimate-tester:
UPDATE: Corrected formatting according to lint, replaced self with this.

ultimate-tester:
So we have 1 approval and 1 remaining with write access. How long does it normally take for pull requests to go through?

alxhotel:
This PR looks good, but there are some things missing:

- The `tcpIncoming` connections are not being throttled. Check out #1762 where they used `clearPipes` and `setThrottlePipes` for this case.
- We need tests for this feature. I suggest checking the `uploadSpeed` and `downloadSpeed` of the client.
- Also, important to check how the `uploadSpeed` of the client is being calculated. Because right now it returning "fake" speeds, since the speed is being calculated before the throttle of the upload. There must be nice way to calculate the speed after being throttled (thus returning the real speed).

ultimate-tester:
> 
> 
> This PR looks good, but there are some things missing:
> 
>     * The `tcpIncoming` connections are not being throttled. Check out #1762 where they used `clearPipes` and `setThrottlePipes` for this case.
> 
>     * We need tests for this feature. I suggest checking the `uploadSpeed` and `downloadSpeed` of the client.
> 
>     * Also, important to check how the `uploadSpeed` of the client is being calculated. Because right now it's returning "fake" speeds, since the speed is being calculated before being throttled. There must be nice way to calculate the speed after being throttled (thus returning the real speed).

Thanks for the valid feedback

1) Fixed, seems like I wasn't so skilled in copy+pasting after all
2) I created a new test, but for some reason every test file doesn't download for me at all. Can anyone reading this spot the error and give a hand?
3) I will do that right after I get the test working

feross:
Is this PR still relevant?

hicom150:
I've added #1969 to try to solve the issues that @alxhotel correctly commented above, all based in https://github.com/webtorrent/webtorrent/pull/1040 and https://github.com/webtorrent/webtorrent/pull/1852

> The tcpIncoming connections are not being throttled.

To solve this problem, instead of reaching `ThrottleGroup` via internal swarm->client->ThrottleGroup, 
I've passed the `ThrottleGroup` dependency through the named constructors in Peer (`createWebRTCPeer`,
`createTCPIncomingPeer`, `createUTPIncomingPeer`, `createTCPOutgoingPeer`, `createUTPOutgoingPeer`)

> We need tests for this feature.

Added an unit test `limit-download-upload.js`

> Check how the uploadSpeed of the client is being calculated since the speed is being calculated before being throttled.

To solve this issue I've changed the approach of where the speed is measured, and instead of doing it inside `Wire`
I measured directly in peer connection pipe. This way we can measure the "real" speed of the data that is being transferred through the pipe.

Let me know if my PR can help fixing those issues, so we can push forward **speed limits** 😉

alxhotel:
Closed by #2062

Issue #1864, opened 2020-05-18
WebTorrent not interested in downloading files


perf2711:
**What version of this package are you using?**
`webtorrent@0.108.2`

**What operating system, Node.js, and npm version?**
Windows 10 1901 18363.836
Node.js 12.16.2
npm 6.14.4
yarn 1.21.1 (I am using yarn to install packages)

**What happened?**
I set up a minimal program to download a file from .torrent I created. The torrent has no trackers, but I plan to manually add peers using `.addPeer()`. When I add a peer, I see on the other side, that the peer has been added (I am using qBitTorrent). However, on their side, webtorrent is visible as "not interested" and "choking". I am 100% sure that the torrent file is the same. 

Sometimes, the file is downloaded (but, like 1/50 of the time).

**What did you expect to happen?**
I expect the file to download every time.

**Are you willing to submit a pull request to fix this bug?**
If I know what it is, sure.

**Example source**
```typescript
import fs from 'fs';
import WebTorrent from 'webtorrent';
import { throttle } from 'lodash';

function onDownload(torrent: WebTorrent.Torrent, bytes: number) {
    const progress = torrent.progress * 100;
    console.trace(`[${torrent.infoHash}] D ${progress}% +${bytes}B (${torrent.downloaded}B/${torrent.length}B)`);
}

function onUpload(torrent: WebTorrent.Torrent, bytes: number) {
    console.trace(`[${torrent.infoHash}] U +${bytes}B`);
}

async function main() {
    const client = new WebTorrent({
        dht: false,
        tracker: false
    });

    client.add('./sample.torrent', {
        path: './download'
    }, (torrent) => {
        torrent.on('download', throttle((bytes) => {
            onDownload(torrent, bytes);
        }, 100));

        torrent.on('upload', throttle((bytes) => {
            onUpload(torrent, bytes);
        }, 100));

        // qBitTorrent's address (this is localhost)
        torrent.addPeer('192.168.1.66:48599');
    });
}

(async () => {
    main();
})();
```

**Output (with DEBUG=*)**
```
 webtorrent [2d57573] new webtorrent (peerId 2d5757303030382d2f7534474c4172786b634875, nodeId 452b7a3e75270ac38a8a000aa9d8ed015797b55f, port 0) +0ms
  webtorrent:tcp-pool create tcp pool (port 0) +0ms
  webtorrent [2d57573] add +11ms
  webtorrent:torrent [2d57573] [unknown infohash] new torrent +0ms
[...]

PR #1870, opened 2020-05-25
Fix setting initial interest

KayleePop:
fixes https://github.com/webtorrent/webtorrent/issues/1864 by actually checking whether we're interested instead of always setting uninterested initially

It looks like in https://github.com/webtorrent/webtorrent/pull/1061 they misread the spec and thought that "Connections start out choked and not interested" meant that we had to send an uninterested message on start (but it really means lack of any message means not interested, at least I think)

@paullouisageneau could you see if this fixes your issue in libtorrent as well?

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1084,7 +1084,7 @@ class Torrent extends EventEmitter {
     })
 
     wire.bitfield(this.bitfield) // always send bitfield (required)
-    wire.uninterested() // always start out uninterested (as per protocol)
+    this._updateWireInterest(wire) // set inital interest
 
     // Send PORT message to peers that support DHT
     if (wire.peerExtensions.dht && this.client.dht && this.client.dht.listening) {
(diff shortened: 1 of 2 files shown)


lib/torrent.js
```diff
@@ -1084,7 +1084,7 @@ class Torrent extends EventEmitter {
     })
 
     wire.bitfield(this.bitfield) // always send bitfield (required)
-    wire.uninterested() // always start out uninterested (as per protocol)
+    this._updateWireInterest(wire) // set inital interest
```
alanmcgovern, line 1087: If you support the FAST extensions then you should also update the interested state when a `HaveAll` message is received.

KayleePop: I don't think webtorrent supports BEP6? https://github.com/webtorrent/bittorrent-protocol (sounds like a good thing to make an issue about though)

`_updateWireInterest()` is also called anytime a selection is made (piece priority changed), and also when a piece is successfully validated or added to the store from `_checkdone()` (because it calls `_gcSelections()`, which calls `_updateSelections()`)

alanmcgovern:
`it really means lack of any message means not interested, at least I think`

Exactly right! You don't need to *send* a NotInterested message at startup as that is the default state. When you receive another peers 'Bitfield' message you can immediately send an 'Interested' message [0]. if they have pieces you need.

[0] The sequence, as per spec, is:
Send the bittorrent handshake message.
Send the extended handshake message if you support the extension protocol.
Send a Bitfield message (or a HaveAll/HaveNone if you support the extension protocol)
Send anything else.

As such please be sure you do not send an `Interested` message before you send your own `Bitfield` message. You probably already do this, but it's worth calling out just in case!

KayleePop:
> As such please be sure you do not send an Interested message before you send your own Bitfield message. You probably already do this, but it's worth calling out just in case!

Looks like this is the case https://github.com/webtorrent/webtorrent/pull/1870/files#diff-65b0a569f99e647ae33e8c31b9bf72baL1086

paullouisageneau:
@KayleePop I've just tested and it indeed solves the issue in libtorrent, thanks!

perf2711:
Is there a change these fixes will be released as official npm version in near time?

Issue #1871, opened 2020-05-25
WebTorrent should not send unchoke messages to seeders


alanmcgovern:
I was playing around with WebTorrent last week to verify a reported bug ( https://github.com/alanmcgovern/monotorrent/issues/300#issuecomment-630380468 ). It looks like the root of that has been resolved ( https://github.com/webtorrent/webtorrent/pull/1870 ), which is great.

I did notice a second bug at the time, but didn't file it. It looks like WebTorrent sends 'Unchoke' messages to peers which have not sent an 'Interested' message. I noticed as the logs for my client indicated it received an `Unchoke` message even though it was a seeder. This should never happen as seeders do not send 'Interested' messages.

Can you double check that you always verify a peer is interested before unchoking it?

KayleePop:
The check happens here https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L1351

isSeeder is set here https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L1044 (on bitfield and have messages as well as initially)

alanmcgovern:
You pointed me at a check for 'isSeeder', but there should be a check for 'isInterested'. Only interested peers should be unchoked.

KayleePop:
hmm. maybe I'm missing something (the code is pretty hard to read), but yeah it looks like uninterested peers are only filtered out for optimistic unchoking and not for bandwidth based unchoking

https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L1373

I'll make a PR, thanks

alanmcgovern:
Sounds great.

I'm not 100% sure which piece of code that refers to, but the piece I was most curious about was this: https://github.com/webtorrent/webtorrent/blob/c9462bd8a789b6599b9ff1cb00b6280597dee519/lib/torrent.js#L1384-L1389 . My reading of this - which could be wrong - is that if a seeder unchokes WebTorrent then WebTorrent will unchoke that seeder, even though the seeder is not interested.

alxhotel:
I think the current implementation is correct based on the specification (https://www.bittorrent.org/beps/bep_0003.html):

> Peers which have a better upload rate but aren't interested get unchoked and if they become interested the worst uploader gets choked

I think this is why the current implementation sends an `unchoke` message to peers not interested. 

[...]

PR #1872, opened 2020-05-26
refactor _rechoke()

KayleePop:
~~https://github.com/webtorrent/webtorrent/issues/1871 fixed by checking if each peer is interested when unchoking. Before, uninterested peers would be unchoked as long as they weren't seeders~~

refactor torrent._rechoke() for readability. Behavior should be the same (can change to only unchoke interested peers later if we want)

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1348,71 +1348,79 @@ class Torrent extends EventEmitter {
   _rechoke () {
     if (!this.ready) return
 
-    if (this._rechokeOptimisticTime > 0) this._rechokeOptimisticTime -= 1
-    else this._rechokeOptimisticWire = null
+    // wires in increasing order of quality (pop() gives next best peer)
+    const wireStack =
+      this.wires
+        .map(wire => ({ wire, random: Math.random() })) // insert a random seed for randomizing the sort
+        .sort((objA, objB) => {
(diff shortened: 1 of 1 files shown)


lib/torrent.js
```diff
@@ -1348,71 +1348,78 @@ class Torrent extends EventEmitter {
...
+      this._rechokeOptimisticTime -= 1
     }
 
-    // Optimistically unchoke a peer
-    if (!this._rechokeOptimisticWire && i < peers.length && this._rechokeNumSlots) {
-      const candidates = peers.slice(i).filter(peer => peer.wire.peerInterested)
-      const optimistic = candidates[randomInt(candidates.length)]
+    let numInterestedUnchoked = 0
+    while (numInterestedUnchoked < this._rechokeNumSlots - 1) { // leave one slot open for optimistic unchoking
+      const wire = wireStack.pop() // next best quality peer
```
alxhotel, line 1388: I would check if `wireStack` has any elements left. Or do something like this:

```suggestion
      const wire = wireStack.pop() // next best quality peer
      if (!wire) break
```

KayleePop: nice catch

KayleePop:
thanks for the review!

PR #1900, opened 2020-08-12
Watch progress of verification checks

ftreesmilo:
**What is the purpose of this pull request? (put an "X" next to item)**

[ ] Documentation update
[ ] Bug fix
[x ] New feature
[ ] Other, please explain:

**What changes did you make? (Give an overview)**
emit 2 new events on torrent so that you can see when a piece is verified or invalidated when verifying

**Which issue (if any) does this pull request address?**
n/a

**Is there anything you'd like reviewers to focus on?**
n/a

PR #1903, opened 2020-08-14
feat: rescan of torrent should get all (in)valid events

ftreesmilo:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

When adding an existing, completed torrent file with `skipVerify` to speed client start time, calling rescanFiles would kick off a scan but only report invalid chunks, because this line short-circuited the call on verification success.

n/a

can you document the importance of the this.pieces array somehow?   a few comments here and there would help me to understand what it means for them to be there or not.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -144,6 +144,10 @@ class File extends EventEmitter {
     render.render(this, elem, opts, cb)
   }
 
+  includes (piece) {
+    return this._startPiece <= piece && this._endPiece >= piece
+  }
+
   _getMimeType () {
     return render.mime[path.extname(this.name).toLowerCase()]
   }
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -620,12 +620,7 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
-            this.pieces[index] = new Piece(piece.length || this.pieceLength)
-            this.bitfield.set(index, false)
-            this.select(index, index, 1)
-            this.files.forEach(file => {
-              if (file.done && file.includes(index)) file.done = false
-            })
+            this._markUnVerified(index)
             this._debug('piece invalid %s', index)
           }
           cb(null)
@@ -659,6 +654,18 @@ class Torrent extends EventEmitter {
     this.pieces[index] = null
     this._reservations[index] = null
     this.bitfield.set(index, true)
+  }
+
+  _markUnVerified (index) {
+    const len = (index === this.pieces.length - 1)
+      ? this.lastPieceLength
+      : this.pieceLength
+    this.pieces[index] = new Piece(len)
+    this.bitfield.set(index, false)
+    this.select(index, index, 1)
+    this.files.forEach(file => {
+      if (file.done && file.includes(index)) file.done = false
+    })
   }
 
   /**
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -103,6 +103,8 @@ class Torrent extends EventEmitter {
     this.metadata = null
     this.store = null
     this.files = []
+
+    // Pieces that need to be downloaded, indexed by piece index
     this.pieces = []
 
     this._amInterested = false
@@ -599,6 +601,10 @@ class Torrent extends EventEmitter {
   }
 
   _verifyPieces (cb) {
+    const done = () => {
+      this._checkDone()
+      cb()
+    }
     parallelLimit(this.pieces.map((piece, index) => cb => {
       if (this.destroyed) return cb(new Error('torrent is destroyed'))
 
@@ -618,12 +624,18 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
+            this.pieces[index] = new Piece(piece.length || this.pieceLength)
+            this.bitfield.set(index, false)
+            this.select(index, index, 1)
+            this.files.forEach(file => {
+              if (file.done && file.includes(index)) file.done = false
+            })
             this._debug('piece invalid %s', index)
           }
           cb(null)
         })
       })
-    }), FILESYSTEM_CONCURRENCY, cb)
+    }), FILESYSTEM_CONCURRENCY, done)
   }
 
   rescanFiles (cb) {
@@ -636,7 +648,6 @@ class Torrent extends EventEmitter {
         return cb(err)
       }
 
-      this._checkDone()
       cb(null)
     })
   }
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -620,7 +620,7 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
-            this._markUnVerified(index)
+            this._markUnverified(index)
             this._debug('piece invalid %s', index)
           }
           cb(null)
@@ -656,7 +656,7 @@ class Torrent extends EventEmitter {
     this.bitfield.set(index, true)
   }
 
-  _markUnVerified (index) {
+  _markUnverified (index) {
     const len = (index === this.pieces.length - 1)
       ? this.lastPieceLength
       : this.pieceLength
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -601,10 +601,6 @@ class Torrent extends EventEmitter {
   }
 
   _verifyPieces (cb) {
-    const done = () => {
-      this._checkDone()
-      cb()
-    }
     parallelLimit(this.pieces.map((piece, index) => cb => {
       if (this.destroyed) return cb(new Error('torrent is destroyed'))
 
@@ -635,7 +631,7 @@ class Torrent extends EventEmitter {
           cb(null)
         })
       })
-    }), FILESYSTEM_CONCURRENCY, done)
+    }), FILESYSTEM_CONCURRENCY, cb)
   }
 
   rescanFiles (cb) {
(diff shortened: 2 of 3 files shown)


alxhotel:
Hi @ftreesmilo,

Thanks for the PR!

The line that you point out doesn't seem to have any influence in the emission of events, since it's already calling `cb(null)` anyway.

If you are referring to the debug messages, then you're right.

Nonetheless, since `_markVerified` seems idempotent. I don't see why we can't remove that line (585); except for optimization's sake.

By the way, since you asked: `this._pieces` contains the array of the current pieces that needs to be downloaded. Once a piece is downloaded, verified and stored, the element is deleted from the array. You can add a comment next to the declaration of the array, if you like.

ftreesmilo:
@alxhotel it return's though, so it skips the emit of the verified event.

ftreesmilo:
So unfortunately... it's hard by looking at events form the outside to see what's going on.
If I add a torrent and skip the initial verification, then markAllVerified gets called and there are no pieces in the array.

when I call rescanFiles, it will check the hashes in the torrent, but won't emit the verified event if there are no pieces in the array.

alxhotel:
If it reaches the line 524: https://github.com/webtorrent/webtorrent/blob/7f95750fc175fc88c50db630cf16dbcb3a5ae3cb/lib/torrent.js#L524

It means that the metadata of the torrent was received. So `this.pieces` should not be empty.

How are you creating the torrent?

ftreesmilo:
@alxhotel  `client.add(torrentId, { skipVerify: true })`

ftreesmilo:
I'm trying to balance performance...    If i have 10 torrents in my client when I start it, it's going to thrash my hd all the time.

Where-as if I know i've previously finished it... I don't need to scan it unless the user wants to check the local files. in case there are errors.

alxhotel:
> @alxhotel `client.add(torrentId, { skipVerify: true })`

After adding the torrent, you must check that the torrent is `ready` before doing a rescan. This is to ensure you have the `this.pieces` array. For example:

```js
const torrent = client.add(torrentId, { skipVerify: true })

torrent.on('ready', () => {
  torrent.rescanFiles((err) => {
    if (err) return console.log('Error: ', err)
    console.log('Rescan done successfully')
  })
})

```

ftreesmilo:
I'm doing the rescan... the problem is I don't get the verified events.. only the invalid ones...   

I did work around something in my client... about to update this pr with the changes to do what I need.

ftreesmilo:
ermm... i've seen others mention they want to migrate to promises... would you mind if I added a dep on bluebird?

alxhotel:
> I'm doing the rescan... the problem is I don't get the verified events.. only the invalid ones...

Sorry, what verified event are you referring to?

alxhotel:
> ermm... i've seen others mention they want to migrate to promises... would you mind if I added a dep on bluebird?

I would rather stick with callbacks for the moment. And then migrate the whole library to promises in a future PR.

ftreesmilo:
@alxhotel the event was the torrent.on('verified'...   it's not fired without deleting the line this PR deletes.

alxhotel:
There's no `verified` event in torrent. I think you might be using a different version.

ftreesmilo:
@alxhotel #1900 🙂

ftreesmilo:
@alxhotel this seems to fix things with progress for finding and fixing invalid pieces

ftreesmilo:
@alxhotel sorry for the flurry of commits.   I use patch-package to maintain my own patches to webtorrent and it gets confusing bringing them here.

Also the code style rules for the project are satanic 😉

ftreesmilo:
@alxhotel can you please re-review?

ftreesmilo:
@alxhotel @feross Was fried at the meeting.   I use this to monitor progress of checking the torrent's integrity.

lib/torrent.js
```diff
@@ -600,12 +606,18 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
+            this.pieces[index] = new Piece(length || this.pieceLength)
+            this.bitfield.set(index, false)
+            this.select(index, index, 1)
+            this.files.forEach(file => {
+              if (file.done && file.includes(index)) file.done = false
+            })
             this._debug('piece invalid %s', index)
```
ftreesmilo, line 615: @jhiesey this part is useful to mark the current torrent that the piece found in the store is not correct and should be re-requested.    Helps keep the torrent consistent

ftreesmilo: @alxhotel  too :)

lib/file.js
```diff
@@ -134,6 +134,10 @@ class File extends EventEmitter {
     render.render(this, elem, opts, cb)
   }
 
+  includes (piece) {
```
alxhotel, line 137: This looks useful! Can you please add this method in the docs?

lib/torrent.js
```diff
@@ -600,12 +606,18 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
+            this.pieces[index] = new Piece(length || this.pieceLength)
+            this.bitfield.set(index, false)
+            this.select(index, index, 1)
+            this.files.forEach(file => {
+              if (file.done && file.includes(index)) file.done = false
+            })
```
alxhotel, line 614: Maybe I'm missing something, but I don't see why this is necessary.

By default, if the pieces is not valid:
- The `this.pieces[index]` element is not emptied 
- The `this.bitfield` is already `false`
- The reselection of the piece is not needed because the `this.bitfield` was not changed
- The `file.done` is not changed because we still don't have the piece

Is this right @ftreesmilo ?

lib/torrent.js
```diff
@@ -582,6 +584,10 @@ class Torrent extends EventEmitter {
   }
 
   _verifyPieces (cb) {
+    const done = () => {
+      this._checkDone()
```
alxhotel, line 588: Why do we need to call it here?

I think we are already calling it in the callback when `_verifyPieces` is called.

alxhotel (changes requested):

lib/file.js
ftreesmilo: will do

lib/torrent.js
ftreesmilo: In a rescan, a fully downloaded file with a corruption issue will think it has the chunk and later find out it doesn't.

lib/torrent.js
ftreesmilo: It does not do it through the codepath of onMetadata where skip verify is false

ftreesmilo: would you rather me add it there or remove it from rescan cb?

lib/torrent.js
```diff
@@ -618,12 +624,18 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
+            this.pieces[index] = new Piece(piece.length || this.pieceLength)
+            this.bitfield.set(index, false)
+            this.select(index, index, 1)
+            this.files.forEach(file => {
+              if (file.done && file.includes(index)) file.done = false
+            })
```
alxhotel, line 632: Ok, I agree.

Then, I think we should create a separate private method (called something like `markUnverified`) with the lines for `pieces` and `bitfield`.

And then we should change the code from `_checkDone` to mark the torrent as not done and mark the files as not done.

I am not sure if this is breaking change or a fix, since the torrent will change state without the user knowing. @feross @jhiesey ?

alxhotel (changes requested):

lib/torrent.js
ftreesmilo: ok I'll put it back... I just remember I added it for a reason, but the reason escapes me now.    I'll retest with it put back and removed from this method

ftreesmilo: ok I'll put it back... I just remember I added it for a reason, but the reason escapes me now. I'll retest with it put back and removed from this method

ftreesmilo: do you know if select is needed? i wasn't sure

ftreesmilo: I don't think I'm a fan of marking the files as not done in _checkdone.
I had a pr (not sure if it's merged) where I wanted to see events as the pieces are verified or not...
The files and pieces will be out of sync if we wait till we are done with the file.  And I'm not sure what will happen when the client sees that it's missing pieces, but the files are all marked done.

Each verification is an async operation.... there's room for other code to run in between.

lib/torrent.js
```diff
@@ -653,6 +656,18 @@ class Torrent extends EventEmitter {
     this.bitfield.set(index, true)
   }
 
+  _markUnVerified (index) {
```
DiegoRBaquero, line 659:
```suggestion
  _markUnverified (index) {
```

lib/torrent.js
```diff
@@ -618,6 +620,7 @@ class Torrent extends EventEmitter {
             this._debug('piece verified %s', index)
             this._markVerified(index)
           } else {
+            this._markUnVerified(index)
```
DiegoRBaquero, line 623:
```suggestion
            this._markUnverified(index)
```

DiegoRBaquero (changes requested):
One nit and good to go

DiegoRBaquero (approved):

lib/torrent.js
DiegoRBaquero: @alxhotel can we resolve this?

alxhotel (approved):

PR #1486, opened 2018-08-26
add utp support

pldubouilh:
Follow up of #1328 - changes were too important since the master branch was partially updated to es6.
Feel free to merge, dismiss or fix (fixing all the conflicts wasn't really pleasant)

PR #1923, opened 2020-10-02
Add uTP support (BEP29)

hicom150:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Heavily inspired in #1328 and #1486 this PR adds uTP (BEP29) support implemented over current master branch and with all tests passing.

fixes https://github.com/webtorrent/webtorrent/issues/68
closes #1486

I initialized uTP as false by default until we find it is stable and works as expected 😅
I've also added some test that are quite slow as they wait for timeouts, so we can disable them once everything works as expected 😉

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ const Peer = require('simple-peer')
 const randombytes = require('randombytes')
 const speedometer = require('speedometer')
 
-const TCPPool = require('./lib/tcp-pool') // browser exclude
+const ConnPool = require('./lib/conn-pool') // browser exclude
 const Torrent = require('./lib/torrent')
 const VERSION = require('./package.json').version
 
@@ -72,6 +72,7 @@ class WebTorrent extends EventEmitter {
     this.tracker = opts.tracker !== undefined ? opts.tracker : {}
     this.torrents = []
     this.maxConns = Number(opts.maxConns) || 55
+    this.utp = opts.utp === true
 
     this._debug(
       'new webtorrent (peerId %s, nodeId %s, port %s)',
@@ -95,8 +96,8 @@ class WebTorrent extends EventEmitter {
       }
     }
 
-    if (typeof TCPPool === 'function') {
-      this._tcpPool = new TCPPool(this)
+    if (typeof ConnPool === 'function') {
+      this._connPool = new ConnPool(this)
     } else {
       process.nextTick(() => {
         this._onListening()
@@ -355,8 +356,8 @@ class WebTorrent extends EventEmitter {
 
   address () {
     if (!this.listening) return null
-    return this._tcpPool
-      ? this._tcpPool.server.address()
+    return this._connPool
+      ? this._connPool.tcpServer.address()
       : { address: '0.0.0.0', family: 'IPv4', port: 0 }
   }
 
@@ -377,9 +378,9 @@ class WebTorrent extends EventEmitter {
       torrent.destroy(cb)
     })
 
-    if (this._tcpPool) {
+    if (this._connPool) {
       tasks.push(cb => {
-        this._tcpPool.destroy(cb)
+        this._connPool.destroy(cb)
       })
     }
 
@@ -394,7 +395,7 @@ class WebTorrent extends EventEmitter {
     if (err) this.emit('error', err)
 
     this.torrents = []
-    this._tcpPool = null
+    this._connPool = null
     this.dht = null
   }
 
@@ -402,9 +403,9 @@ class WebTorrent extends EventEmitter {
     this._debug('listening')
     this.listening = true
 
-    if (this._tcpPool) {
+    if (this._connPool) {
       // Sometimes server.address() returns `null` in Docker.
-      const address = this._tcpPool.server.address()
+      const address = this._connPool.tcpServer.address()
       if (address) this.torrentPort = address.port
     }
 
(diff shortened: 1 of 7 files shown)


lib/conn-pool.js
```diff
@@ -92,7 +127,7 @@ class TCPPool {
     self._pendingConns.push(conn)
     conn.once('close', cleanupPending)
 
-    const peer = Peer.createTCPIncomingPeer(conn)
+    const peer = this.utpServer ? Peer.createUTPIncomingPeer(conn) : Peer.createTCPIncomingPeer(conn)
```
KayleePop, line 130: Doesn't this assume that if UTP is enabled every incoming connection will be UtP? (from both TCPServer and UTPServer)

lib/torrent.js
```diff
@@ -750,7 +751,7 @@ class Torrent extends EventEmitter {
       }
     }
 
-    const wasAdded = !!this._addPeer(peer)
+    const wasAdded = !!this._addPeer(peer, this.client.utp ? 'utp' : 'tcp')
```
KayleePop, line 754: just a note for other reviewers: if the utp connection fails to connect, then it is replaced with a tcp connection to the same ip:port

maybe a comment here explaining that would be good

lib/peer.js
```diff
@@ -45,6 +46,22 @@ exports.createTCPIncomingPeer = conn => {
...
+  const addr = `${conn.remoteAddress}:${conn.remotePort}`
+  const peer = new Peer(addr, 'utpIncoming')
+  peer.conn = conn
+  peer.addr = addr
+
+  peer.onConnect()
+
+  return peer
+}
+
```
KayleePop, line 64: seems like you could make these into one function to be more DRY but not important.

lib/peer.js
```diff
@@ -57,6 +74,18 @@ exports.createTCPOutgoingPeer = (addr, swarm) => {
...
+ * available connection), the client can attempt to connect to the address.
+ */
+exports.createUTPOutgoingPeer = (addr, swarm) => {
+  const peer = new Peer(addr, 'utpOutgoing')
+  peer.addr = addr
+  peer.swarm = swarm
+
+  return peer
+}
+
```
KayleePop, line 88: seems like you could make these into one function to be more DRY but not important.

KayleePop (changes requested):
the tests don't actually make sure UtP works, just that the downloads complete when the `utp` setting is used on one or both clients. Not required to merge IMO, but it would be nice if possible to actually make sure UtP is working with CI

KayleePop:
it would be good to give commit credit to @pldubouilh somehow for this too if that's possible

pldubouilh:
Thanks ! Nice to revive these stalled commits :+1: 
One point to maybe keep in mind is that depending on `utp-native` would make webtorrent depend on a non-js, native binding module (afaiu for the first time).

lib/torrent.js
```diff
@@ -24,6 +24,7 @@ const sha1 = require('simple-sha1')
 const speedometer = require('speedometer')
 const utMetadata = require('ut_metadata')
 const utPex = require('ut_pex') // browser exclude
+const utp = require('utp-native') // browser exclude
```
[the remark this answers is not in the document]
hicom150: You are right! Done 👍

index.js
```diff
@@ -14,7 +14,7 @@ const Peer = require('simple-peer')
 const randombytes = require('randombytes')
 const speedometer = require('speedometer')
 
-const TCPPool = require('./lib/tcp-pool') // browser exclude
+const ConnPool = require('./lib/conn-pool') // browser exclude
```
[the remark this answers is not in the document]
hicom150: Done 👍

lib/conn-pool.js
hicom150: As you pointed out it can cause confusion, so I've refactored a bit this part 😉

lib/torrent.js
hicom150: Right! I've added a comment 👍

lib/peer.js
hicom150: Added a helper function to improve this 👍

lib/peer.js
hicom150: Added a helper function to improve this 👍

KayleePop (approved):
:+1:

hicom150:
@KayleePop First of all thank you for the feedback! 💪
I've made all the improvements you have suggested hope it is better now 😅
Regarding the tests that don't actually make sure UtP works, any suggestion on how to approach that would be more than welcome 😉
Last but not least, you are absolutely right that this is all based in the awesome PRs made by @pldubouilh so kudos to him! 🤟

KayleePop:
you could force push with this https://github.blog/2018-01-29-commit-together-with-co-authors/ to give github credit to @pldubouilh 

not sure if that matters to them. Otherwise LGTM! uTP is a big win :rocket:

hicom150:
I've been giving a though to this... 
> One point to maybe keep in mind is that depending on utp-native would make webtorrent depend on a non-js, native binding module (afaiu for the first time).

@pldubouilh
Do you think it could be more appropriate to use  https://github.com/mafintosh/utp-wasm instead of https://github.com/mafintosh/utp-native ? It seems that the share the same API although the wasm one is noted as "much slower" that the native one 😓

feross:
Perhaps we can switch to `utp-wasm` if/when that library's performance improves. In the meantime, hopefully it's fine to use a native dependency -- we'll need to make sure the correct binary is used on each platform in WebTorrent Desktop.

If we end up having install errors on various platforms, then we can change `utp-native` to an optionalDependency and make sure that the library works even when it fails to install.

Issue #1932, opened 2020-10-13
UTP: Torrent Destroy Potential NPE


rhyswilliamsza:
**What version of this package are you using?**
0.108.6 (master head e5bea58),

**What operating system, Node.js, and npm version?**
All

**What happened?**
If torrent.destroy() is called, any dangling (specifically future reconnecting) peers throw NPEs when checking whether the client has UTP enabled.

```        
const newPeer = this._addPeer(peer.addr, this.client.utp ? 'utp' : 'tcp')
                                                ^
TypeError: Cannot read property 'utp' of null
    at Timeout._onTimeout (/root/rocket/api/node_modules/webtorrent/lib/torrent.js:1733:62)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
```

**What did you expect to happen?**
Either all outstanding reconnections attempts should be cancelled if the client is destroyed, or a simple check should be done before executing this command.

**Are you willing to submit a pull request to fix this bug?**
Sure, if this isn't a known bug :)

hicom150:
Thank you @rhyswilliamsza for reporting this! 👍
**uTP** support is something that has been just recently added, so a PR to fix this would be more than welcome 😉
If you find anything other problem related to **uTP** or you need any help with the PR just let me know 😊

hicom150:
Fixed! 🎉

PR #1933, opened 2020-10-16
[BugFix] Peer reconnect timeout throwing NPE on destroyed.

rhyswilliamsza:

[  ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Very small change to the handling of reconnections. When a torrent is destroyed, there is no current way of cancelling all outstanding peer reconnection attempts. This is due to the scoped design of the reconnection timeout reference. Thus, in the case where a torrent is destroyed, reconnection attempts still take place. When this happens, a NPE is thrown due to a missing `client` reference. Ideally, one would be able to cancel all outstanding reconnection attempts, however as described above, this is more challenging than anticipated. As an alternative solution, this simply checks whether the torrent is destroyed, and chokes the reconnection attempt at the start.

Described above. Referenced in https://github.com/webtorrent/webtorrent/issues/1932

If there's a better way of cancelling the `reconnectTimeout` upon `_destroy()`, please let me know! :)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "webtorrent",
   "description": "Streaming torrent client",
-  "version": "0.108.6",
+  "version": "0.108.7",
```
hicom150, line 4: If you don't mind, please undo this change as it will be automatically cincremented with the release process we use 😉

lib/torrent.js
```diff
@@ -1730,6 +1730,7 @@ class Torrent extends EventEmitter {
       )
 
       const reconnectTimeout = setTimeout(() => {
+        if (this.destroyed) return
         const newPeer = this._addPeer(peer.addr, this.client.utp ? 'utp' : 'tcp')
         if (newPeer) newPeer.retries = peer.retries + 1
       }, ms)
```
hicom150, line 1736: Although this check will certainly fix the problem, as you have suggested I would change the `reconnectTimeout` reference scope to class level, in the same way that it is done with `_rechokeIntervalId`. So this code block could be something like this:
```
this._reconnectTimeoutId = setTimeout(() => {
  const newPeer = this._addPeer(peer.addr, this.client.utp ? 'utp' : 'tcp')
  if (newPeer) newPeer.retries = peer.retries + 1
}, ms)
if (this._reconnectTimeoutId.unref) this._reconnectTimeoutId.unref()
```
and we will also need to add in line 89
```
this._reconnectTimeoutId = null
```
and also add in line 661
```
clearTimeout(this._reconnectTimeoutId)
```

hicom150 (changes requested):
Hey @rhyswilliamsza thank you for this great PR, it solves the problem perfectly 👍

As you correctly suggested maybe there could be a more "elegant" way to deal with the cancelling of `reconnectTimeout` so I made a small suggestion 😇

Let me know what do you think 😉

hicom150:
@rhyswilliamsza any chance we can push forward this fix? 😉 I think it is an important one 👍

rhyswilliamsza:
Hi @hicom150! Yes of course, sorry. Have been crazy busy with work. Will have a go this evening!

lib/torrent.js
rhyswilliamsza: Haha, sorry for being busy and blocking this. You should definitely have just run with the changes! xD

rhyswilliamsza:
Sorry for the holdup on the rather simple change. One last thing. Line 1718:

```
      if (peer.retries >= RECONNECT_WAIT.length) {
        if (this.client.utp) {
          const newPeer = this._addPeer(peer.addr, 'tcp')
          if (newPeer) newPeer.retries = 0 <------------------------- 1718
        } else {
          this._debug(
            'conn %s closed: will not re-add (max %s attempts)',
            peer.addr, RECONNECT_WAIT.length
          )
          return
        }
      }
```

Should we not have a return here? It seems like this will attempt to add a TCP peer in the case of UTP, and if successful, immediately jump out and attempt a reconnection after 1s. This reconnection would likely die in `addPeer()` as a duplicate, however it seems a little odd?

hicom150:
> Should we not have a return after this?

You are absolutely right! The `return` of the line `1722` should be placed outside the `if else` so it "returns" in both cases 😅

```
if (peer.retries >= RECONNECT_WAIT.length) {
  if (this.client.utp) {
    const newPeer = this._addPeer(peer.addr, 'tcp')
    if (newPeer) newPeer.retries = 0
  } else {
    this._debug(
      'conn %s closed: will not re-add (max %s attempts)',
      peer.addr, RECONNECT_WAIT.length
    )
  }

  return
}
```

package.json
```diff
@@ -154,4 +154,4 @@
       "webtorrent.chromeapp.js"
     ]
   }
-}
```
hicom150, line 157: I think that we should leave this newline at the end 😉

hicom150 (approved):

package.json
rhyswilliamsza: Bizarre! No clue why that happened. Thanks!

rhyswilliamsza:
> > Should we not have a return after this?
> 
> You are absolutely right! The `return` of the line `1722` should be placed outside the `if else` so it "returns" in both cases 😅
> 
> ```
> if (peer.retries >= RECONNECT_WAIT.length) {
>   if (this.client.utp) {
>     const newPeer = this._addPeer(peer.addr, 'tcp')
>     if (newPeer) newPeer.retries = 0
>   } else {
>     this._debug(
>       'conn %s closed: will not re-add (max %s attempts)',
>       peer.addr, RECONNECT_WAIT.length
>     )
>   }
> 
>   return
> }
> ```

Awesome, thought as much. Have made the change :)

hicom150:
Nice work @rhyswilliamsza 💪 It's been a pleasure helping you with this PR 😉

feross:
Nice work @rhyswilliamsza! Thanks @hicom150 for shepherding this PR 👍 

Released as 0.109.1.

feross:
So taking a closer look at this PR, it doesn't seem like it works if there is more than one peer that times out since `this._reconnectTimeoutId` will just get overwritten the next time we call `setTimeout`. I think the cleanest way to solve this was probably the original approach, even though it leaves a few timers around for a few seconds (not a big deal, since they're `unref`'d which means they won't keep the process alive.

feross:
Sent a PR: https://github.com/webtorrent/webtorrent/pull/1944

hicom150:
I'm sorry I pointed in the wrong direction 😓
I've approved and merged #1944 😉

feross:
Thanks for reviewing the PR, @hicom150. Released as 0.109.2.

Issue #1939, opened 2020-10-23
Support 'x.pe' parameter on magnet URIs


pataquets:
**What version of this package are you using?**
All

**What problem do you want to solve?**
When opening/decoding magnet links, parsing x.pe parameter would allow faster download and bootstrapping of torrents.

**What do you think is the correct solution to this problem?**
More info: http://bittorrent.org/beps/bep_0009.html#magnet-uri-format

**Are you willing to submit a pull request to implement this change?**
Not able

feross:
Thanks for tackling this task in #1942, @hicom150. I super appreciate it!

feross:
Fixed in https://github.com/webtorrent/webtorrent/pull/1942

feross:
Released as 0.110.0

PR #1942, opened 2020-10-26
Implement magnet peer-address (x.pe) BEP09

hicom150:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Implement magnet peer-address (x.pe) BEP09.

https://github.com/webtorrent/webtorrent/issues/1939

Before this PR can be merged, we need to take some steps

- [x] Merge https://github.com/webtorrent/magnet-uri/pull/38
- [x] Publish new version of https://github.com/webtorrent/magnet-uri/
- [x] Publish new version of https://github.com/webtorrent/parse-torrent/
- [x] Update package.json dependencies

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -323,10 +323,8 @@ class Torrent extends EventEmitter {
     }
 
     // add BEP09 peer-address
-    // TODO: It should only be included if the client can discover its public IP address and determine its reachability.
-    if (this.peerAddress) {
-      const peers = Array.isArray(this.peerAddress) ? this.peerAddress : [this.peerAddress]
-      peers.forEach(peer => this.addPeer(peer))
+    if (this.peerAddresses) {
+      this.peerAddresses.forEach(peer => this.addPeer(peer))
     }
 
     // begin discovering peers via DHT and trackers
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -322,6 +322,13 @@ class Torrent extends EventEmitter {
       })
     }
 
+    // add BEP09 peer-address
+    // TODO: It should only be included if the client can discover its public IP address and determine its reachability.
+    if (this.peerAddress) {
+      const peers = Array.isArray(this.peerAddress) ? this.peerAddress : [this.peerAddress]
+      peers.forEach(peer => this.addPeer(peer))
+    }
+
     // begin discovering peers via DHT and trackers
     this.discovery = new Discovery({
       infoHash: this.infoHash,
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "multistream": "^4.0.0",
     "package-json-versionify": "^1.0.4",
     "parse-numeric-range": "^1.2.0",
-    "parse-torrent": "^7.1.3",
+    "parse-torrent": "^9.0.0",
     "pump": "^3.0.0",
     "random-iterate": "^1.0.1",
     "randombytes": "^2.1.0",
--- a/test/node/download-from-ip.js
+++ b/test/node/download-from-ip.js
@@ -52,3 +52,54 @@ test('Download via torrent.addPeer()', (t) => {
     })
   })
 })
+
+test('Download via magnet x.pe (BEP09)', (t) => {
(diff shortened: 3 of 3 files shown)


feross:
Published new versions of `magnet-uri` and `parse-torrent`.

lib/torrent.js
```diff
@@ -323,6 +323,13 @@ class Torrent extends EventEmitter {
       })
     }
 
+    // add BEP09 peer-address
+    // TODO: It should only be included if the client can discover its public IP address and determine its reachability.
+    if (this.peerAddress) {
```
feross, line 328: This needs to be `peerAddresses` since the API changed in `magnet-uri`.

lib/torrent.js
```diff
@@ -323,6 +323,13 @@ class Torrent extends EventEmitter {
...
+      const peers = Array.isArray(this.peerAddress) ? this.peerAddress : [this.peerAddress]
```
feross, line 329: Upon seeing this, I think we should change the `magnet-uri` API so that `peerAddresses` is always an array, since that's what the plural name indicates. If the user wants the true raw value, they can access `x.pe`, but the `peerAddresses` API is ours so we can make it friendlier if we want. And it saves everyone from needing to do this `isArray` check.

test/node/download-from-ip.js
```diff
@@ -52,3 +52,54 @@ test('Download via torrent.addPeer()', (t) => {
     })
   })
 })
+
+test('Download via magnet x.pe (BEP09)', (t) => {
```
feross, line 56: Test looks good!

feross (changes requested):

feross:
I just submitted a new PR to `magnet-uri` making the changes I suggested: https://github.com/webtorrent/magnet-uri/pull/39

@hicom150 Once that's reviewed and published, do you want to finish up this PR?

lib/torrent.js
```diff
@@ -323,6 +323,13 @@ class Torrent extends EventEmitter {
       })
     }
 
+    // add BEP09 peer-address
+    // TODO: It should only be included if the client can discover its public IP address and determine its reachability.
```
feross, line 327: What's the downside of not following this line, which I think comes from the spec?

hicom150: I left this as a `TODO` as I don't know how to implement this check 😅
Any help or suggestion here will be more than welcome ❤️

test/node/download-from-ip.js
hicom150: ❤️

lib/torrent.js
hicom150: Changed! 👍

lib/torrent.js
hicom150: Fixed! 👍

lib/torrent.js
feross: Oh, re-reading it again, I think I understand what it means! It's saying that if you're creating a magnet link you should not include your own IP address until you are sure that your IP is public and reachable. For example, if your IP is 192.168.1.100, it's a private IP and there's no point including it.

Since this PR is just about processing these URLs and not generating them, I think we can safely ignore it and remove the TODO.

package.json
```diff
@@ -54,7 +54,7 @@
     "multistream": "^4.0.0",
     "package-json-versionify": "^1.0.4",
     "parse-numeric-range": "^1.2.0",
-    "parse-torrent": "^7.1.3",
+    "parse-torrent": "^8.1.0",
```
feross, line 57: Please update parse-torrent to ^9.0.0 and magnet-uri to ^6.0.0

feross (approved):
Please update the versions in `package.json` and then this is ready to merge!

package.json
hicom150: Updated! 👍

lib/torrent.js
hicom150: Done! 👍

feross:
Released as 0.110.0

PR #1950, opened 2020-11-03
Update .airtap.yml

I-Cat:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x ] Other, please explain:

Android has a system webviewer that should be considered when trying to test the browsers.
Android webviewer is more compatable accrossed android systems as some android versons will not be supported
I Wounder If you could make a Organization so that we can have badges under our id if you will accept me of course 😄

### DIFF
--- a/.airtap.yml
+++ b/.airtap.yml
@@ -1,14 +1,16 @@
 loopback: airtap.local
 browsers:
+  - name: Android
+ version: latest
+  - name: Android System WebView
+ version: 81.0.4044.138
   - name: chrome
-    version: latest
+ version: latest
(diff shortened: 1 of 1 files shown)


feross:
We're not interested in supporting the system viewer. We have limited resources and it's only worth our time to support Chrome on Android.

> I Wounder If you could make a Organization so that we can have badges under our id if you will accept me of course 😄

People who contribute good work over a long period of time are eligible to be added to the organization.

Also on a related note, please stop repeatedly inviting me to your @I-Meow organization or I will be forced to block you.

PR #1959, opened 2020-11-10
Remove erronous `.default` at the end of `require("bitfield").default`

justchen1369:
[x] Bug fix

In `torrent.js` and `webconn.js`, bitfield is imported with `const BitField = require('bitfield').default`. However, the `default` property is nonexistent.


justchen1369:
Whoops, was using bitfield 3.0.0 for whatever reason.

feross:
No worries – probably just forgot to `npm install`

Issue #1961, opened 2020-11-11
Remaining `bittorrent-lsd` issues


feross:
We need to:

- [x] Pass the `lsd` option through to `torrent-discovery`, like how we do for the `tracker` and `dht` options.

- [x] Document the `lsd` option in the webtorrent docs, like how we do for the `tracker` and `dht` options.

- [x] Fix the webtorrent tests so that tests which e.g. test the DHT are not only disabling the tracker (with `tracker: false`) but also disable the lsd functionality (with `lsd: false`). Right now, peers are discovered through LSD when they shouldn't be.

- [x] [Added due to @justchen1369's comment] bittorrent-lsd currently only uses dgram, but should use chrome-dgram when running `npm build chromeapp".

justchen1369:
Additionally: bittorrent-lsd currently only uses dgram, but should use chrome-dgram when running `npm build chromeapp".

justchen1369:
I spent a fair amount of time figuring out why the chromeapp build didn't work :)

feross:
@justchen1369 Good point! Yep. Added it to the list above

justchen1369:
would adding the chromeapp field to the `package.json` work?

feross:
> would adding the chromeapp field to the package.json work?

Yep, just merged your PR https://github.com/webtorrent/bittorrent-lsd/pull/6

feross:
Released 0.111.0.

PR #1962, opened 2020-11-12
Remaining `bittorrent-lsd` issues

hicom150:

[x] Documentation update
[x] Bug fix
[x] New feature
[ ] Other, please explain:

- [x] Add `lsd` option to client.
- [x] Document `lsd` option.
- [x] Fix test disabling `lsd`option.

https://github.com/webtorrent/webtorrent/issues/1961

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,6 +70,7 @@ class WebTorrent extends EventEmitter {
     this.torrentPort = opts.torrentPort || 0
     this.dhtPort = opts.dhtPort || 0
     this.tracker = opts.tracker !== undefined ? opts.tracker : {}
+    this.lsd = opts.lsd === true
     this.torrents = []
     this.maxConns = Number(opts.maxConns) || 55
     this.utp = opts.utp === true
--- a/index.js
+++ b/index.js
@@ -70,7 +70,7 @@ class WebTorrent extends EventEmitter {
     this.torrentPort = opts.torrentPort || 0
     this.dhtPort = opts.dhtPort || 0
     this.tracker = opts.tracker !== undefined ? opts.tracker : {}
-    this.lsd = opts.lsd === true
+    this.lsd = opts.lsd !== false
     this.torrents = []
     this.maxConns = Number(opts.maxConns) || 55
     this.utp = opts.utp === true
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -334,14 +334,16 @@ class Torrent extends EventEmitter {
       dht: !this.private && this.client.dht,
       tracker: trackerOpts,
       port: this.client.torrentPort,
-      userAgent: USER_AGENT
+      userAgent: USER_AGENT,
+      lsd: this.client.lsd
     })
 
     this.discovery.on('error', (err) => {
       this._destroy(err)
     })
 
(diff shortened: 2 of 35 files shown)


index.js
```diff
@@ -70,6 +70,7 @@ class WebTorrent extends EventEmitter {
     this.torrentPort = opts.torrentPort || 0
     this.dhtPort = opts.dhtPort || 0
     this.tracker = opts.tracker !== undefined ? opts.tracker : {}
+    this.lsd = opts.lsd === true
```
feross, line 73: If no `opts.lsd` value is specified, then this will be `false`. I think you meant:

```js
opts.lsd !== false
```

lib/torrent.js
```diff
@@ -334,14 +334,16 @@ class Torrent extends EventEmitter {
...
+      lsd: this.client.lsd
     })
 
     this.discovery.on('error', (err) => {
       this._destroy(err)
     })
 
-    this.discovery.on('peer', (peer) => {
+    this.discovery.on('peer', (peer, source) => {
+      this._debug('peer %s discovered via %s', peer, source)
```
feross, line 346: This is a useful log. Thanks for adding!

test/node/download-lsd-magnet.js
```diff
@@ -0,0 +1,37 @@
...
+
+  client2.on('error', function (err) { t.fail(err) })
+  client2.on('warning', function (err) { t.fail(err) })
+
+  // Start seeding
+  client2.seed(fixtures.leaves.content, {
+    name: 'Leaves of Grass by Walt Whitman.epub',
+    announce: []
+  }, function (torrent) {
+    torrent.discovery.lsd._announce() // hack to send a lsd announce skipping 5min interval
```
feross, line 23: You might be able to clean up this hack by creating `client1` before `client2` so that the first announce that `client2` sends here https://github.com/webtorrent/bittorrent-lsd/blob/ded22d3e79d97b5082f5eff667be649113596850/index.js#L141 will immediately notify `client1`.

feross (approved):
LGTM! See inline comments since there's one problem to fix before merge. Also, I think the test can be improved -- see comment. If my suggestion doesn't actually work for some reason, then feel free to just merge this as-is :) Thank you!

index.js
justchen1369: another nifty trick is double negation: `!!opts.lsd`

feross: @justchen1369 Yours is a bit different – it coerces the argument to a boolean, then inverts it. In this case, doesn't matter much. But in general, I lean towards explicitness. If they pass anything other than `false`, it should just stick with the default, which is `true`.

test/node/download-lsd-magnet.js
hicom150: Thank you! That was a helpful tip 😉

lib/torrent.js
hicom150: 😊

index.js
hicom150: Fixed the way @feross suggested 👍

feross (approved):

feross:
Released 0.111.0.

Issue #1947, opened 2020-11-01
NWjs/utp-native - "Error: A dynamic link library (DLL) initialization routine failed" when upgrading Webtorrent from 0.108 to 0.109


vankasteelj:
**What version of this package are you using?**
0.109.2

**What operating system, Node.js, and npm version?**
Windows 10, Node 12.13.1, npm 6.12.1

**What happened?**
Upgrading from 108.6 to 109.2 broke something. I believe this PR is to blame, because utp-native gives loads of errors: https://github.com/webtorrent/webtorrent/commit/692b3d113c0a08a17ee53bca7e16180a94b4fbfe#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R77 

This is my snippet for reproducing the issue (but it breaks everytime I use webtorrent for anything, actually): 

```js
      let magnet = '<insert a magnet link>'
      const wtorrent = new (require('webtorrent'))()
      let done

      setTimeout(() => {
        if (done) return
        wtorrent.destroy()
      }, 3500)

      wtorrent.add(magnet, (t) => {
        wtorrent.destroy()
        done = true
        console.log('magnet size is', t.length)
      })
```
Here's the line I suspect is causing this, when installing the new webtorrent with npm: 
![mintty_7XxtTJxNBJ](https://user-images.githubusercontent.com/12599850/97801448-21c3e080-1c3d-11eb-9964-ff4b4d0db414.png)

Here is the error message in console when trying to use webtorrent: 
![nw_nMDpQnW1HQ](https://user-images.githubusercontent.com/12599850/97801459-2e483900-1c3d-11eb-99f2-3016a1a2d783.png)

**What did you expect to happen?**
I'm not sure what it "should do", but not erroring, it should just "do magic" and return something to me, be it file size, informations and/or fileparts. 

**Are you willing to submit a pull request to fix this bug?**
I doubt it, I have no idea how most of webtorrent works, it just works :D

alxhotel:
I can't reproduce the bug. The code snippet provided works for me.

Are you using the node version of Webtorrent in a browser?

If so, can you try to use the "browserified" version (`webtorrent.min.js`)?

vankasteelj:
I forgot to say I was using nwjs 0.36.4 (but 0.49.2), it also works for me in console, but not in nwjs

vankasteelj:
It does work with the browserified version. That's a first for me... I've never used a browserified module with nwjs

[...]

PR #1966, opened 2020-11-20
fix: make utp-native optional

hicom150:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

As we have added uTP support via `utp-native`, it seems that in some scenarios it is not possible to use native packages as is. This PR tries to solve this problem by wrapping `utp-native` dependency an exposing `UTP_SUPPORT` property to choose if we can use uTP or not. 

https://github.com/webtorrent/webtorrent/issues/1947


index.js
```diff
@@ -73,7 +74,7 @@ class WebTorrent extends EventEmitter {
     this.lsd = opts.lsd !== false
     this.torrents = []
     this.maxConns = Number(opts.maxConns) || 55
-    this.utp = opts.utp === true
+    this.utp = utp.UTP_SUPPORT ? opts.utp === true : false
```
feross, line 77: Rather than use a ternary operator, this is more clearly expressed as `utp.UTP_SUPPORT && opts.utp === true`

lib/utp.js
```diff
@@ -0,0 +1,13 @@
+module.exports = (() => {
+  try {
+    const utp = require('utp-native')
+    utp.UTP_SUPPORT = true
+    return utp
+  } catch (err) {
+    console.error('WebTorrent: uTP not supported. Using TCP as fallback')
```
feross, line 7: We don't want this message to print out in the browser. Can you update the browser exclude information in `package.json` to exclude `./lib/utp.js` instead of `utp-native` since this is now the proper file to exclude in the browser.

lib/utp.js
```diff
@@ -0,0 +1,13 @@
+module.exports = (() => {
+  try {
+    const utp = require('utp-native')
+    utp.UTP_SUPPORT = true
```
feross, line 4: This modifies the exported value of the package, even for users who `require('utp-native')` in their own app since Node.js caches the exported object. Ideally, we shouldn't modify it.

Can you eliminate the  `UTP_SUPPORT ` property from this file and just export either the result of `require('utp-native')` or an empty object? This way, we treat both cases -- (1) missing uTP support in Node.js and (2) running in the browser -- in the same way. In either of these cases, running `require('./lib/utp')` will return an empty object.

You can still export UTP_SUPPORT if you want, but I'd do it on the WebTorrent object the way we do for `WebTorrent.WEBRTC_SUPPORT`.

feross (changes requested):
Looks good!

A few inline changes, plus can you:

- Move `utp-native` to `optionalDependencies` so install-time build failures won't break the whole `webtorrent` install (though this is different than the reported issue)
- Update the browser exclude in `package.json` to exclude `'./lib/utp.js` instead of `utp-native`.

lib/utp.js
vankasteelj: This error message seems unavoidable, if utp is not supported, shouldn't it be a warning? Or be ignored if specifically passed to webtorrent's options `{utp: false}`?

feross: In the browser there's never a way for uTP to be supported. Also, the message says TCP will be used as a fallback, which is also not true. Also, we usually don't use `console.error` to print warnings to the user in WebTorrent.

Once uTP has stabilized a bit more, we should default it to on by default (when the environment supports it) like we do with WebRTC support.

index.js
hicom150: 👍

hicom150:
I've made the changes you have suggested, hope it is better now 😅
If there is anything else that should be improved, just let me know 😉

ivan1986:
if you need
https://github.com/popcorn-time-ru/webtorrent/tree/optional_utp_native
rebased to master

DiegoRBaquero (approved):
LGTM now, need rebase and @feross approval

alxhotel:
Made a change to reduce the require call, since we can use the `require('lib/utp')` from `ConnPool`

vankasteelj:
Any updates on this? I'm stuck on 0.108.6

DiegoRBaquero (approved):

PR #1977, opened 2020-12-08
Update README.md

Slender1808:
[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Added README badge

This pull is not to solve a specific problem but I thought it could be useful


alxhotel:
Unless other member of the team says otherwise, I'm closing this PR since I think we don't need this badge now

PR #1983, opened 2020-12-21
Show a simple status line while downloading

KennethKinLum:

[X] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

Show a simple status line while downloading

Since it may take 20 minutes or even an hour or a few hours to finish downloading, it may be good to see a status, to see that at least the downloading has started and what speed it is going on with.

This shows a simple status line as to what speed it is moving along and what percentage is done.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -98,9 +98,14 @@ var client = new WebTorrent()
 var magnetURI = 'magnet: ...'
 
 client.add(magnetURI, { path: '/path/to/folder' }, function (torrent) {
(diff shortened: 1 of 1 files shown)


KennethKinLum:
I found that if the line is too long, then at least on a Mac, the line is wrapped, and then `\r` doesn't move the cursor back to the beginning of the line... so it is better to keep the line short, say, less than 70 character wide. 

So I will create another PR.

PR #1984, opened 2020-12-21
Show a simple status line while downloading

KennethKinLum:

[X] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

Show a simple status line while downloading

Since it may take 20 minutes or even an hour or a few hours to finish downloading, it may be good to see a status, to see that at least the downloading has started and what speed it is going on with.

This shows a simple status line as to what speed it is moving along and what percentage is done.


docs/get-started.md
```diff
@@ -98,8 +98,15 @@ var client = new WebTorrent()
...
+  var isDone = false
+
   torrent.on('done', function () {
-    console.log('torrent download finished')
+    console.log('\ntorrent download finished')
+    isDone = true
+  })
+
+  torrent.on('download', function (bytes) {
+    if (!isDone) process.stdout.write(`\rtotal: ${torrent.length}  progress: ${(client.progress * 100).toFixed(2)}%  ${Math.round(client.downloadSpeed / 1024).toFixed(1)} KB/s  downloading...`)
```
alxhotel, line 109: You can use directly `torrent.done`

alxhotel, line 109: I don't think there is any need in adding this condition.

Every time the event `download` is called, is because you have downloaded something.

whothisbeme, line 109:
```suggestion
    process.stdout.write(`\r(${Math.round(torrent.downloaded / Math.pow(1024,2))}/${Math.round(torrent.length / Math.pow(1024,2))}MB) [${Math.round(torrent.downloadSpeed / 1024).toFixed(1)}KB/s][${Math.round(torrent.uploadSpeed / 1024).toFixed(1)}KB/s] (${torrent.numPeers}) [${Math.floor((torrent.timeRemaining / (60 * 1000)) % 60)}:${Math.floor((torrent.timeRemaining / 1000) % 60)}] [${(torrent.progress * 100).toFixed(2)}%]`)  })
```

whothisbeme: I've tested and it does seem that without `isDone` you will continue to get a few `download` calls showing the progress after `done` is called. Is this what you are experiencing as well?

whothisbeme (commented):
I think this LGTM however I usually like to see total downloaded/total size, download speed, upload speed, peers, and progress as well. Also, it would be cool to see an ETA from the `timeRemaining` property. I tried adding minutes & seconds but it doesn't look that good IMO. 

I use a separate function usually to convert to human readable ETA however it isn't a one-liner.

Maybe someone has a better solution for the milliseconds to ETA format?

Also not sure why the `downloadSpeed` is coming from the `client` and not the `torrent` object?

docs/get-started.md
whothisbeme: I've tested and it does seem that without `isDone` you will continue to get a few `download` calls showing the progress after `done` is called. Is this what you are experiencing as well?

KennethKinLum: aha, at first I didn't have the `isDone`, and after the "Done" message is printed, then there is one additional progress message that got printed that messed up the screen.  If the event occurs such that the progress always happens first, and then the `done` occurs, then there is no need for the `isDone` flag.

KennethKinLum: I had one other pull request that printed out more info, but I also noticed that unless if I use 

```
const go_beginning_of_line = "\033[G"
const clear_till_end_of_line = "\033[K"
```
to deal with the output, then using the `\r`, if the line is wider than the terminal's width, then the line would wrap to the next line and `\r` only goes back to the position one of that next line. So the screen will have many lines of output in this case.  So that's why I kept the output short, just to show some status.  Also if I use those ANSI escape sequence, it may not work if the terminal is not compatible with it and I really should use a `curses` library... but then it probably defeats the purpose of the author showing a simple program to show how it can download all the files for a torrent and save them in a directory.

KennethKinLum: true... I really should use `torrent.downloadSpeed` and I will correct it in the next minute.  Just this case it is one torrent so the client and torrent have the same speed.

DiegoRBaquero: We could defer done to next event loop to solve this

alxhotel: Yep, I agree @DiegoRBaquero

PR #1999, opened 2021-02-09
Support opts.preloadedStore in webtorrent.seed

jhiesey:

[ ] Documentation update
[ ] Bug fix
[x ] New feature
[ ] Other, please explain:

opts.preloadedStore can be used to seed a torrent using a store that already contains data.

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -53,7 +53,10 @@ class FileStream extends stream.Readable {
     if (this._torrent.destroyed) return this._destroy(new Error('Torrent removed'))
 
     const p = this._piece
-    this._torrent.store.get(p, (err, buffer) => {
+    const length = p === this._torrent.pieces.length - 1 ? this._torrent.lastPieceLength : undefined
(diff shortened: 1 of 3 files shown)


feross (approved):
LGTM

feross:
0.114.0

lib/file-stream.js
```diff
@@ -53,7 +53,10 @@ class FileStream extends stream.Readable {
     if (this._torrent.destroyed) return this._destroy(new Error('Torrent removed'))
 
     const p = this._piece
-    this._torrent.store.get(p, (err, buffer) => {
+    const length = p === this._torrent.pieces.length - 1 ? this._torrent.lastPieceLength : undefined
```
ftreesmilo, line 56: @feross @jhiesey 
wouldn't it be better to pass `this._torrent.pieceLength` instead of `undefined` for stores that might not be written well?

I can see why you might say it's not needed, but it just feels like it would be more complete to pass in the value we want, because we know what it is, and we're making a choice here anyway.

PR #2000, opened 2021-02-12
Add speed limit v2

greyones:

[ ] Documentation update
[ ] Bug fix
[ x ] New feature
[ ] Other, please explain:

This is the PR [#1969](https://github.com/webtorrent/webtorrent/pull/1969) by [hicom150](https://github.com/hicom150). Since there has been no recent activity on PR [#1969](https://github.com/webtorrent/webtorrent/pull/1969) and there were only a couple of minor review comments left to be resolved. So I created this PR based PR ([#1969](https://github.com/webtorrent/webtorrent/pull/1969)) + I resolved the review comments.

What changes did you make? (Give an overview)
Adds global speed limits. Based on the work done in [#1040](https://github.com/webtorrent/webtorrent/pull/1040), [#1852](https://github.com/webtorrent/webtorrent/pull/1852) and [#1969](https://github.com/webtorrent/webtorrent/pull/1969)

Which issue (if any) does this pull request address?
There where some unresolved review comments from the PR #1969, which is resolved in this PR.

Is there anything you'd like reviewers to focus on?


ThaUnknown:
I found 2 issues with this, if the download/upload limit opts are not specified, the torrent will not download at all, and for some reason it says it uploads 1/1000 of the download speed and vice versa, even when not uploading/downloading any data

mathiasvr:
The chunksize of the `ThrottleGroup` seems to become too big by default.
This can help fix it, but we should probably also adjust the chunksize when changing speed limits.
Also, it might be better to bypass the throttle stream by default.
```js
    this.throttleGroups = {
      down: new ThrottleGroup({ rate: this.downloadLimit, chunksize: Math.min(this.downloadLimit / 10, 1000000) }),
      up: new ThrottleGroup({ rate: this.uploadLimit, chunksize: Math.min(this.uploadLimit / 10, 1000000) })
    }
```

ftreesmilo:
> Also, it might be better to bypass the throttle stream by default.

It sounds to me that this may be a really good way to pause torrents (dl / ul speed set to 0).
Does this introduce a ton of overhead?

mathiasvr:
> It sounds to me that this may be a really good way to pause torrents (dl / ul speed set to 0).
> Does this introduce a ton of overhead?

No, I think it's minimal but I still think it might be better no bypass it when there is no limit, meaning you could still set dl/ul to 0 to enable it and pause.

alxhotel:
Closed by #2062

PR #2004, opened 2021-02-14
feat: support adding paused torrents.

nkavian:

[X] New feature

The use case I'd like to solve is that a torrent file and it's contents are stored in disk, I'd like to add the torrent to the `client` and be able to benefit from the programmatic interface provided by `client`, `torrent`, and `file`; without being connected to peers.

Another example, if I `pause` a torrent and restart my application, I'd like to maintain that state across restarts.

Adding a torrent to the client and then pausing it is not an adequate workaround.

N/A

I'm new to your code, I'm unaware if I'm changing everything that's necessary. I wanted to be able to send a PR and not just create an issue / feature request.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -94,7 +94,7 @@ class Torrent extends EventEmitter {
 
     this.ready = false
     this.destroyed = false
-    this.paused = false
+    this.paused = opts.paused || false
     this.done = false
 
     this.metadata = null
--- a/test/client-add.js
+++ b/test/client-add.js
@@ -205,3 +205,24 @@ test('client.add: invalid torrent id: short buffer', function (t) {
 
   client.add(Buffer.from('abc'))
 })
+
+test('client.add: paused torrent', function (t) {
+  t.plan(5)
+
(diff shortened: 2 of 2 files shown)


nkavian:
Tests pass but something called airtap failed.

DiegoRBaquero (approved):

DiegoRBaquero:
Do you mind rebasing to re-retrigger CI?

nkavian:
@DiegoRBaquero the change you made was incorrect. you should have sent your own PR and allow people to comment. I would have pointed out that

A) you didn't consistently change it within the same method. You updated lines 214/215 but missed 223 and 226 in the same method.

B) the whole file uses the syntax I used, whereas now, your two lines are different, and you've associated my PR to this mistake.

DiegoRBaquero:
It’s not a mistake, just different syntax. Will switch everything to arrow functions

DiegoRBaquero:
See https://github.com/webtorrent/webtorrent/pull/2134

nkavian:
I know it's a different syntax.

PR #2008, opened 2021-02-26
Allow passing a custom connection object to torrent.addWebSeed

jhiesey:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Allow passing a custom web seed implementation to `torrent.addWebSeed`. Docs are updated to include the new functionality.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -31,6 +31,7 @@ const File = require('./file')
 const Peer = require('./peer')
 const RarityMap = require('./rarity-map')
 const Server = require('./server') // browser exclude
+const WebConn = require('./webconn')
 
 const MAX_BLOCK_LENGTH = 128 * 1024
 const PIECE_TIMEOUT = 30000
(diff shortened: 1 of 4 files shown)


lib/torrent.js
```diff
@@ -820,28 +821,49 @@ class Torrent extends EventEmitter {
...
+      }
+
+      if (this._peers[id]) {
+        this.emit('warning', new Error(`ignoring duplicate web seed: ${id}`))
+        this.emit('invalidPeer', id)
+        return
+      }
+
+      conn = new WebConn(url, this)
+    } else if (urlOrConnection && typeof urlOrConnection.connId === 'string') {
```
feross, line 846: I don't think that checking for a `connId` property will correctly detect the default `WebConn` objects. Is there a better way to detect instances of `Wire`? Could check for a property on `Wire`, like `peerId` as a sanity check?

lib/torrent.js
```diff
@@ -820,28 +821,49 @@ class Torrent extends EventEmitter {
...
+    let conn
+    if (typeof urlOrConnection === 'string') {
+      const url = urlOrConnection
+      id = urlOrConnection
+
+      if (!/^https?:\/\/.+/.test(url)) {
+        this.emit('warning', new Error(`ignoring invalid web seed: ${id}`))
+        this.emit('invalidPeer', id)
+        return
+      }
```
feross, line 837: The mixing of `id` and `url` here is a bit confusing

feross (changes requested):

PR #2015, opened 2021-03-08
Only pass length to store.get when defined

### LINKED ISSUE #2014: Unable to add torrent from store
While re-adding torrent from store, the `_verifyPieces` method failed with `piece invalid` error.

For example, this would failed, even if stored datas are valid:

`client.add(parsedTorrent, {store: idb})`

This line is the problem: 
[torrent.js#L589](https://github.com/webtorrent/webtorrent/blob/a860773ec4d379cb10922b60de3489559c6913cd/lib/torrent.js#L589)

`const length = index === this.pieces.length - 1 ? this.lastPieceLength : undefined`

Should be i guess:

`const length = index === this.pieces.length - 1 ? this.lastPieceLength : this.pieceLength`

Not sure the reason `length` is set to `undefined`?!

jhiesey:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Only pass length in `store.get` options when it is defined

Fixes #2014

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -53,10 +53,13 @@ class FileStream extends stream.Readable {
     if (this._torrent.destroyed) return this._destroy(new Error('Torrent removed'))
 
     const p = this._piece
-    const length = p === this._torrent.pieces.length - 1 ? this._torrent.lastPieceLength : undefined
-    this._torrent.store.get(p, {
-      length
-    }, (err, buffer) => {
+
+    const getOpts = {}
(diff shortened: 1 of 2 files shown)


jhiesey:
`indexeddb-chunk-store` is testing if `opts.length` is passed using `'length' in opts` instead of `opts.length !== undefined`. This is a workaround.

feross (approved):
LGTM – thanks for the fix!

feross:
Published as 0.115.3.

PR #2018, opened 2021-03-20
Made 3 arrays into a set

jimmywarting:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain: Refactor

What's looks to be a unik array could instead use Set, and use the delete method instead of 
arrayRemove(item, item.indexOf(onReady))

### DIFF
--- a/lib/conn-pool.js
+++ b/lib/conn-pool.js
@@ -1,4 +1,3 @@
-const arrayRemove = require('unordered-array-remove')
 const debug = require('debug')('webtorrent:conn-pool')
 const net = require('net') // browser exclude
 const utp = require('utp-native') // browser exclude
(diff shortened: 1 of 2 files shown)


alxhotel (approved):
LGTM

lib/conn-pool.js
```diff
@@ -1,4 +1,3 @@
-const arrayRemove = require('unordered-array-remove')
```
DiegoRBaquero, line 1: Could `unordered-array-remove` dep be removed or are the still places where it's used?

DiegoRBaquero (approved):
LGTM

lib/conn-pool.js
ThaUnknown: if you mean this specific file, then it's not used, but unordered-array-remove is still used in other files such as [peer.js](https://github.com/webtorrent/webtorrent/blob/master/lib/peer.js#L272)

jimmywarting: I think i only touched the internals (no public facing properties) that would mean any breaking changes. that's why i think i left out peer.js perhaps.

PR #2023, opened 2021-04-07
Add speed limit v3/v4

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

This is the continuation of a continuation of [#1969](https://github.com/webtorrent/webtorrent/pull/1969) by [hicom150](https://github.com/hicom150), because again, the PR went stale.

Overall:
added a download/upload speed limit which can be enabled by specifying `downloadLimit` or `uploadLimit` in the opts when declaring a `new WebTorrent` instance. Setting this value to 0 or not specifying it bypasses the respective `ThrottleGroup`

From last PR:
Bypassing `ThrottleGroup` as described above, reverted the `download` and `upload` emitting changes, meaning dl/up speeds are calculated before throttle, doing this as it was throttled caused too many issues which somehow noone noticed before [speed displayed as 0, speeds were not representative of actual speeds, they were closer to x1.1 than x1.0, upload/download had an extra 1/1000 of download/upload added on top of it [????]]

#2000 #1969 #1852 #1762 #1040 #163
is there a need for accurate download speeds, if yes, is there a better way of doing it

speed limit tests


ThaUnknown:
speed cap tests fail because of [I assume] how throttlegroup calculates averages, so during the initial start of the download it will spike above the cap in the 1st second, I'm not too sure how to address this test

ThaUnknown:
https://github.com/webtorrent/webtorrent/pull/2062

PR #2024, opened 2021-04-13
Make file.downloaded getter not throw when destroyed

jhiesey:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

When the `file.downloaded` getter was called after the file (or associated torrent) was destroyed, it would throw.

Other methods on `file` still will throw if called once destroyed. But IMHO getters should be protected at least.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -36,7 +36,7 @@ class File extends EventEmitter {
   }
 
   get downloaded () {
-    if (!this._torrent.bitfield) return 0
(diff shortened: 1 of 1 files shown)


jhiesey:
I ran all tests locally; `All browsers passed`

alxhotel (approved):

feross (approved):

feross:
0.116.3

PR #2025, opened 2021-04-14
Much better timeouts for webseeds

jhiesey:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Currently, there is the standard 30 second timeout to complete one or more HTTP requests to get a piece from a webseed. Depending on the piece size, this may cause incorrect timeouts.

Instead, this PR moves the timeouts into `lib/webconn.js` and applies them as a socket timeout (interval between packets/data events), NOT a timeout for the entire request. This timeout is set to 60 seconds in this PR.

If an HTTP request fails, the webseed will cancel its requests using the `lt_donthave` extension, and then re-advertise them after a delay (currently 10 seconds) using a `have` message. This allows retries to work without completely hammering webseeds.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -11,6 +11,7 @@ const fs = require('fs')
 const FSChunkStore = require('fs-chunk-store') // browser: `memory-chunk-store`
 const get = require('simple-get')
 const ImmediateChunkStore = require('immediate-chunk-store')
+const ltDontHave = require('lt_donthave')
 const MultiStream = require('multistream')
 const net = require('net') // browser exclude
 const os = require('os') // browser exclude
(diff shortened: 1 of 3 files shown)


jhiesey:
To work properly in the browser, this needs an updated version of `stream-http` from this PR: https://github.com/jhiesey/stream-http/pull/121

That should be merged shortly.

feross:
This completely fixes the issue where a web seed may hang and the torrent will never complete because webtorrent becomes uninterested in the web seed peer!

feross:
0.117.0

jhiesey:
`stream-http` with fixed timeout support is `v3.2.0`

PR #2030, opened 2021-04-21
FileStream: remove onclose argument to destroy()

feross:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

FileStream: remove onclose argument to destroy(). This is a non-standard interface and it's confusing. It caused a bug in https://wormhole.app

This is a breaking change, so we'll release it as a minor version (we're still below 1.0.0)

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -85,11 +85,7 @@ class FileStream extends stream.Readable {
     this._piece += 1
   }
 
-  destroy (onclose) {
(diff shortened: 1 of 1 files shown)


jhiesey (approved):
LGTM. Fixes exceptions when only `err` is passed to `destroy`

feross:
0.118.0

PR #1516, opened 2018-09-25
Global download and upload limit

kdiogenes:
I updated PR #1040 with the current code base. I will start to look for the test suite soon.

PR #2062, opened 2021-05-14
feat: add speed limit to client

### LINKED ISSUE #163: Limit download/upload speeds (throttling)

alxhotel:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

- Adds speed throttling for uploading and downloading data
- Adds two new attributes in the constructor options: `downloadLimit` and `uploadLimit`
- Adds two new methods to the client: `thottleDownload` and `throttleUpload`

This work is based on these PRs: #1516, #1762, #1852, #1969, #2000 and #2023

This PR intends to close #163 once and for all

Please check if the new attributes and methods of the API can be improved; in other words, if it makes sense to the user of the library or if it can be changed to be more friendly and intuitive.

TODO:
- ~Add more tests for all the possible values of `rate` in `throttleUpload` and `throttleDownload`~


ThaUnknown:
I'm having some issues with this....
even tho my upload and download limit is 73400320, it never reaches > 1.5MiB/s, when not defining any limit it doesn't seem to have this issue and it reaches high speeds, even though it's still going through the throttle groups, additionally whenever anything is being uploaded/downloaded the speedometer reports 1/1000 of upload/download as download/upload which looks something like this:
image

I'm kind of conflicted as to pausing torrents since by setting any of the limits to 0 you effectively pause all torrents globally, but still create new peers, ideally when upload/download is disabled webtorrent shouldn't connect to new leechers/seeders?

alxhotel:
Hi @ThaUnknown,

Thanks for the feedback! I have made a small example of how is the speed limiter wotking: https://codepen.io/alxhotel/pen/ExWNLog

Please let me know if you are still seeing the same problems when setting the limits with a number bigger than zero and `-1`.

ThaUnknown:
> Hi @ThaUnknown,
> 
> Thanks for the feedback! I have made a small example of how is the speed limiter wotking: https://codepen.io/alxhotel/pen/ExWNLog
> 
> Please let me know if you are still seeing the same problems when setting the limits with a number bigger than zero and `-1`.

thanks for getting back to me

with a speed limit:
image
without a speed limit:
image

both the upload "artifacting" and weirdly low upload speed limits still happen even on that codepen, I changed from sintel to some other torrent I hosted locally for testing

this happens on all browers I tested: chrome, edge, firefox, brave, even when declaring a speed limit before adding the torrent and during the downloading

SilentBot1:
I've attempted to go through and replicate the issue reported by @ThaUnknown, and have received the following results when using the codepen provided above.

The results I've received are the following:
image

- Prior to enabling the throttle, downloads were running at a normal download speed.
- While the throttle was enabled, with both the Upload and Download throttle being set to 50000 bytes/s (I also tested with larger sizes, with the same results), the upload and download events returned that the client was only uploading and downloading 1 byte per event (only download pictured above).
- After disabling the throttle, due to the number of events (and thus my modification of calling `console.log`), the client remained unresponsive. Without my modification, the client remains responsive, but the download and upload rates are still 1 byte per event.

This was tested on a Windows 10 20H2 device, running Microsoft Edge (Chromium) Version 90.0.818.66 (latest).

If I can provide any further information to help with this, just let me know.

alxhotel:
Hi @ThaUnknown and @SilentBot1,

Thank you for the testing :+1: 

I have updated the version of the package that throttles the speed (`speed-limiter`) and I think all the problems mentioned are now solved (I have also updated the codepen)

If you want you can check again if the issues mentioned have been fixed.

ThaUnknown:
> Hi @ThaUnknown and @SilentBot1,
> 
> Thank you for the testing 👍
> 
> I have updated the version of the package that throttles the speed (`speed-limiter`) and I think all the problems mentioned are now solved (I have also updated the codepen)
> 
> If you want you can check again if the issues mentioned have been fixed.

@alxhotel I'm running into issues where changing the throttle speed to many times stops any uploading/downloading, also the issue with reporting 1/1000 of the upload as download, or the other way around still happens

alxhotel:
> @alxhotel ~I'm running into issues where changing the throttle speed to many times stops any uploading/downloading

Do you have the steps to reproduce when your uploading/downloading stops?

>  also the issue with reporting 1/1000 of the upload as download, or the other way around still happens

I think the upload speed being 1/1000 seems ok, since you are only requesting the pieces to an web seed.

> it also seems like chunksize sometimes gets set to a float not integer which could cause issues too?

What are the steps to reproduce this?

> download stops when upload is also set to 0

The download stops when upload is 0 because the client can not request more pieces if the upload is 0.

ThaUnknown:
> Do you have the steps to reproduce when your uploading/downloading stops?

This was caused by:
> The download stops when upload is 0 because the client can not request more pieces if the upload is 0.

My bad.
> What are the steps to reproduce this?

This doesn't seem to cause any big issues, but just in case, setting the speed to for example: `100000.1`
> I think the upload speed being 1/1000 seems ok, since you are only requesting the pieces to a web seed.

I used a local torrent, not a webseeded one for testing this.
As you said, this is caused by requesting additional pieces, I don't think requesting pieces should be reported as upload speed in this case, since it might be misleading, especially for UX, I also don't think that requesting pieces should count to speed caps even tho it's a nice way of working around people that only leech :)

alxhotel:
> This was caused by:

Do you mean is caused by the text below? Due to the upload speed being 0?

> This doesn't seem to cause any big issues, but just in case, setting the speed to for example: 1000001

Got it! i have fixed this in the new version of `speed-limiter`.

> I used a local torrent, not a webseeded one for testing this.

I've fix the bug that reduces the speed even with a high max speed. I have tested this with a local file and I'm able to reach approximately the same speed as without the throttling enabled. Although my speed in both cases is ~7MB/s, so let me know if it works for you (since you seem to have higher speeds locally)

ThaUnknown:
> Do you mean is caused by the text below? Due to the upload speed being 0?

yes
> I've fix the bug that reduces the speed even with a high max speed. I have tested this with a local file and I'm able to reach approximately the same speed as without the throttling enabled. Although my speed in both cases is ~7MB/s, so let me know if it works for you (since you seem to have higher speeds locally)

This now seems fixed, I was able to reach speeds of 250-300 Mbit/s, let's hope there isn't much overhead here tho.

Crauzer:
Any update on this please ?
I find it ridiculous that this isn't a feature yet.

DiegoRBaquero:
> Any update on this please ?
> 
> I find it ridiculous that this isn't a feature yet.

This PR is the latest progress towards this feature. 

Why do you find it ridiculous? For many years unpaid contributors and maintainers haven't aligned, that's it. 

If you are in the need of certain features and are unwilling to code, please sponsor the project or post a bounty

Crauzer:
> > Any update on this please ?
> > I find it ridiculous that this isn't a feature yet.
> 
> This PR is the latest progress towards this feature.
> 
> Why do you find it ridiculous? For many years unpaid contributors and maintainers haven't aligned, that's it.
> 
> If you are in the need of certain features and are unwilling to code, please sponsor the project or post a bounty

Because this is pretty much a "basic" feature of any kind of torrent client out there.
The word "ridiculous" is maybe a bit too much but it surely is quite sad that it's still not implemented in such a huge library.

Having the ability to limit download/upload speeds on individual torrents or even the client would be a godsend.

docs/api.md
```diff
@@ -60,8 +60,11 @@ If `opts` is specified, then the default options (shown below) will be overridde
   dht: Boolean|Object,     // Enable DHT (default=true), or options object for DHT
   lsd: Boolean,            // Enable BEP14 local service discovery (default=true)
   webSeeds: Boolean,       // Enable BEP19 web seeds (default=true)
+  utp: Boolean,            // Enable BEP29 uTorrent transport protocol (default=false)
```
DiegoRBaquero, line 63: it's defaulted to true

DiegoRBaquero:
```suggestion
  utp: Boolean,            // Enable BEP29 uTorrent transport protocol (default=true)
```

index.js
```diff
@@ -76,11 +77,19 @@ class WebTorrent extends EventEmitter {
     this.maxConns = Number(opts.maxConns) || 55
     this.utp = WebTorrent.UTP_SUPPORT && opts.utp !== false
 
+    this._downloadLimit = Math.max(Number(opts.downloadLimit), -1)
+    this._uploadLimit = Math.max(Number(opts.uploadLimit), -1)
```
DiegoRBaquero, line 81: If opts is not passed (`undefined`), these would end up being NaN

index.js
```diff
@@ -352,6 +361,32 @@ class WebTorrent extends EventEmitter {
       : { address: '0.0.0.0', family: 'IPv4', port: 0 }
   }
 
+  /**
+   * Set global download throttle rate.
+   * @param  {Number} rate (must be bigger or equal than zero, or -1 to disable throttling)
+   */
+  throttleDownload (rate) {
+    rate = Number(rate)
+    if (isNaN(rate) || !isFinite(rate) || rate < -1) return
```
DiegoRBaquero, line 370: how about we return false to indicate it wasn't set?

index.js
```diff
@@ -352,6 +361,32 @@ class WebTorrent extends EventEmitter {
...
+    this.throttleGroups.down.setRate(this._downloadLimit)
+  }
+
+  /**
+   * Set global upload throttle rate
+   * @param  {Number} rate (must be bigger or equal than zero, or -1 to disable throttling)
+   */
+  throttleUpload (rate) {
+    rate = Number(rate)
+    if (isNaN(rate) || !isFinite(rate) || rate < -1) return
```
DiegoRBaquero, line 383: Same here

package.json
```diff
@@ -71,6 +71,7 @@
     "simple-get": "^4.0.0",
     "simple-peer": "^9.11.0",
     "simple-sha1": "^3.1.0",
+    "speed-limiter": "^0.2.0",
```
DiegoRBaquero, line 74: Could you publish 1.0.0 so we follow semver?

test/node/limit-download-upload.js
```diff
@@ -0,0 +1,82 @@
+const fixtures = require('webtorrent-fixtures')
```
DiegoRBaquero, line 1: Could you run `lebab` to modernize this code? See https://github.com/webtorrent/webtorrent/pull/2134/files

test/node/limit-methods.js
```diff
@@ -0,0 +1,88 @@
+const fixtures = require('webtorrent-fixtures')
```
DiegoRBaquero, line 1: Same, LGTM but needs to be consistent, so needs to be modernized

DiegoRBaquero (approved):
LGTM!!

DiegoRBaquero:
Flaky test failing, but good to go!

YOUPERT-INC:
Hi @alxhotel I have applied speed-limiter to my WebTorrent-based Electron app. Seems like speed itself is limited but CPU/RAM usage has not meaningfully decreased. What I expected from speed-limiter was more about hardware usage optimization of Webtorrent but it seems not have worked that way. Could you advise or help me with this? FYI, I'm the 1st sponsor of your repo 😊and willing to donate more if you can help me with this. Thanks. If you're OK, please contact me at bkkecho@gmail.com

Issue #2055, opened 2021-05-08
ES8 Target


DiegoRBaquero:
- [ ] addr-to-ip-port
- [ ] bep53-range
- [ ] bittorrent-dht
- [ ] bittorrent-peerid
- [ ] bittorrent-protocol
- [ ] bittorrent-tracker
- [ ] bittorrent-lsd
- [ ] create-torrent
- [ ] fs-chunk-store
- [ ] instant.io
- [ ] ip-set
- [ ] load-ip-set
- [ ] lt_donthave
- [ ] magnet-uri
- [ ] node-bencode
- [ ] parse-torrent
- [ ] render-media
- [ ] string2compact
- [ ] torrent-discovery
- [ ] torrent-piece
- [ ] ut_metadata
- [ ] ut_pex
- [ ] webtorrent
- [ ] webtorrent-cli
- [ ] webtorrent-desktop
- [ ] webtorrent-fixtures
- [ ] webtorrent-hybrid
- [ ] webtorrent.io

Things we can use:

- `const / let`
- `async / await`
- classes
- block-scoped variables
- arrow functions
- `Array.includes`
- template literals
- de-structuring
- ESM

DiegoRBaquero:
For the first two, they are supported in ES9 (2018), which we don't plan to support YET, because it drops Firefox 78 support. Less use of Buffer is def on the way :)

DiegoRBaquero:
Example https://github.com/webtorrent/bittorrent-tracker/commit/e5994d2ebdec10fe2165e31f5b498382eeeaaf5f

DiegoRBaquero:
Example https://github.com/webtorrent/bittorrent-protocol/commit/3d3e244319036583230d64824ce1388287233e02

king-11:
Hey @DiegoRBaquero I would like to work on it do you have any suggestions on which section I can work on refactoring. Thanks!

PR #2096, opened 2021-06-03
convert to ESM

jimmywarting:

To convert to ESM

one of #2055

- tape & airtap don't support modules
- i tried tape-es and that did work okey
- it's not possible to load package.json (possible solution is to search and replace the files for some version string)

Not sure how to tackle this three...

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,23 +1,25 @@
 /*! webtorrent. MIT License. WebTorrent LLC <https://webtorrent.io/opensource> */
 /* global FileList */
 
-const { EventEmitter } = require('events')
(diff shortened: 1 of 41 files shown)


jimman2003:
this can open the door to use many more modules e.g date-fn instead of moment

jimmywarting:
@DiegoRBaquero do you have any suggestion of how to go forward with this?

DiegoRBaquero:
Yes, or switch test framework entirely, but that'd be lengthy.

Also, if we match with 16 LTS, we can target Node 14 in October, instead of 12.

Issue #1522, opened 2018-10-04
I made a createServer equivalent for browsers


jimmywarting:
With the power of service worker:
https://github.com/jimmywarting/webtorrent-server-browser

Have a look if it's something of an intresse 

---- 

the catch is you need to have the main window running to pipe a file stream to the service worker.
Would ofc be better if the service worker where able to read directly from say indexedDB or some storage layer and wouldn't need to communicate thought postMessage

jimmywarting:
It's similar to how my StreamSaver lib works...

DiegoRBaquero:
Hadn't seen this, it's a really nice experiment. Do you maybe want to submit it to the list of things built with WebTorrent?

jimmywarting:
> Do you maybe want to submit it to the list of things built with WebTorrent?

Nah, it's not even a "thing" (so called project, module or a service) just a experiment as you said.

What I perhaps would like to see is if it would be worth considering implementing this somehow maybe even replace the existing way of how media is played. The benefit of my method is that you will take of lots of work from the main thread, keeping the main thread more responsive. You will also get more support for other containers.

- mp4-box-encoding
- mp4-stream
- mediasource
- render-media
- videostream
- range-slice-stream
- stream-to-blob-url

jimmywarting:
IMHO i think videostream and render-media and all of that is out of the scope of webtorrent. So I think that should be removed. it's not necessary for sharing things p2p

But serving the files in a streaming fashion from the torrent client isn't and that is what createServer is doing.

WatchSoMuch:
Its great idea Jimmy, have you tested it for streaming?
Is there any sample or code example how to set it u?

jimmywarting:
1) You need to set up the service worker on [sw.js](https://github.com/jimmywarting/webtorrent-server-browser/blob/master/sw.js) your own domain that listens to any request made by the website

2) Since WebTorrent only runs in main thread it also has to have something to communicate too and you would also need to [patch the createServer stuff](https://github.com/jimmywarting/webtorrent-server-browser/blob/437fde40d1db27fbdeb82b601be950ce6b0da25c/index.html#L30-L183) so it works with service worker instead.

after that it's as simple as just adding

```js
video.src = `./webtorrent/${torrent.infoHash}/${torrent.files[5].path}`
```

> Is there any sample or code example how to set it u?

Did you even bother looking in the repo for demo and some code? the readme links to a demo that streams a video
[...]

PR #2098, opened 2021-06-04
feat: add service worker server as an alternative to renderMedia

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

Added a service worker server as an alternative to renderMedia and [somewhat] equivalent of node's `createServer()`, this idea was first suggested by @jimmywarting , this allows the user to play back any video container and or codec as long as its supported by the browser, meaning file support is browser dependent, it also allows the user to start saving a file to drive before a file is fully downloaded without any extra RAM overhead on the browser [previously getBlobURL was used to accomplish this] 

This is done by first running `loadWorker (registration, callback)` on the client, and later `getStreamUrl (callback)` or `streamTo (element, callback)` on a file

`loadWorker` accepts an already registered and activated ServiceWorker object

`lib/file-server.js` is created so that users can bundle their own service worker, and only import the required function to their existing worker

`File` has a new event emitter called `stream` which emits readStreams for users to pipe through for a example a parser, and then return the piped stream

This entire implementation should work on a single service worker, for multiple WebTorrent instances, on multiple different scopes

#1522 
This also addresses these issues:
- file was too big to play back
- video container was unsupported
- video codec was unsupported
- file was too big to save to drive

- current structure

**Other:**
- streamTo and getStreamURL is only temporary, ideally this should be an option for renderTo/appendTo?
- the user needs to host the service worker file themselves due to scope and origin registration
- if the user has any files hosted under a /webtorrent/ path, the sw will take it over and requesting that file will fail, however if the user makes a custom sw bundle they can easily work around this


ThaUnknown:
@jimmywarting @ftreesmilo @DiegoRBaquero feedback please.

lib/file-server.js
```diff
@@ -0,0 +1,67 @@
+self.addEventListener('fetch', evt => {
```
DiegoRBaquero, line 1: I don't like the file name, doesn't mention anything about service worker

lib/file-server.js
```diff
@@ -0,0 +1,67 @@
...
+  const { request } = evt
+  const { url, method, headers } = request
+  if (!url.includes(self.registration.scope + 'webtorrent/')) return null
+  function getConsumer (clients) {
+    return new Promise(resolve => {
+      // Use race condition for whoever controls the response stream
+      for (const client of clients) {
+        const mc = new MessageChannel()
```
DiegoRBaquero, line 9:
```suggestion
        const messageChannel = new MessageChannel()
```

lib/file-server.js
DiegoRBaquero, line 1:
```suggestion
self.addEventListener('fetch', event => {
```

lib/file-server.js
```diff
@@ -0,0 +1,67 @@
...
+        const { port1, port2 } = mc
+        port1.onmessage = evt => {
```
DiegoRBaquero, line 11:
```suggestion
        port1.onmessage = event => {
```

lib/file-server.js
```diff
@@ -0,0 +1,67 @@
...
+        }, [port2])
+      }
+    })
+  }
+
+  evt.respondWith(
+    clients.matchAll({ type: 'window', includeUncontrolled: true })
+      .then(getConsumer)
+      .then(([data, mc]) => {
+        let tm = null
```
DiegoRBaquero, line 28: tm?

lib/file-server.js
```diff
@@ -0,0 +1,67 @@
...
+          scope: self.registration.scope
+        }, [port2])
+      }
+    })
+  }
+
+  evt.respondWith(
+    clients.matchAll({ type: 'window', includeUncontrolled: true })
+      .then(getConsumer)
+      .then(([data, mc]) => {
```
DiegoRBaquero, line 27:
```suggestion
      .then(([data, messageChannel]) => {
```

lib/file.js
```diff
@@ -147,6 +150,63 @@ class File extends EventEmitter {
...
+    if (this._serviceWorker.state !== 'activated') throw new Error('Worker isn\'t activated')
+    const url = `${window.location.pathname.substr(0, window.location.pathname.lastIndexOf('/') + 1)}webtorrent/${this._torrent.infoHash}/${encodeURI(this.path)}`
+    cb(null, url)
+  }
+
+  streamTo (elem, cb) {
+    if (typeof window === 'undefined') throw new Error('browser-only method')
+    if (!this._serviceWorker) throw new Error('No worker registered')
+    if (this._serviceWorker.state !== 'activated') throw new Error('Worker isn\'t activated')
+    // TODO: fix this \/
```
DiegoRBaquero, line 205: Please go deeper in detail as what there is to fix

package.json
```diff
@@ -158,6 +160,7 @@
   "standard": {
     "ignore": [
       "webtorrent.min.js",
+      "worker.min.js",
```
DiegoRBaquero, line 163: shouldn't this be `sw.min.js` ?

lib/file.js
ThaUnknown: oh, when I was first writing it I didin't bother checking if there are any HTML elements which load their content in a way different than using `src` but then realized all do so that can be removed

lib/worker.js
```diff
@@ -0,0 +1,13 @@
+const fileResponse = require('./worker-server')
+
+self.addEventListener('install', () => {
+  self.skipWaiting()
+})
```
DiegoRBaquero, line 5:
```suggestion
})

```

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 package-lock.json
 webtorrent.debug.js
+sw.debug.js
```
DiegoRBaquero, line 4:
```suggestion
sw.debug.js

```

DiegoRBaquero:
The `self` would need to be fixed, or switched to `this`. As for globals, you can add them: https://standardjs.com/#i-use-a-library-that-pollutes-the-global-namespace-how-do-i-prevent-variable-is-not-defined-errors

lib/file.js
```diff
@@ -147,6 +150,62 @@ class File extends EventEmitter {
...
+        // Support range-requests
+        'Accept-Ranges': 'bytes',
+        'Content-Type': this._getMimeType()
+      }
+    }
+    // force the browser to download the file if there is a specific header present
+    if (req.destination !== 'iframe' && req.headers['upgrade-insecure-requests'] === '1') {
+      res.headers['Content-Type'] = 'application/octet-stream'
+      res.headers['Content-Disposition'] = 'attachment'
+    }
```
ftreesmilo: I don't understand this part.   Using a service worker, all the incoming requests should be for a secure page (localhost is a secure origin in chrome while developing).
And if they did want to pipe this into an iframe without downloading the file, I think they should be able to.

ThaUnknown: Yeah, I've already changed this, the upgrade-insecure-requests header was a hack really, to figure out of the user is requesting from a new tab, I've changed this to use destination to exclude iframes, but then realised there's a `document` destination I could just target instead, now you can display the content in an iframe, and open it in a new tab to download it

index.js
```diff
@@ -155,6 +156,65 @@ class WebTorrent extends EventEmitter {
...
+        // }
+        throw e
+      })
+    } else {
+      if (!(registration instanceof ServiceWorker)) throw new Error('Invalid worker registration')
+      if (registration.state !== 'activated') throw new Error('Worker isn\'t activated')
+    }
+
+    this.serviceWorker = registration instanceof ServiceWorker ? registration : navigator.serviceWorker.controller
+
```
ftreesmilo, line 188: I think the utility here is all in the parts below... and hiding it inside a fn to register a sw will prevent people who already have afunctioning sw from hooking the sw message logic.   I think a well-defined message protocol for page<-> sw would be a better approach with addins to both hook sw fetch events and page message events that people could register themselves.

wt could automatically look for sw parents and by default use the below implementation, or could be provided with an alternate compliant handler for devs who want to change behavior.
or... i've seen this be a common way to get a certain default config
`{ option: true | () => {}}`  true for default impl, fn to override.   not specified, doesn't do anything.

I don't think trying to register by default would create a problem...  you could invert it and take an impl if it's provided and not register if it's === false

I still do not think wt should be in the business of managing sws for webpages.

index.js
```diff
@@ -155,6 +156,65 @@ class WebTorrent extends EventEmitter {
...
+    navigator.serviceWorker.addEventListener('message', event => {
+      if (!event.data.url) return null
+      let [infoHash, ...filePath] = event.data.url.slice(event.data.url.indexOf(event.data.scope + 'webtorrent/') + 11 + event.data.scope.length).split('/')
+      filePath = decodeURI(filePath.join('/'))
+      if (!infoHash || !filePath) return null
+
+      const [port] = event.ports
+
+      const file = this.get(infoHash) && this.get(infoHash).files.find(file => file.path === filePath)
+      if (!file) return null
```
ftreesmilo, line 198: I think it would be nice to provide the page with some information about the request...    like a 404 response.

lib/file.js
```diff
@@ -33,6 +34,8 @@ class File extends EventEmitter {
       this.done = true
       this.emit('done')
     }
+
+    this._serviceWorker = torrent.client.serviceWorker
```
ftreesmilo, line 38: this should not be needed.  pass the data port to _serve

lib/file.js
```diff
@@ -147,6 +150,62 @@ class File extends EventEmitter {
     render.render(this, elem, opts, cb)
   }
 
+  _serve (req) {
```
ftreesmilo, line 153: I actually think all of this logic belongs with the onMessage impl, not in file.

index.js
ThaUnknown: yeah, the registering service workers part is kinda meh, but I do want to let users just run WT's function to register this service worker, so they consciously know what's happening with the service worker and what state it's in, I could just assume "hey this page has a service worker, lets trust it has the built-in WT functionality" or like you said create a message channel checking if said functionality exists, but I honestly was too lazy to do that, so I let users just pass a registered service worker as an option which bypasses the registration

lib/file.js
ThaUnknown: this isn't for the _serve functionality, but for checking if the worker is registered and activated when running `getStreamURL` or `streamTo`

index.js
ThaUnknown: general issue here, is that you can have multiple webtorrent instances running of off a single service worker, and only one wt instance can be the controller, so some instances might have the files, some might not, that's why I didn't do any error handling for this

lib/worker-server.js
```diff
@@ -0,0 +1,67 @@
...
+                }
+                // 'media player' does NOT signal a close on the stream and we cannot close it because it's locked to the reader,
+                // so we just empty it after 5s of inactivity, the browser will request another port anyways
+                clearTimeout(timeOut)
+                timeOut = setTimeout(() => {
+                  controller.close()
+                  messageChannel.port1.postMessage(false) // send timeout
+                  messageChannel.port1.onmessage = null
+                  resolve()
+                }, 5000)
```
ftreesmilo, line 52: magic number should probably be a const at top of file

lib/file.js
ftreesmilo: how would it get this far unless there was a request from a sw?

ThaUnknown: `getStreamURL` or `streamTo` is something like `renderTo` you can call it on any file at any point, and all it does is returns/sets the URL for the specific file, it could be that some idiot decides to run this without a service worker, or that he doesn't activate his service worker even tho its registered

index.js
ftreesmilo: Ok, I see that...  but that's a pretty complex use-case.    It's going to be up to the WT maintainers what they want, but if it were me, I'd vote for the simple impl.   The page requesting the url is the page that owns the wt instance.

logic in fetch handler could be cleaned up a little to message the requesting client.  Not loop over all of them.
I don't know how common a popout video player window would be.
I don't personally have that use case.   But that's where others could implement the interfaces we should define.

How much do we want to try to support in the default impl?  you'll notice that wt doesn't have many Store impls by default.

lib/file.js
ftreesmilo: oh...     yeah...     i dunno.    I think that breaks your desire to support multiple pages.   if a user has to use this method to populate the src of the video, then the wt instance is already in the page requesting the file

It might be more portable to other implementations if you made a new fn to getSwRelativeUrl() so the client can augment it with their chosen scope.  then again i'm also fine with assuming they used '/' for the chosen scope for the default impl

ThaUnknown: I completely forgot, you are right on this, it needs a path relative to the service worker registration scope, not the current scope, idk how I missed this

index.js
ThaUnknown: I changed this, and removed the service worker registration, and now only allow a sw object to be passed in, because I still want to force the user to consciously make the decision that they "activate" the sw functionality in wt, instead of automatically detecting when a service worker is registered, this entire thing could be worked around by implementing functionality which checks if the sw has some custom wt functionality it otherwise wouldn't have, so I'm gonna keep this open for now

lib/file.js
ThaUnknown: I think this is the right place for it, but it doesnt seem right either?... it's really up to the maintainers to decide on this

index.js
ThaUnknown: IMO the user should be in the control of what SW webtorrent uses, especially if there are multiple, so I wont do auto-detection

ThaUnknown:
@jhiesey I ended up fixing the file downloading issue completely, at a small cost of RAM, TLDR in theory this solution can leak a bit of RAM on some browsers if: the user cancels the download, and it doesn't send the download cancel, but it's such a laughably small amount, that it will most likely get GC'd by the time it becomes an issue, so it should be fine as long as the user doesn't cancel a few hundred downloads in a few seconds

jimmywarting:
There are two alternative methods to create a `evt.respondWith(new Response(new ReadableStream(...)))` that is way better than keep sending chunks with postMessages and is much faster than how i originally did in my PoC that have involved.

- First of all, it would be faster to transfer the chunks using the [transferable list](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) instead of copying the data over to the worker. But there are other methods also.
- Chrome supports [transferable streams](https://chromestatus.com/feature/5298733486964736). This means that you can construct a ReadableStream in the main thread and transfer it to the service worker. The service worker would not be responsible for creating the ReadableStream pull stream and be responsible for enqueuing data. This makes the service worker a tiny bit pointless so it makes it possible for the service worker to go idle, b/c you would enqueue the chunks from the main thread instead. With this solution you wouldn't need to ping the service worker to keep it alive. cuz the service worker would need to do anything. https://github.com/w3c/ServiceWorker/issues/882#issuecomment-262754779
- But not all browser supports transferable streams, But there is another way you can create a pipe stream that also don't require the service worker to be responsible for generating the output. and that is to simply echo the `request.body` back to a 2nd request and from the main thread create a streamable request that pipes right back at you
```js
// main thread
fetch('./echo-stream-back-to-next-request', {
  headers: {
    'x-url': `/webtorrent/${infohash}/sample.zip`
    'content-type': `application/zip`,
    'content-size': `123`,
    // accept partial byte request, bla bla bla
    ...
  },
  method: 'POST',
  body: new ReadableStream({
    pull(ctrl) {
      ctrl.enqueue(new Uint8Array([97]))
      ctrl.close()
    }
  })
})

// service worker
const map = new Map()
onfetch = evt => {
  if (evt.request.url === './echo-stream-back-to-next-request') {
    map.set(evt.request.headers.get('x-url'), evt.request)
    evt.respondWith(new Response('ok'))
  }
  if (map.has(evt.request.url)) {
    const req = map.get(evt.request.url)
    evt.respondWith(new Response(req.body), { headers: req.headers })
  }
}
```

ftreesmilo:
@jimmywarting  How on earth is your #3 better than postmessage with chunk transfer?

jimmywarting:
> How on earth is your 3rd suggestion better than postmessage with chunk transfer

The Service Worker don't need to do any work so it can go idle, hence why you don't need any keep alive hack with postMessages either. It's still just only one ReadableStream, It's almost just exactly like a Transferable ReadableStream. 
The ReadableStream you create in the main thread is directly piped to a `<video>` element

the code example is also just a rough sketch of how you would go about solving it. maybe you won't solve it exactly like i did, i only wanted to point out that you can create a `new Request(url, { body: new ReadableStream(...)})` as a way of transfering a ReadableStream in a round about way if Transferable Streams are not supported.

jhiesey:
@ftreesmilo @jimmywarting sorry I've been busy. I promise I'll take a look at this tomorrow.

ThaUnknown:
> > How on earth is your 3rd suggestion better than postmessage with chunk transfer
> 
> The Service Worker don't need to do any work so it can go idle, hence why you don't need any keep alive hack with postMessages either. It's still just only one ReadableStream, It's almost just exactly like a Transferable ReadableStream.
> The ReadableStream you create in the main thread can go directly to the `<video>` element
> It's just a other way of transferring a ReadableStream if 2nd option is not possible
> 
> the code example is also just a rough sketch of how you would go about solving it. maybe you won't solve it exactly like i did, i only wanted to point out that you can create a `new Request(url, { body: new ReadableStream(...)})` as a way of transfering a ReadableStream in a round about way if Transferable Streams are not supported.
> 
> All doe it's still a compatibility issue with this 2nd & 3rd way of doing it.
> You can ofc do `new Request(url, { body: blob })` also...

ngl, this is getting complex as fuck, but how or when would you know to remove the "cached" request body from the worker? 

from what I understand, before the video element makes a request you want to create and cache a readable stream request, but you can't just delete it on the first fulfilment because the video element might make more requests, or the user might want to load the same content multiple times? or am I misunderstanding something? is there a way to modify the incoming request from the video? or do you want to create a new request every time the video makes a new request and tell the main thread about that with post message?

jimmywarting:
It's not possible to reuse an already consumed ReadableStream. so you can't reuse a cached request over again that is why why i deleted it directly after. i guess the workflow would have to be something in terms of: 

Service worker wants to fullfill a request with `evt.respondWith(new Response(readableStream))` but the service worker don't have readableStream at hand, it would kind of have to broadcast back to WebTorrent (with postMessage) that it needs a readableStream with this (optional partial range) request with file x. WebTorrent makes a streaming post request back to the service worker so service worker can resolve the readableStream

ThaUnknown:
> It's not possible to reuse an already consumed ReadableStream. so you can't reuse a cached request over again that is why why i deleted it directly after. i guess the workflow would have to be something in terms of:
> 
> Service worker wants to fullfill a request with `evt.respondWith(new Response(readableStream))` but the service worker don't have readableStream at hand, it would kind of have to broadcast back to WebTorrent (with postMessage) that it needs a readableStream with this (optional partial range) request with file x. WebTorrent makes a streaming post request back to the service worker so service worker can resolve the readableStream

hmmm, I personally don't think all this is worth 2MB of RAM unless there are some major CPU usage reductions to be had here, this would however work great if webtorrent used readable streams for file readStreams, since you'd just reference that directly

what I'm saying is, is this worth it with the current implementation of wt's streams?

jimmywarting:
> what I'm saying is, is this worth it with the current implementation of wt's streams?

All I'm saying is: your impl looks fine as it's (it probably works grate too - I haven't tested it). It probably has more compatibility with older browsers since streaming upload with fetch+request+readableStream came later.
I'm just giving you an alternative solution that don't involves constructing a ReadableStream on the service worker that allows the service worker to go idle. where a keepAlive hack is necessary by pinging the sw ever x second to prevent it from being terminated

ThaUnknown:
> All I'm saying is: your impl looks fine as it's (it probably works grate too - I haven't tested it). 

it gets very CPU hungry at a few hundred MB/s, but I'm not sure if it's because of the IPC or because of the buffer stuff going on in wt

ThaUnknown:
> > It's not possible to reuse an already consumed ReadableStream. so you can't reuse a cached request over again that is why why i deleted it directly after. i guess the workflow would have to be something in terms of:
> > Service worker wants to fullfill a request with `evt.respondWith(new Response(readableStream))` but the service worker don't have readableStream at hand, it would kind of have to broadcast back to WebTorrent (with postMessage) that it needs a readableStream with this (optional partial range) request with file x. WebTorrent makes a streaming post request back to the service worker so service worker can resolve the readableStream
> 
> hmmm, I personally don't think all this is worth 2MB of RAM unless there are some major CPU usage reductions to be had here, this would however work great if webtorrent used readable streams for file readStreams, since you'd just reference that directly
> 
> what I'm saying is, is this worth it with the current implementation of wt's streams?
> 
> edit: it just clicked with me, I get the benefits of this now, I'll see if I can create a possible implementation for this since much less IPC is happening with this idea, and then we'd have a ready placeholder for once we uses readable streams, this idea doesn't seem to reduce any browser compatibility either, at least on the surface

turns out the support for this is kind-of iffy, but it was a nice idea :/

WatchSoMuch:
Sorry if this question is stupid, I'm not a Github pro, is this usable using last release? and how?

DiegoRBaquero (approved):

ThaUnknown:
>  /home/runner/work/webtorrent/webtorrent/index.js:85:1: More than 1 blank line not allowed.
thanks github auto-merge

index.js
```diff
@@ -157,6 +162,68 @@ class WebTorrent extends EventEmitter {
...
+          if (!chunk) cleanup()
+          if (!this.workerKeepAliveInterval) this.workerKeepAliveInterval = setInterval(() => fetch(`${this.serviceWorker.scriptURL.substr(0, this.serviceWorker.scriptURL.lastIndexOf('/') + 1).slice(window.location.origin.length)}webtorrent/keepalive/`), keepAliveTime)
+        } else {
+          cleanup()
+        }
+      }
+      this.workerPortCount++
+      port.postMessage(response)
+    })
+    cb(this.serviceWorker)
```
ftreesmilo, line 224: shouldn't this be 
```js
cb(null, this.serviceWorker)
```

ThaUnknown: ah yes, by the node callback standard it should, this originally was a promise, which i had to un-promisify to keep inline with the current implementations of webtorrent

a few things changed since i made this, but i havent had time to fix them, for example chrome finally cancels streams once they are out of use by consumers such as media players or image containers, which it didn't before. this was such a niche issue that i didnt even consider it ever being fixed and now it spams errors

additionally i think this needs to be made inline with the http server used by individual files, which i believe should work something like webseeds where you'd create the server on the client itself, rather than individual file/torrent

PR #2103, opened 2021-06-13
refactor: trigger _onWireWithMetadata after selections are updated

alxhotel:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

This PR:
- Changes the moment when `_onWireWithMetadata` is triggered.

The aim of this PR is to reduce the number of requests.

When the client gets the metadata, it tries to connect to other peers that it already has, as shown here:

https://github.com/webtorrent/webtorrent/blob/2335139b7529e6dd19cda01c0e4b3cbfafaeaf93/lib/torrent.js#L525-L530

This forces the client to start requesting pieces to peers before checking if those pieces are necessary.

This PR delays the call of `_onWireWithMetadata` until the `_selections` field is updated. Therefore, the client requests only the necessary pieces.

Could this break something?


lib/torrent.js
```diff
@@ -1090,7 +1091,7 @@ class Torrent extends EventEmitter {
     // More info: https://github.com/webtorrent/bittorrent-protocol#extension-api
     this.emit('wire', wire, addr)
 
-    if (this.metadata) {
+    if (this.ready) {
```
DiegoRBaquero, line 1094: What's the reasoning behind this change?

DiegoRBaquero: @alxhotel

DiegoRBaquero: @alxhotel

DiegoRBaquero (approved):

PR #2113, opened 2021-06-25
Cleanup duplicated deselect() code

feross:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

The `fileStream._destroy()` function would never run because when `fileStream.destroy()` is called, it sets a `destroyed` getter to `true` (inside the implementation of `readable-stream` which we subclass from.) This getter shadows our own `fileStream.destroyed` property.

The code outside running `eos()` was reaching into `fileStream` internals, so I just moved it in there.

None

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -1,5 +1,6 @@
 const debug = require('debug')('webtorrent:file-stream')
 const stream = require('readable-stream')
+const eos = require('end-of-stream')
 
 /**
  * Readable stream of a torrent file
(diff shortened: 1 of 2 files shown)


alxhotel (approved):

nuzzesick (approved):

DiegoRBaquero (approved):

jhiesey (approved):
LGTM; it's good to clean this up!

Issue #164, opened 2014-10-25
file.deselect supposedly not working


feross:
Heard from someone on IRC that after calling `file.deselect` on a file, it still gets fully downloaded.

Wingman4l7:
The problem is in [the conditional in](https://github.com/feross/webtorrent/blob/513e6fa11de121d083da60ed3cb959dad2257b23/lib/storage.js#L244) `torrent.deselect()`, which needs to be rewritten.  Alternatively, you can possibly rewrite the code that handles the default selection & storage of the entire piece range in the torrent when `client.add()` is called.

If you put a debug line inside the conditional and then attempt to download only particular files in a torrent via deselecting all of the files which you do **not** want, you will never see it -- because the conditional is never met!

The issue appears to be that the conditional is trying to match the piece range of the file you wish to deselect, against the stored piece range.  However, it will never find a match, as the only stored range is the piece range of the torrent in its entirety.

A way to demonstrate how this is broken *(and a hacky fix to get the effective functionality of `deselect()` in the meantime)* is to deselect the entire piece range, then **select** the files that you want.  

You also might want to change the readme section for `torrent.deselect()`.  Right now it only mentions that it can "deprioritize a range".  Perhaps say something along the lines of "Use a non-zero value for `priority` -- zero means that the pieces will **not** be downloaded."

I did notice that unfortunately, you will get "false" copies of the files on either side of the piece range you selected, due to the nature of files not beginning or ending exactly where pieces do.  This should probably be opened as a separate bug unless you can easily add in code for cleanup of those / prevent the creation of those false-files during this bugfix. 

Credit goes to [AlliedEnvy](https://github.com/AlliedEnvy) for figuring this out.

dcposch:
Yeah, the `deselect()` API is janky.

You have to deselect the whole torrent first, then select individual files.

Here's how we made it work in WebTorrent Desktop:

```
  // Remove default selection (whole torrent)
  torrent.deselect(0, torrent.pieces.length - 1, false)

  // Add selections (individual files)
  for (let i = 0; i < selections.length; i++) {
    const file = torrent.files[i]
    if (selections[i]) {
      file.select()
    } else {
      console.log('deselecting file ' + i + ' of torrent ' + torrent.name)
      file.deselect()
    }
  }
```

@feross we should probably improve this API before WebTorrent 1.0, because afterward we don't want to change it.

[...]

PR #2115, opened 2021-06-28
feat: change `deselect` and `select` behaviour

alxhotel:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

This PR:
- Changes the data structure of `_selections`. It uses a non-overlapping interval list
- Separates the notification callbacks of each selection in a new data structure

Tries to tackle the problem with the `deselect` and `select` methods #164

Let me know if this approach makes sense for what we need
 
**NOTE**: I'm not sure if this really a fix instead of a breaking change


feross (approved):
Alex and I discussed this PR on a video call. We decided that the right way forward is:

Make a (technically) breaking change to the way that the user-facing torrent.select(), torrent.deselect(), file.select(), and file.deselect() functions work. Right now they work like a reference counter, where you need to call deselect() for every select(), similar to the way you use malloc() and free().

For example, if you call `file.select(); file.select()`, you need to call `file.deselect(); file.deselect()` in order to undo it.

This PR changes these functions to use non-overlapping intervals, and to make it so that for example, `file.select(); file.select()` is the same as a single `file.select()`, since the ranges overlap. So you can do `file.select(); file.select(); file.deselect()`.

However, we need to change this PR so that the selections created by calling `file.createReadStream()` are not affected by the user-facing methods. We need the old behavior that acts like a reference counter for this method, otherwise this would break:

```js
stream1 = file.createReadStream() // selects the whole file
stream2 = file.createReadStream() // no-op since the whole file is already selected
stream1.destroy() // removes the selections, now nothing is selected

// now stream2 will stall since there are no selections!
```

But if we keep the old behavior for streams, then the selections can overlap and be duplicated to support the separate streams.

Finally, we're going to add a new option (e.g. `deselect: true`) to the torrent constructor which starts the torrent out deselected by default, so then the user can do:

```js
torrent = client.add(torrentId, { deselect: true })
torrent.on('ready', () => {
  torrent.files[0].select() // select just the file that you want, the rest are deselected already
})
```

Stefanuk12:
> Finally, we're going to add a new option (e.g. `deselect: true`) to the torrent constructor which starts the torrent out deselected by default, so then the user can do:
> 
> ```js
> torrent = client.add(torrentId, { deselect: true })
> torrent.on('ready', () => {
>   torrent.files[0].select() // select just the file that you want, the rest are deselected already
> })
> ```

Any updates on this yet? Hasn't been added yet as far as I can see.

ThaUnknown:
#2757

PR #2121, opened 2021-06-29
fix: add infoHash to path, specify store path, align with docs

ThaUnknown:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

1. added infoHash to the torrent's path
2. added a new option passed to the store `path` which specifies where the store folder itself should be created
3. changed files passed to the store, to use wt's file objects
4. passed the torrent object to the store

This PR heavily overlaps with #2090 but for different reasons:
1. previously the user needed to specify the unique path to the torrent themselves, however 9/10 times the user doesn't know the torrent's infoHash so they can't create a unique path for the torrent, this is even an issue on webtorrent-desktop, where if you create 2 torrents with files of the same name, webtorrent will overwrite the first file with the second, this fixes the issue

2. previously webtorrent included folders in it's paths, and created said folders, but didin't specify in any way if the folders were 
created by webtorrent or not, so when you went to delete the store, it couldn't know what folders it should delete, and left empty folders behind, this now specifies the path of the store's folder

3. allows more control and information, additionally this changed to use file paths relative to the torrent path, not absolute paths

4. previously the torrent object was only it's infoHash which was the same as the name, this was changed in order to align with the documentation

[WIP] in fs-chunk-store in order to support these changes, this PR won't work without them


lib/torrent.js
```diff
@@ -476,23 +476,18 @@ class Torrent extends EventEmitter {
...
 
+    this.files = this.files.map(file => new File(this, file))
+
     this.store = new ImmediateChunkStore(
       this._preloadedStore || new this._store(this.pieceLength, {
-        torrent: {
-          infoHash: this.infoHash
-        },
-        files: this.files.map(file => ({
-          path: path.join(this.path, file.path),
```
DiegoRBaquero, line 485: isn't this lost?

ThaUnknown: the file object includes it's [relative path](https://github.com/webtorrent/webtorrent/blob/master/lib/file.js#L20), torrent object includes its infohash too

ftreesmilo:
I kept the paths joined for the files because everything that used file.path used it joined on to the torrent path.
If you pass the torrent and files to the store, you can get relative paths easily by using node's path module.

I think this pr has value in the storeopts, perhaps it should just be that, and both can be delivered together

lib/torrent.js
```diff
@@ -229,7 +230,7 @@ class Torrent extends EventEmitter {
       return this._destroy(new Error('Malformed torrent data: No info hash'))
     }
 
-    if (!this.path) this.path = path.join(TMP, this.infoHash)
+    if (!this.path) this.path = TMP
```
DiegoRBaquero, line 233: I don't understand this change, why remove it, where is this being re added? in the store?

ThaUnknown: the infoHash was only added to the path if the user didn't define his own path, and the infoHash was already passed in the store's name which I ended up using [here](https://github.com/feross/fs-chunk-store/pull/17/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R45)

lib/torrent.js
```diff
@@ -473,19 +474,15 @@ class Torrent extends EventEmitter {
...
-        },
-        files: this.files.map(file => ({
-          path: path.join(this.path, file.path),
-          length: file.length,
-          offset: file.offset
-        })),
+        ...this.storeOpts,
+        torrent: this,
+        path: this.path,
+        files: this.files,
```
DiegoRBaquero, line 485: does the store need the whole `File` instance reference?

ThaUnknown: it does simply things nicely, and there really isn't a reason not to do it, additionally custom webtorrent specific stores can for example listen to the 'done' event to save a full file to drive, or close an opened stream to the file

jhiesey: The current change, as discussed here https://github.com/feross/fs-chunk-store/pull/17#discussion_r671794460 requires that the infoHash is in the path name.

I realize this is useful to prevent collisions, but is probably not what casual users want, and if we change it that would have to be in a major version of webtorrent (2.0).

I think we need to think through the logic a bit more carefully. My current rough idea:
* Don't change the store at all, at least to solve this specific issue
* Add a new param to the torrent opts: `basePath`, which you set instead of `path` if you want the infoHash appended to avoid conflicts. If you set the normal `torrentOpts.path` option instead the behavior won't change.
* In `_onParsedTorrent`, do something like this:
```js
if (!this.path) {
  this.path = path.join(this.basePath || TMP, this.infoHash)
  this.cleanupPath = this.path
}
```

* In `torrent._destroy`, if `torrent.cleanupPath` is set, recursively delete it here in webtorrent, and set `torrent._destroyStoreOnDestroy` to false to disable the store's cleanup logic.

jhiesey (changes requested):
As is, this change alters the location files are saved in a way that can't easily be overridden. I agree having the `infoHash` in the path is very useful, but not everyone wants that.

There should be a new opt to get the new behavior instead of changing existing behavior in a way that will likely break projects that use webtorrent.

lib/torrent.js
ThaUnknown: I really don't think that webtorrent should have any involvement in the actual deletion of files tho

ThaUnknown:
how about this? also removed the name concat from fs-chunk-store

ThaUnknown:
@jhiesey as to the last meet, I suggested to use multiple chunk stores, one named, and one not, decided that isn't the best idea due to bundle size and did this instead, added an option for adding the UID, but instead of it specifying the store like I suggested, it passes that option to the store, this way if more filesystem-like stores come along, they can just use that specific non-standard option

Issue #2114, opened 2021-06-26
Error: invalid argument when connecting to uTP peers via IPv6


ThaUnknown:
**What version of this package are you using?**
v1.0.2
**What operating system, Node.js, and npm version?**
Windows 10 LTSC 1809, electron 13.1.2, node 14.16.0
**What happened?**
Adding/connecting to a peer which uses IPv6 string throws an error:
```
path\node_modules\utp-native\lib\connection.js:197
  binding.utp_napi_connect(this._utp._handle, this._handle, port, ip)
          ^

Error: invalid argument
    at Connection._connect (path\node_modules\utp-native\lib\connection.js:197:11)
    at path\node_modules\utp-native\lib\connection.js:189:10
    at processTicksAndRejections (internal/process/task_queues.js:81:21) {
  code: 'EINVAL'
  ```
code used:
```js
      const peers = [{ type: 'utpOutgoing', addr: '[2001:5b0:4ec0:c848:4985:2d28:bde4:40e2]:39214' }]
      for (const peer of peers) {
        if (peer.type === 'webSeed') {
          this.torrent.addWebSeed(peer.addr)
        } else if (peer.type !== 'webrtc') {
          this.torrent.addPeer(peer.addr)
        }
      }
```
this also occurs when naturally connecting to uTP peers via trackers
this doesnt occur for webSeeds/tcp peers

jczasgit:
Having the same issue midway download. Works fine on my windows 10 local machine, but in ubuntu 18.04 throws the error.

```
Error: invalid argument 
    at Connection._connect (/mnt/sdb/backends/synchronos/node_modules/utp-native/lib/connection.js:197:11)
    at /mnt/sdb/backends/synchronos/node_modules/utp-native/lib/connection.js:189:10 at processTicksAndRejections (node:internal/process/task_queues:82:21) 
    at runNextTicks (node:internal/process/task_queues:63:3) 
    at listOnTimeout (node:internal/timers:526:9) at processTimers (node:internal/timers:500:7) { code: 'EINVAL' }
```

code:
```typescript
this.client.add(this._torrentSource, (_torrent) => {
        // deselect all unnessesary files.
        _torrent.files.forEach((file) => {
          file.deselect();
        });
        _torrent.deselect(0, _torrent.pieces.length - 1, 0);

        // get only mp4 file
        let file = _torrent.files.find((file) => file.name.endsWith(".mp4"));
        file.select();
[...]

PR #2125, opened 2021-07-02
fix: ensure uTP peer address is IPv4

hicom150:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Check peer address to ensure it is a valid IPv4 when using uTP protocol.

It should fix https://github.com/webtorrent/webtorrent/issues/2114

I've added an internal `_isIPv4` method, just in case we want to replace its implementation with something different than https://github.com/sindresorhus/is-ip

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -762,8 +762,9 @@ class Torrent extends EventEmitter {
     if (this.destroyed) throw new Error('torrent is destroyed')
     if (!this.infoHash) throw new Error('addPeer() must not be called before the `infoHash` event')
 
+    let host
+
     if (this.client.blocked) {
-      let host
       if (typeof peer === 'string') {
         let parts
         try {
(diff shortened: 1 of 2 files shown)


ThaUnknown:
why not use the same detection as utp-native?
https://github.com/mafintosh/utp-native/blob/master/index.js#L9

DiegoRBaquero:
I agree, no need for the extra dep

hicom150:
You are right about reusing https://github.com/mafintosh/utp-native code, but given that https://github.com/mafintosh/utp-native/blob/master/index.js#L9 nor https://github.com/mafintosh/utp-native/blob/master/index.js#L274 are exposed, I've fixed that by just copying the regex 😅

If in the future https://github.com/mafintosh/utp-native exposes this methods, we can easily refactor to use them 😉

ThaUnknown:
> You are right about reusing https://github.com/mafintosh/utp-native code, but given that https://github.com/mafintosh/utp-native/blob/master/index.js#L9 nor https://github.com/mafintosh/utp-native/blob/master/index.js#L274 are exposed, I've fixed that by just copying the regex 😅
> 
> If in the future https://github.com/mafintosh/utp-native exposes this methods, we can easily refactor to use them 😉

hopefully in the future they support ipv6 instead :p

IdanAdar:
@DiegoRBaquero Anything else needed before approving?

DiegoRBaquero (approved):

DiegoRBaquero:
My bad, I was on vacation

PR #2135, opened 2021-07-13
fix: call public FileStream.destroy method so cb is defined

jhiesey:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Previously, early exits in FileStream._notify would
throw since cb was not defined in _destroy

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -59,7 +59,7 @@ class FileStream extends stream.Readable {
     if (this._notifying) return
     this._notifying = true
 
-    if (this._torrent.destroyed) return this._destroy(new Error('Torrent removed'))
+    if (this._torrent.destroyed) return this.destroy(new Error('Torrent removed'))
 
     const p = this._piece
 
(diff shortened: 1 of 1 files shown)


DiegoRBaquero:
TIL: [`The _destroy() method is called by writable.destroy(). It can be overridden by child classes but it must not be called directly.` ](https://nodejs.org/api/stream.html#stream_writable_destroy_err_callback)

DiegoRBaquero (approved):

feross:
@DiegoRBaquero Yes, this is confusing. Streams have changed a lot over time. They used to never call `_destroy` but then that changed at some point. When node calls it, it passes a callback, but for backwards compatibility it doesn't require that you actually call it.

A lot of old code implemented a custom `_destroy` or `destroy` method that did not receive a callback and would manually emit `'close'` at the end.

To update to the new style, you need to:

- Accept a `cb` argument in `_destroy`
- Stop calling `_destroy` directly
- Delete the custom `destroy` method since it shadows the one coming from the new streams implementation
- Stop emitting a `'close'`event in `_destroy` since that's handled automatically when you call the `cb` now.

I made some of these other fixes in this PR: https://github.com/webtorrent/webtorrent/pull/2113

feross (approved):
LGTM, thanks John!

Issue #2110, opened 2021-06-23
Fix and improve the `noPeers` event


alxhotel:
**What version of this package are you using?**
1.0.1

**What operating system, Node.js, and npm version?**
Ubuntu 16.04, NodeJS 14, NPM 7

**What happened?** and  **What did you expect to happen?**
I have been going through the `noPeers` event issue (https://github.com/webtorrent/webtorrent/issues/1622) and I think that there are some problems with it beside the one exposed in the Github issue. These are my thoughts:
- The event doesn't actually tell you that there are no peers, it only says that the client hasn't been able to have a successful handshake with a peer.
- Not only that, but since it is checking the `numPeers` field, this does not differentiate between the discovery mechanism used to find the peer, so the event just says that the client hasn't been able to connect to a peer, not that it hasn't found a peer through that mechanism.
- LSD is missing an event to trigger the `noPeers` event. This means that (contrary to what it says in the docs) not finding a peer with LSD will not trigger the `noPeers` event.
- Having the event for the tracker get triggered "on each announce received" also doesn't seem to make sense, unless you either differentiate between trackers or you set a time interval to wait for all trackers to answers inside each time period
- Having this event being triggered "on each announce received" for LSD and DHT doesn't seem to make sense either, since receiving an announce of a peer will mean that the peer is interested in the torrent (unless I misunderstood the spec). Thus, the noPeers event will never get triggered.
- Also what about Peer Exchange (PEX)? Should we include it?

**Are you willing to submit a pull request to fix this bug?**
So I guess my solution will be to store the discovery mechanism used in each Peer object. And then use a global time interval to check that we have found or not a peer with that mechanism.

I'll open a PR with this solution mentioned.

PR #2142, opened 2021-07-21
refactor: change behaviour of `noPeers` event

### LINKED ISSUE #1622: What is the noPeers event expected behavior?
**What version of WebTorrent?**
@latest
**What operating system and Node.js version?**
@latest 

Reproductible snippet : 
```JAVASCRIPT
const wt = require('webtorrent');
var sintel = 'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F';
var client = new wt();
var tor = client.add(sintel);
tor.on('noPeers', a => console.log('No '+a+' peers'));
```

Since the 'Sintel' torrent is a very healthy torrent with many active trackers and peers sharing it, I don't understand what would be the use of 'noPeers' if it's called on every announce with no peers. 

Shouldn't the event only be called on edge cases when there are no active peers to continue download of the file, or on a very 'unhealthy' torrent with no peers at all to fetch metadata? This behavior makes the event almost useless in every case, unless I'm understanding it wrong. Thanks!

alxhotel:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

This PR:
- Changes the behaviour of the `noPeers` event
- Adds a new option to the client constructor to disable `ut_pex` discovery mechanism
- Refactor `Peer` constants

Closes #1622 #2110

**Note**: This is a breaking change so it should be merged to v2 branch.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -97,7 +97,7 @@ class Torrent extends EventEmitter {
     this._rechokeOptimisticTime = 0
     this._rechokeIntervalId = null
     this._noPeersIntervalId = null
-    this._noPeersIntervalTime = opts.noPeersIntervalTime || DEFAULT_NO_PEERS_INTERVAL
+    this._noPeersIntervalTime = opts.noPeersIntervalTime ? opts.noPeersIntervalTime * 1000 : DEFAULT_NO_PEERS_INTERVAL
 
     this.ready = false
     this.destroyed = false
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -38,22 +38,22 @@ const utp = require('./utp') // browser exclude
 const WebConn = require('./webconn')
 
 const MAX_BLOCK_LENGTH = 128 * 1024
-const PIECE_TIMEOUT = 30000
-const CHOKE_TIMEOUT = 5000
+const PIECE_TIMEOUT = 30_000
+const CHOKE_TIMEOUT = 5_000
 const SPEED_THRESHOLD = 3 * Piece.BLOCK_LENGTH
 
 const PIPELINE_MIN_DURATION = 0.5
 const PIPELINE_MAX_DURATION = 1
 
-const RECHOKE_INTERVAL = 10000 // 10 seconds
+const RECHOKE_INTERVAL = 10_000 // 10 seconds
 const RECHOKE_OPTIMISTIC_DURATION = 2 // 30 seconds
 
-const NO_PEERS_INTERVAL = 30_000 // 30 seconds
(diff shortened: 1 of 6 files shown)


lib/torrent.js
```diff
@@ -48,6 +48,8 @@ const PIPELINE_MAX_DURATION = 1
 const RECHOKE_INTERVAL = 10000 // 10 seconds
 const RECHOKE_OPTIMISTIC_DURATION = 2 // 30 seconds
 
+const NO_PEERS_INTERVAL = 30_000 // 30 seconds
```
DiegoRBaquero, line 51: shouldn't this be an opt in the client? defaulted to 30 seconds

DiegoRBaquero (approved):
I don't think this needs to be a breaking change major version, at least one event is being emitted now

lib/torrent.js
```diff
@@ -94,6 +96,8 @@ class Torrent extends EventEmitter {
     this._rechokeOptimisticWire = null
     this._rechokeOptimisticTime = 0
     this._rechokeIntervalId = null
+    this._noPeersIntervalId = null
+    this._noPeersIntervalTime = opts.noPeersIntervalTime || DEFAULT_NO_PEERS_INTERVAL
```
DiegoRBaquero, line 100:
```suggestion
    this._noPeersIntervalTime = opts.noPeersIntervalTime ? opts.noPeersIntervalTime * 1000 : DEFAULT_NO_PEERS_INTERVAL
```

DiegoRBaquero (changes requested):
opts states seconds, so we must multiply

lib/torrent.js
alxhotel: Oh good catch!

DiegoRBaquero (approved):

ThaUnknown:
why do incoming peers not have a source?

PR #2187, opened 2021-09-04
heroku deploy

sathishsksk:
heroku.yml


DiegoRBaquero:
I don't really understand why GH actions token is being changed, and app.json is not referring to this repo at all

Issue #86, opened 2014-09-12
Implement storage in browser clients


feross:
Currently, when webtorrent is used in the browser, it stores files entirely in memory. We need a solution that uses disk whenever possible, to prevent excessive memory usage and tab crashes.
### Possible approaches:
- **localStorage** (possibly with the http://filldisk.com workaround that I developed)
- **Filesystem API** (chrome only, but allows for temporarily using 20% of user's available disk space without a prompt)
- **IndexedDB** (horrible API, but there are shims. unsure what the UX prompt / limits are). Possible shims to use:
  - https://github.com/mozilla/localforage
  - https://github.com/js-platform/filer
  - https://www.npmjs.org/package/level-js

Are there other ways to do browser storage? Are there other approaches to consider? Are there better storage modules to use?

Feedback welcome!

yciabaud:
I am doing some experiments to persist data in the browser and reduce memory consumption. I wrote a chunk store based on

santiagogil:
## Managed to get fs-chunk-store working in the browser over browserify-fs instead of fs.
I had to modify:
* node-mkdirp (remove sync methods and make it use browserify-fs instead of fs).
* rimraf (remove sync methods and make it use browserify-fs instead of fs).
* random-access-file (make it use browserify-fs instead of fs).
* fs-chunk-store itself (avoid using os in browser).

The only remaining problem is the fact that fs-chunk-store needs the temp path to exist in order to run.
Right now if you run my version in the browser it fails until the page is reloaded and the temp path gets created.

### Some questions:
* Which would be the best way to solve the temp path problem?
* Is it preferable to make PRs to each of the dependencies making them browserifyables or just fork them in order to get this working as soon as possible?
* Is the chunk store the only change that needs to be done?

There are some other fs replacements for the browser (some with sync methods included) but those look too big and complex.

santiagogil:
Got fs-chunk-store working without reloading.

Even with this, it's still impossible to build webtorrent. 
It seems that something I can't find yet it's trying to open a directory that doesn't exist.

Going to work on it tomorrow.
Still think that it's viable to solve the browser storage problem using a leveldb version of fs-chunk-store.

Cheedoong:
I discussed this with PLH (http://www.w3.org/People/LeHegaret/) from W3C and he strongly suggested using IndexedDB in this scenario.

yciabaud:
[...]

PR #2208, opened 2021-10-25
feat: use storage (FSA+IDB) instead of memory in browser

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

Added an option to use drive storage instead of memory in browser.

Replaced memory chunk store with hybrid chunk store which will ***try*** to figure out how much space the user has on their drive, use all of it - 16MB, then fallback to memory [should this be blobs instead of memory?] for the rest, this increases the capacity on firefox and safari by 2^31 or 2.14GB, and on chrome by as much as the user has space.

Additionally on *most* chromium versions the user can specify a rootDir, which allows the user to specify a custom target download folder, which replicates the folder and file structure in the same way as node does. 

Just like node, if the browser crashes the files won't be removed, this is something the developer should handle.

**FSA specific:**

Because of the limitations of the FSA API, you cant write and save from the same file at once, [this is meant to improve](https://bugs.chromium.org/p/chromium/issues/detail?id=1323922) but there's no ETA, and [in-place read/writes are only possible in web workers](https://web.dev/file-system-access/#accessing-files-optimized-for-performance-from-the-origin-private-file-system). To work around this the FSA chunk store works in 2 modes:
- [OPFS](https://web.dev/file-system-access/#accessing-the-origin-private-file-system) mode: if no custom directory is defined the built in default storage location is used, where each chunk has it's own file, and each write opens and closes a file.
- custom directory mode: files are written to the custom directory, but individual chunks are also stored in OPFS, if a chunk exist in OPFS it's read from there, otherwise it's read from the respective files from the custom directories, this means x2 data usage, however the excess data is cleared on store close. If the browser closes too fast, and it doesn't manage to remove the store in time [ex: multi-gigabyte torrent], on the next load of the page, the excess data will be removed.

The stores are ones made by wormhole, just the FSA one is expanded to allow for extra functionality.

- https://github.com/orgs/webtorrent/projects/1#card-60702730
- https://github.com/webtorrent/webtorrent/issues/86
- https://github.com/webtorrent/webtorrent/issues/2389

the new package, 
if the store is trying to do too many things at once, 
the increasing amt of store opts and how to possibly reduce this?,


ThaUnknown:
@DiegoRBaquero I've removed the github dependencies, and pulled latest, you can test the functionality of this over at https://haven.pages.dev/torrent-client/public

DiegoRBaquero:
If hybrid-chunk-store is ESM, webtorrent can't yet support it

ThaUnknown:
> If hybrid-chunk-store is ESM, webtorrent can't yet support it

yes, but that can easily be changed, I could compile hybrid to 2 files, one module and one normal, focus on its code pls

ThaUnknown:
forgot this went to v2 and kinda borked it, fixed merge conflict and updated hcs version

ThaUnknown:
@DiegoRBaquero I've updated the package, also made it pass all these tests: https://github.com/ThaUnknown/hybrid-chunk-store/blob/master/test.js which are probably more strict than any of the tests used by the stores that we currently use, this is as good to merge as it's probably gonna get

DiegoRBaquero (approved):

Issue #1995, opened 2021-02-01
FAST extensions and reqq


arvidn:
**What version of this package are you using?**

"WebTorrent 0.0.1.2" (that's the version advertised in the extended handshake).

**What problem do you want to solve?**

As a client uploading or seeding content, I would like to be able to apply back-pressure to the WebTorrent client, telling it that it has requested more outstanding pieces than I'm allowing it. I (as an uploading bittorrent client) want to cap the amount of memory I need to allocate to store the list of outstanding requests.

Right now it seems WebTorrent either has no limit, or has a very high limit (several thousands) of outstanding piece requests to a single peer.

The immediate symptom of this problem is that a seeding client is likely to just drop peer requests that exceed their limit, causing the connection to stall and eventually time out.

**What do you think is the correct solution to this problem?**

there are two prongs to solving this problem. The immediate issue can be relieved by supporting the FAST extensions of 2008. Documented [here](https://www.bittorrent.org/beps/bep_0006.html). These new messages patch up a few holes in the original protocol.

Some of those FAST messages can simply be ignored. However, the ones that need to do something are:

HAVE_ALL, HAVE_NONE and REJECT_REQUEST. The last one allows a client to actively acknowledge it received a request, and says it will not service it. The requesting client can then keep accurate records of what requests it still has outstanding. This is especially helpful around CHOKE and UNCHOKE messages. But it can also be used to issue back-pressure when an outstanding-request queue limit is exceeded.

The other, presumably much simpler, feature would be to recognize and honor the "reqq" field in the extension message. This field is in the top level extension handshake dictionary, sibling to the client name (`v`). The value of the "reqq" field is the max number of outstanding piece request messages the client accepts. i.e. any REQUEST message sent to this client, where the corresponding PIECE message hasn't been returned, counts as an outstanding request.

**Are you willing to submit a pull request to implement this change?**

No, but I'd be happy to assist in explaining any aspects of the protocol.

rom1504:
this seems to be an important (blocker status) issue to solve for efficient communication between libtorrent and webtorrent

arvidn:
not having a limit of the number of outstanding requests one allows to queue is a potential DoS vector. If WebTorrent doesn't have a limit, i would also recommend it adds one.

feross:
This is something we should look at.

ThaUnknown:
> This is something we should look at.

[...]

PR #2243, opened 2022-01-14
feat: add BEP6 Fast Extension support

paullouisageneau:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

This PR implements [BEP6](https://www.bittorrent.org/beps/bep_0006.html) Fast Extension logic. It requires https://github.com/webtorrent/bittorrent-protocol/pull/75.

It addresses https://github.com/webtorrent/webtorrent/issues/1995

The main behavior change is the introduction of the allowed fast set mechanism. I chose to only implement the leecher side of it, as generating the allowed fast set is not documented for IPv6 (and even relying on the IPv4 feels hacky with WebRTC). Besides, it is out of the scope to fix https://github.com/webtorrent/webtorrent/issues/1995.

### DIFF
--- a/lib/peer.js
+++ b/lib/peer.js
@@ -327,7 +327,8 @@ class Peer extends EventEmitter {
 
   handshake () {
     const opts = {
-      dht: this.swarm.private ? false : !!this.swarm.client.dht
+      dht: this.swarm.private ? false : !!this.swarm.client.dht,
+      fast: true
     }
     this.wire.handshake(this.swarm.infoHash, this.swarm.client.peerId, opts)
     this.sentHandshake = true
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -650,6 +650,28 @@ class Torrent extends EventEmitter {
     this.bitfield.set(index, true)
   }
 
+  _hasAllPieces () {
+    for (let index = 0; index < this.pieces.length; index++) {
+      if (!this.bitfield.get(index)) return false
+    }
+    return true
+  }
+
+  _hasNoPieces () {
+    return !this._hasMorePieces(0)
+  }
+
+  _hasMorePieces (threshold) {
+    let count = 0
+    for (let index = 0; index < this.pieces.length; index++) {
+      if (this.bitfield.get(index)) {
+        count += 1
+        if (count > threshold) return true
+      }
+    }
+    return false
+  }
+
   /**
    * Called when the metadata, listening server, and underlying chunk store is initialized.
    */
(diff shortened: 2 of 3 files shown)


DiegoRBaquero (approved):

lib/peer.js
```diff
@@ -327,7 +327,8 @@ class Peer extends EventEmitter {
 
   handshake () {
     const opts = {
-      dht: this.swarm.private ? false : !!this.swarm.client.dht
+      dht: this.swarm.private ? false : !!this.swarm.client.dht,
+      fast: true
```
arvidn, line 331: there is a bit in the reserved handshake field used to indicate whether the peer supports fast extension or not. Perhaps I'm misreading this, but it looks like you just assume the peer supports the FAST extension.

http://bittorrent.org/beps/bep_0006.html

> reserved[7] |= 0x04

arvidn (commented):
there's also the reject-request message, which can't really be ignored. This is just affecting the higher level bittorrent engine, right? because I don't see the actual bittorrent messages in wire-format anywhere.

lib/peer.js
paullouisageneau: It only enables Fast if both peers support it, this is actually done in the dependency [webtorrent/bittorrent-protocol](https://github.com/webtorrent/bittorrent-protocol/blob/94f72e96e55a90a261c064495dcbda13c653a0cd/index.js#L752).

paullouisageneau:
> there's also the reject-request message, which can't really be ignored. This is just affecting the higher level bittorrent engine, right? because I don't see the actual bittorrent messages in wire-format anywhere.

Indeed, this is only the higher-level logic, the wire-format and base logic is added in https://github.com/webtorrent/bittorrent-protocol/pull/75. In particular, the reject message is handled there by [removing the request from the list and triggering its callback with an error](https://github.com/webtorrent/bittorrent-protocol/blob/94f72e96e55a90a261c064495dcbda13c653a0cd/index.js#L893).

paullouisageneau:
> there's also the reject-request message, which can't really be ignored. This is just affecting the higher level bittorrent engine, right? because I don't see the actual bittorrent messages in wire-format anywhere.

@arvidn Indeed, this is only the higher-level logic, the wire-format and base logic is added in https://github.com/webtorrent/bittorrent-protocol/pull/75. In particular, the reject message is handled there by [removing the request from the list and triggering its callback with an error](https://github.com/webtorrent/bittorrent-protocol/blob/94f72e96e55a90a261c064495dcbda13c653a0cd/index.js#L893).

paullouisageneau:
> there's also the reject-request message, which can't really be ignored. This is just affecting the higher level bittorrent engine, right? because I don't see the actual bittorrent messages in wire-format anywhere.

@arvidn Indeed, this is only the higher-level logic, the wire-format and base logic is added in https://github.com/webtorrent/bittorrent-protocol/pull/75. In particular, the reject message is handled there by removing the request from the list and triggering its callback with an error.

PR #2282, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -214,7 +214,7 @@ class WebTorrent extends EventEmitter {
           }
           port.postMessage(chunk)
           if (!chunk) cleanup()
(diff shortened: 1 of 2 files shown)


DiegoRBaquero (approved):

DiegoRBaquero:
Thank you <3

PR #2337, opened 2022-06-25
feat: Unifiy HTTP server and SW renderer.

ThaUnknown:
Requires https://github.com/webtorrent/webtorrent/pull/2336 merged first!!!

[ ] Documentation update
[X] Bug fix
[X] New feature
[ ] Other, please explain:

Unified HTTP server and SW renderer. Both now use almost the same methods and functionality, and behave in the same way.
This means all the functionality the SW was lacking and HTTP server had is available in SW and vice versa.

The SW can now display root indexes of webtorrent which will list all torrents, view indexes of torrents which will list all files etc.

The HTTP server now targets the ENTIRE client, which means its root directory will first list all torrents, then all files.
The HTTP server now causes files to emit `stream` events, just like the SW renderer.

These are breaking changes, the url's for the HTTP server is no longer `/` but `/webtorrent/` and its paths are `/webtorrent/<infohash>` just like my FS changes, to make sure files don't overwrite each other. There's also no more of that useless file identifier thing with index.

The getStreamURL and streamTo methods now also work with the HTTP server for mixed environments such as NWjs or Electron. They also no longer use callbacks.

The client.createServer can be force overwritten to a specific implementation for those mixed environments.

Fixed an edge case where the SW renderer could leave a port open.
Fixed an edge case where weird SW registration paths would cause errors.

Considerations:
for the browser version, should handling the service worker instance the user wants the server to use be done via createServer? or listen method once the server is created?

for the browser version, should we implement the origin path which in theory is possible? or would that be too missleading for some cases?

### DIFF
--- a/docs/tutorials.md
+++ b/docs/tutorials.md
@@ -14,7 +14,6 @@ import WebTorrent from 'webtorrent'
 const client = new WebTorrent()
 const torrentId = 'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent'
 const player = document.querySelector('video')
-const scope = './'
 
 function download () {
   client.add(torrentId, torrent => {
@@ -31,7 +30,7 @@ function download () {
     console.log('Ready to play!')
   })
 }
-navigator.serviceWorker.register('./sw.min.js', { scope }).then(reg => {
+navigator.serviceWorker.register('./sw.min.js', { scope: './' }).then(reg => {
   const worker = reg.active || reg.waiting || reg.installing
   function checkState (worker) {
     return worker.state === 'activated' && client.createServer({ controller: reg }) && download()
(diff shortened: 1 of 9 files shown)


ThaUnknown:
forgot to clean this up, now it's cleaned up @DiegoRBaquero

docs/tutorials.md
```diff
@@ -12,8 +12,9 @@ Code example:
 import WebTorrent from 'webtorrent'
 
 const client = new WebTorrent()
-const torrentId = "magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent"
+const torrentId = 'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent'
 const player = document.querySelector('video')
+const scope = './'
```
DiegoRBaquero, line 17: I'd prefer this inlined

docs/tutorials.md
```diff
@@ -82,30 +82,30 @@ Code example:
...
-          return worker.state === 'activated' && client.loadWorker(worker, download)
-        }
-        if (!checkState(worker)) {
-          worker.addEventListener('statechange', ({ target }) => checkState(target))
-        }
-      })
+const client = new WebTorrent()
+const torrentId = 'magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent'
+const player = document.querySelector('video')
+const scope = './'
```
DiegoRBaquero, line 88: Same

docs/tutorials.md
ThaUnknown: I left this out like this, in case someone runs in a sub-scope like say github pages which would be for example `https://thaunknown.github.io/webtorrent_app/` where the scope would be changed to `./webtorrent_app/`, in the same way as the magnet would be changed by the user

do you still think it should be inlined?

DiegoRBaquero: They can still change it, but for the sake of the tutorial, where the variable is only used once in a single place, it should be inlined.

ThaUnknown: > They can still change it, but for the sake of the tutorial, where the variable is only used once in a single place, it should be inlined.

its done the same for the magnet tho, which is also used once, imho its better to keep it not inline

DiegoRBaquero: magnet is long tho and in a real app it will come from a variable

DiegoRBaquero (approved):

Issue #1971, opened 2020-11-24
Consider migrating from `readable-stream` to `streamx`


feross:
Advantages of [`streamx`](https://github.com/streamxorg/streamx) over [`readable-stream`](https://github.com/nodejs/readable-stream):

- Better performance (according to @mafintosh's testing... would be nice to see some numbers)
- Massively smaller browser bundle size (due to much simpler implementation)
  - stream: 18053 bytes (gzipped)
  - streamx: 2806 bytes (gzipped)
- Preserves 95% of the same Node.js API so most users shouldn't notice a difference
- Written by maintainers of `readable-stream` (so builds on those learnings, and possibly will become the next version of Node.js core streams)
- Opens the possibility of removing `Buffer` and using `Uint8Array` directly throughout `webtorrent`
  - This is possible because unlike `readable-stream`, `streamx` doesn't require `buffer`
  - Without `streamx`, it would be pointless to stop requiring `buffer` in `webtorrent` since `readable-stream` would just bring it into the bundle anyway

In order for the smaller browser bundle size benefits to be realized, we'd need to port the whole `webtorrent` ecosystem of packages. It looks like that's 13 packages using `readable-stream` in total.

```
webtorrent@0.111.0 /Users/feross/code/webtorrent
├─┬ bittorrent-protocol@3.2.0
│ └── readable-stream@3.6.0  deduped
├─┬ bittorrent-tracker@9.15.0
│ └─┬ simple-websocket@8.1.1
│   └── readable-stream@3.6.0  deduped
├─┬ chunk-store-stream@4.1.1
│ ├─┬ block-stream2@2.0.0
│ │ └── readable-stream@3.6.0  deduped
│ └── readable-stream@3.6.0  deduped
├─┬ create-torrent@4.4.2
│ ├─┬ filestream@5.0.0
│ │ └── readable-stream@3.6.0  deduped
│ └── readable-stream@3.6.0  deduped
├─┬ multistream@4.0.1
│ └── readable-stream@3.6.0  deduped
├── readable-stream@3.6.0 
├─┬ render-media@4.0.0
│ ├─┬ mediasource@2.4.0
│ │ └── readable-stream@3.6.0  deduped
│ └─┬ videostream@3.2.2
│   ├─┬ mp4-stream@3.1.0
│   │ └── readable-stream@3.6.0  deduped
│   └─┬ range-slice-stream@2.0.0
│     └── readable-stream@3.6.0  deduped
[...]

PR #2339, opened 2022-06-26
fix: use streamx instead of stream

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[X] Other, please explain:

Migrated from `stream` to `streamx`. It's usage is 99% the same except for some function parameters and the requirement of callbacks after each method.
https://github.com/webtorrent/webtorrent/issues/1971
Potential compat with other libs which use `stream`

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -13,7 +13,7 @@ const debug = debugFactory('webtorrent:file-stream')
  */
 class FileStream extends Readable {
   constructor (file, opts) {
-    super(opts)
+    super(opts ?? {})
 
     this._torrent = file._torrent
 
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -1,6 +1,5 @@
-const stream = require('stream')
+const { Readable } = require('streamx')
 const debugFactory = require('debug')
-const eos = require('end-of-stream')
 
 const debug = debugFactory('webtorrent:file-stream')
 
@@ -12,7 +11,7 @@ const debug = debugFactory('webtorrent:file-stream')
  * @param {number} opts.start stream slice of file, starting from this byte (inclusive)
  * @param {number} opts.end stream slice of file, ending with this byte (inclusive)
  */
-class FileStream extends stream.Readable {
+class FileStream extends Readable {
   constructor (file, opts) {
     super(opts)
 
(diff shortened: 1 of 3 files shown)


lib/file-stream.js
```diff
@@ -12,9 +11,9 @@ const debug = debugFactory('webtorrent:file-stream')
  * @param {number} opts.start stream slice of file, starting from this byte (inclusive)
  * @param {number} opts.end stream slice of file, ending with this byte (inclusive)
  */
-class FileStream extends stream.Readable {
+class FileStream extends Readable {
   constructor (file, opts) {
```
DiegoRBaquero, line 15:
```suggestion
  constructor (file, opts = {}) {
```

lib/file-stream.js
```diff
@@ -12,9 +11,9 @@ const debug = debugFactory('webtorrent:file-stream')
...
-    super(opts)
+    super(opts ?? {})
```
DiegoRBaquero, line 16:
```suggestion
    super(opts)
```

ThaUnknown: issue with this, is that it allows null, and http server creates null ranges, which will error this

ThaUnknown: compare 1st vs 2nd commit, look at the test fail

lib/file-stream.js
```diff
@@ -92,11 +87,12 @@ class FileStream extends stream.Readable {
       this.push(buffer)
 
       if (this._missing === 0) this.push(null)
+      cb()
     })
     this._piece += 1
   }
 
-  _destroy (err, cb) {
+  _destroy (cb, err) {
```
DiegoRBaquero, line 95: Why is this inverse?

ThaUnknown: oh, reallistically err shouldnt be here

ThaUnknown: no, error should be there, this is correct, in streamx the order is inversed

DiegoRBaquero (approved):

lib/file-stream.js
ThaUnknown: see:
https://github.com/streamxorg/streamx/blob/master/index.js#L189
https://github.com/streamxorg/streamx/blob/master/index.js#L440

this is in case _destroy is called with an error without calling destroy(), which is... shouldnt be done, but since it was there for the original i kept it

PR #2347, opened 2022-07-03
fix: dedupe packages

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Dropped stream-to-blob and stream-to-blob-url, as create-torrent now uses fast-blob-stream which is functionally the same as the 2 [except that fast-blob-stream forces a streamx depencency, while steam-to-blob didn't as it was read only, no write (or other way around, depends on how you look at it)]. Functionally these are the same.

This de-dupes packages. Less depencencies. Smaller bundle size. I don't know why this didn't use file-stream as file and blob are the same thing, except that file has an extra name and mod date.


DiegoRBaquero (approved):

DiegoRBaquero:
Why not merge?

ThaUnknown:
> Why not merge?

I feel like I'm not the one to make that call @DiegoRBaquero as I haven't had many interactions with the maintainers so I don't know what the ideology is, or what the target is. If it was up to me I would have probably changed 80% of webtorrent as of now

DiegoRBaquero:
You are a member, as soon as it has approvals you are good to go :)

PR #2239, opened 2022-01-09
refactor: Return Torrent when client.seed finds a duplicate

Bruce-Hopkins:
**What is the purpose of this pull request? (put an "X" next to item)**

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

**What changes did you make? (Give an overview)**
When calling `client.seed` it would usually throw an error. Now it passes the torrent instance in the callback and adds a warning on the console.

**Which issue (if any) does this pull request address?**
 #2215

**Is there anything you'd like reviewers to focus on?**

The way I added the callback, there is no way to know if the torrent is a duplicate or not. Should we also pass a boolean to represent whether the torrent is a duplicate or not?

PR #2372, opened 2022-09-11
fix: return duplicate torrent on Webtorrent.add()

resession:
if the there is a duplicate torrent, return the torrent

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[X ] Other, please explain:

when we add duplicate torrents, we get an error. this will throw and error but also return the duplicate torrent

duplicate torrents

no, there is nothing specific, only two lines were added

### DIFF
--- a/index.js
+++ b/index.js
@@ -286,6 +286,8 @@ class WebTorrent extends EventEmitter {
       for (const t of this.torrents) {
         if (t.infoHash === torrent.infoHash && t !== torrent) {
           torrent._destroy(new Error(`Cannot add duplicate torrent ${torrent.infoHash}`))
+          ontorrent(t)
+          this.emit('torrent', t)
           return
         }
       }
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -286,6 +286,8 @@ class WebTorrent extends EventEmitter {
       for (const t of this.torrents) {
         if (t.infoHash === torrent.infoHash && t !== torrent) {
           torrent._destroy(new Error(`Cannot add duplicate torrent ${torrent.infoHash}`))
+          ontorrent(t)
+          this.emit('torrent', t)
```
DiegoRBaquero, line 290: I don't agree on emitting torrent again. The callback does make sense.

Similar to https://github.com/webtorrent/webtorrent/pull/2239/files

resession: fixed

DiegoRBaquero (approved):

Issue #2217, opened 2021-11-08
sha1 wrapper (Node 14, WebCrypto)


DiegoRBaquero:

ThaUnknown:
https://github.com/jimmywarting/wt-common
?

ThaUnknown:
yes

DiegoRBaquero:
To be honest, I don't remember. I think we can close this

ThaUnknown:
> To be honest, I don't remember. I think we can close this

I'd happily drop rusha, it's a lot less bundle size

ThaUnknown:
yeah, re-opening this for https://github.com/orgs/webtorrent/projects/1#card-60702974 and 4.8% bundle size reduction

ThaUnknown:
cool, web crypto is AT BEST 900% slower than `node:crypto` and only 25% faster than `rusha` this is really bad....

ThaUnknown:
related PR's:
https://github.com/webtorrent/bittorrent-protocol/pull/98
https://github.com/webtorrent/create-torrent/pull/184
https://github.com/webtorrent/parse-torrent/pull/117
https://github.com/webtorrent/ut_metadata/pull/79
https://github.com/webtorrent/bittorrent-dht/pull/257
https://github.com/webtorrent/webtorrent/pull/2390
but this is pretty much umergable util ESM/Promises, but it's there

PR #2390, opened 2022-11-02
feat: drop rusha

ThaUnknown:

[ ] Documentation update
[X] Bug fix
[X] New feature
[ ] Other, please explain:

REQUIRES ESM

Dropped `simple-sha1` for `uint8-util` in order to drop `rusha`. `uint8-util` uses `node:crypto` in node and Web Crypto API in browser, as `node:crypto`'s sha1 is an order of magnitude faster than the Web Crypto API.

This also exposes a few snippets from [wt-common](https://github.com/jimmywarting/wt-common/) for the future removal of `Buffer.from` and `Buffer.toString`, however not all of them, as some plain JS are a LOT faster `Buffer.from`, and wt-common forced `Buffer.from` if node was used.

*For now* `uint8-util` is 85% smaller than `simple-sha1`, but that will change to roughly 65% smaller once I get around to adding all the Buffer.from-like functionality to uint8.
https://github.com/webtorrent/webtorrent/issues/2217
https://github.com/orgs/webtorrent/projects/1#card-60702974


DiegoRBaquero:
Given this requires ESM, it should target v2

ThaUnknown:
> Given this requires ESM, it should target v2

probably, but realistically we will only be able to merge this after v2 is merged into master

DiegoRBaquero (approved):

Issue #2373, opened 2022-09-13
torrent.remove() error in remove peers


wentony30:
![image](https://user-images.githubusercontent.com/62290720/190013209-49433504-b9ac-495b-9998-91ff939e4274.png)

When removing the torrent it gives an error when removing the peers.
Validate when getting the peer id, I think that would solve it.

Jassi10000:
Also please check if  ( this._peers  ) is getting some value

ThaUnknown:
https://github.com/webtorrent/webtorrent/issues/2299

Issue #2299, opened 2022-04-11
Cannot get peer. Crashes application.


Ameliaaaaaaa:
**What version of this package are you using?**: 1.8.13

**What operating system, Node.js, and npm version?**: Ubuntu 20.04 LTS x64, Node v16.14.2, NPM v8.5.0

**What happened?**:
```
/root/App/node_modules/webtorrent/lib/torrent.js:954
    peer = this._peers[id]
TypeError: Cannot read properties of null (reading 'IP_ADDRESS:50065')
    at Torrent.removePeer (/root/App/node_modules/webtorrent/lib/torrent.js:954:23)
    at Peer.destroy (/root/App/node_modules/webtorrent/lib/peer.js:386:22)
    at Timeout.<anonymous> (/root/App/node_modules/webtorrent/lib/peer.js:347:12)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

**What did you expect to happen?**:
Handled so it doesn't crash.

**Are you willing to submit a pull request to fix this bug?**: If I know how.

This started after I destroyed torrents once they were finished (keeping store)

flyingspoon:
Bump!

Having the same issue:

```
/root/project/node_modules/webtorrent/lib/torrent.js:954
    peer = this._peers[id]

TypeError: Cannot read property 'IP_ADDRESS:30779' of null
    at Torrent.removePeer (/root/project/node_modules/webtorrent/lib/torrent.js:954:23)
    at Peer.destroy (/root/project/node_modules/webtorrent/lib/peer.js:386:22)
    at Timeout.<anonymous> (/root/project/node_modules/webtorrent/lib/peer.js:355:12)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7)
```
How can I handle it, are there any fixes or solutions?

Ameliaaaaaaa:
It happens when you destroy the torrent. Only way I know of is not destroying the torrent

flyingspoon:
For my use-case, I need to destroy the torrent sadly.
Is there no way to handle the error and after re-adding the torrent, maybe change peers for example, and if one is null, then change to one that is not null?

Hope someone can post a solution

risharde:
This issue is still not fixed!!!

flyingspoon:
> This issue is still not fixed!!!

[...]

PR #2398, opened 2022-11-12
fix: removePeer error after destroy

ThaUnknown:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

!!! THIS IS A PATCH, NOT A FIX

seems like `removePeer` was being called after a torrent was destroyed, this was either caused by the peers own [`destroy` function](https://github.com/webtorrent/webtorrent/blob/master/lib/peer.js#L386) or by [pex](https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L1098). I'm not sure as I'm not able to reproduce the issue, but this will fix the ERROR, not any other underlying problems.
#2373 #2299
Does this potentially leak memory? should `peer.destroy` still be called if the peer is an object, not a string ID?

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -951,17 +951,18 @@ class Torrent extends EventEmitter {
   }
 
   removePeer (peer) {
-    const id = (peer && peer.id) || peer
-    peer = this._peers[id]
+    const id = peer?.id || peer
+    if (peer && !peer.id) peer = this._peers?.[id]
 
     if (!peer) return
+    peer.destroy()
+
+    if (this.destoyed) return
 
     this._debug('removePeer %s', id)
 
     delete this._peers[id]
     this._peersLength -= 1
-
-    peer.destroy()
 
     // If torrent swarm was at capacity before, try to open a new connection now
     this._drain()


risharde:
This was one of them I remember off hand, great to see a fix, at least it being catched to avoid a full on crash, thanks!

DiegoRBaquero (approved):

ThaUnknown:
@DiegoRBaquero are you sure.... this is a patch not a fix, could leak memory like crazy

DiegoRBaquero:
We could throw and error just like in `addPeer` https://github.com/webtorrent/webtorrent/blob/4b599f9f0dbfa95ed7168262d138ea46098cc7ea/lib/torrent.js#L783

ThaUnknown:
> We could throw and error just like in `addPeer`
> 
> https://github.com/webtorrent/webtorrent/blob/4b599f9f0dbfa95ed7168262d138ea46098cc7ea/lib/torrent.js#L783

lol no
https://github.com/webtorrent/webtorrent/blob/4b599f9f0dbfa95ed7168262d138ea46098cc7ea/lib/torrent.js#L703-L723

lib/torrent.js
```diff
@@ -951,6 +951,7 @@ class Torrent extends EventEmitter {
   }
 
   removePeer (peer) {
+    if (!this._peers) return
```
DiegoRBaquero, line 954:
```suggestion
    if (!this._peers) throw new Error('torrent is destroyed')
```

ThaUnknown: this will error when PeX is closing/cleaning up, or when PeX send a drop peer request as the torrent is being destroyed
https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js#L1098

ThaUnknown: also this is just relocating the error, as webtorrent iself would throw the error instead of erroring 2 lines later on `peer = this._peers[id]` aka what this is trying to fix

ThaUnknown:
there @DiegoRBaquero that seems safe, it will try to destroy the peer, even if stuff has already started cleaning up, might be duplicate, but doesn't really matter as peer.destroy() also returns if its already destroyed, so no loop is made, and this is sure to not leak any memory

DiegoRBaquero (approved):

Issue #2351, opened 2022-07-12
Reduced transfer speed after v1.3.5


marcovidonis:
I'm noticing a greatly reduced transfer speed when importing the project from a `<script>` tag or with a bundler other than browserify.
As a comparison of how much reduced, I tested a file transfer between 2 instant.io clients, running in 2 browser windows on the same machine, and easily achieved a transfer rate of 20-30MB/s.
When transferring the same file between 2 instances of a simple browser client I made, in which I imported the latest WebTorrent version, speed seemed to be capped at around 3MB/s. 

This happened both in a simple HTML page where I imported WebTorrent from a `<script>` tag and in a simple React app that I made using Vite, where the project was imported using scriptjs. I could also see the same speed cap in other WebTorrent projects, including [btorrent.xyz](https://btorrent.xyz).

Here's a link to my [example app](https://github.com/nimbleape/webtorrent-browser-react).

I noticed that [another project using WebTorrent](https://github.com/blenderskool/blaze) was reaching transfer speeds of 30-40MB/s (same file, same test conditions). This project does not rely on browserify, but imports a much older version of WebTorrent.

I then tried different versions of the package to see whether that affected transfer speed and indeed I found that I reliably get >30MB/s with v.1.3.4 and older, while speed is capped at 3MB/s with v.1.3.5 and later.

**What version of this package are you using?**
v1.3.4, v1.3.5 and v1.8.26.

**What operating system, Node.js, and npm version?**
macOS v12.4, node v14.18.1, npm v8.5.4.

**What happened?**
Limited transfer speed when using WebTorrent v1.3.5 or later and not bundling the app with browserify.

**What did you expect to happen?**
Same speed transfer as older versions.

**Are you willing to submit a pull request to fix this bug?**
I'm not really sure where the problem lies.

ThaUnknown:
while I can reproduce this on your apps, on mine such an issue doesn't exist:
![image](https://user-images.githubusercontent.com/6506529/178608468-fd1401bc-94fe-45c5-8d40-116a9b91ce4f.png)
i'm uncertain what the underlying issue could be :/

ThaUnknown:
I checked every dependency change from 1.3.4 -> 1.3.5, the issue doesn't lie there I don't think

marcovidonis:
I've had a look at the vite config you used [here](https://github.com/ThaUnknown/miru), is this the app you're referring to?

[...]

Issue #2191, opened 2021-09-17
shim process/process-browserify with process-fast


ThaUnknown:
This should improve performance as process-fast doesn't use set timeouts, but I don't know how to do it in browserify. Mainly meant for sub-dependencies such as readable-stream

Issue #1965, opened 2020-11-17
Use Browserify builtins feature to specify Chrome App builtins instead of "chromeapp" package.json field


feross:
To support a Chrome App build for the Brave Browser WebTorrent extension, I added replacements for the Node Builtins that work in a Chrome App environment. For example, `net` is replaced by `chrome-net`, `dgram` is replaced by `chrome-dgram`, and `dns` is replaced by `chrome-dns`, etc.

In order to pull this off, we had to add e.g. `chrome-net` as a dependency of any package that used `require('net')` and then we'd use a custom `package.json` "browser" field with the `chromeapp` key to specify that in a chrome app the following replacement should occur:

```json
{
  "chromeapp": {
    "net": "chrome-net"
  }
}
```

And then we run browserify with the `--browser-field=chromeapp` argument.

However, this is hard to do for every WebTorrent dependency and it means that extra dependencies are installed even when the user is not using WebTorrent in a Chrome App.

Instead, let's remove all these `chrome-*` dependencies and just handle these replacements in the `webtorrent` repo. There's a description of how we can do this in this issue: https://github.com/jscissr/http-node/pull/2

Specifically, we'll tell browserify that instead of the usual replacements it does for browser builds, we want different replacements for `net`, `dgram`, and `dns`. We'll need to write a short build script using the Browserify API to do this, though eventually Browserify will support a command line option for this, as @goto-bus-stop mentioned in a comment in that issue.

Issue #1568, opened 2019-01-11
Transfer is seriously strangled while browser tab is in background


jimmywarting:
**What operating system and Node.js version?**
Dosen't affect Node

**What version of WebTorrent?**
0.103.x

**What browser and version? (if using WebTorrent in the browser)**
Chrome v72

**What did you expect to happen?**
Expected `process.nextTick` to not use `setTimeout`

**What actually happened?**
The problem is with `process` browserified version of `process.nextTick` it uses a `setTimeout(fn, 0)` as a fallback. The `setTimeout` is delayed well enough in foreground to strangle the speed from 6-7000 kb/s down to just 400 kb/s when one tab is hidden "to save resources". if you instead use two windows the speed will be fast (normal) again.

> `setTimeout(fn, 0)` can potentially be used, however as it is clamped to 4ms for timers nested more than 5 deep per the HTML spec, it does not make for a suitable [polyfill](https://github.com/YuzuJS/setImmediate) for the natural immediacy of setImmediate. - [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate)

Downloading from hybrid torrents works fine cuz they uses node nextTick version.
So to test this you basically need to upload some large files to [instant.io](https://instant.io/) and open up the sharable instant.io link in a new tab or new window

This affect users who want to download something and watch cat videos while its uploading/downloading

RangerMauve:
What would be a better alternative to process.nextTick? It might be possible to detect the browser environment and reassign it to something better or do a PR against browserify.

jimmywarting:
> What would be a better alternative to process.nextTick?

If possible don't use nextTick at all in both `memory-chunk-store` and `immediate-chunk-store` and possible other places where it's not needed? not sure if streams uses nextTick also.

To me it looks like: 
- PeerA request a piece from PeerB
- PeerB calls `store.get()`
- store calls memory `store.get()`
- memory `store.get` waits 4ms and then return it to immediate store
- immediate waits 4ms and then gives that to bittorent-protocol

now you waisted 8ms or more. Both storage layer don't use the same eventLoop when nextTicket is called in both of them it's more like a chain reaction then a parallel execution

It seems to me that this is also almost all pointless b/c of `process.nextTick()` turns into `setTimeout(fn, 0)`
https://github.com/webtorrent/webtorrent/blob/5b7c5eebf41c9e1e1fe8e9e1d4a6ad06e93e7430/lib/torrent.js#L46-L47

[...]

PR #2399, opened 2022-11-13
perf: drop browserify

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[X] Other, please explain:

Replaced browserify with webpack.
Why webpack? Because I couldn't get esbuild/rollup to do what I wanted without writing custom plugins, too much time.

Why ditch browserify?
- bad polyfills
- lack of customizability
- unnecessary polyfills [over 500kb!!!!!]

What actually changed?
- process-browserify was replaced with a modified version of [mathiasvr's process-fast](https://github.com/mathiasvr/process-fast/), modified because it didn't ponyfill queueMicrotask
- zlib-browserify was replaced with pako [for chromeapp] which is a LOT faster [again thanks for mathiasvr for figuring this out]
- added sourcemaps [wooo \o/] and ditched build-debug for that reason
- output is now an ESM module

TLDR: end result is a lot faster, a lot smaller and easier to debug

Oh my... it pretty much removes all those useless setTimeout calls in process.nextTick
#2351 #2191 #1568 #1026 [kind of? not really?] #903 [partial as this is also hashing, but can safely be closed]
I dont quite understand #1965 but I feel like it could be addressed here too? explain pls?


dist/webtorrent.chromeapp.js.LICENSE.txt
```diff
@@ -0,0 +1,97 @@
+/*
```
DiegoRBaquero, line 1: Can we not generate these LICENSE files?

dist/webtorrent.min.js.LICENSE.txt
```diff
@@ -0,0 +1,74 @@
+/*!
```
DiegoRBaquero, line 1: same

dist/webtorrent.chromeapp.js.LICENSE.txt
ThaUnknown: is that.... a good idea?[ I don't really know much about copyright](https://github.com/ThaUnknown/jassub/issues/11) so I left it in

DiegoRBaquero:
Given the new `dist` folder, a lot of external usage might break. So it's good about this being v2, any other consequences you can think of?

dist/webtorrent.chromeapp.js.LICENSE.txt
DiegoRBaquero: We've never had it, never had any issues, we can include them once some copyright nerd requires them (?)

ThaUnknown: maybe I could generate only one license file instead of both?

ThaUnknown:
> Given the new `dist` folder, a lot of external usage might break. So it's good about this being v2, any other consequences you can think of?

chromeapp.js might be broken? but I don't know how to even test that as there are no webtorrent chromeapps in the world I don't think, from what I tested all of webtorrent works fine when I ran some random local tests. Other people's browserify configs wont work anymore, but that's because it doesn't support ESM, not my fault. Since we don't do setTimeouts for process.nextTick some events might potentially run in different orders if some1 was to say run multiple webtorrent instances in the same window, but its a very minor thing and it wouldnt cause any errors, a good example is [this test](https://github.com/ThaUnknown/bittorrent-protocol/commit/e87a4ab052067e3ad9e341373faabf2aba5187a9#diff-61b342b8e5ffd8d1a2485be9eb3785549ecb329ac5d3d236c233656d8eaecef3R60)

dist/webtorrent.chromeapp.js.LICENSE.txt
ThaUnknown: @DiegoRBaquero done

ThaUnknown:
I'm debating if the _polyfills should be in /lib or their own folder

ThaUnknown:
image
image
welp all tests pass so that's great, we still don't have sw tests as airtap doesn't do static files I don't think, idk what to do about that

ThaUnknown:
@DiegoRBaquero good news, unlike the `master` branch's `webtorrent.chromeapp.js` this one actually works XD at least I think so
TLDR this is mostly useless as in a few months chrome is deprecating v2 so no socket access, and I couldn't ACTUALLY get it to work as I would need to publish some package, create a security key and some other shit, which I can't do, but IN THEORY this should work fine
I ran this code:
```js
import WebTorrent from './webtorrent.chromeapp.js'
localStorage.debug = '*'
const noop = () => {}
chrome.sockets = {
  tcpServer: {
    create: noop
  },
  udp: {
    create: noop
  },
  tcp: {
    create: noop
  }
}
chrome.dns = {
  resolve: noop
}

console.log(client)
fetch('http://webtorrent.io/torrents/sintel.torrent').then(async res => {
  const blob = await res.blob()
  const tor = client.add(blob, console.log)
  tor.on('metadata', console.warn)
  tor.on('_infoHash', console.log)
  console.log(tor)
})
```
had to make noops for the network stuff as I don't have access to it, and everything seems to work fine, metadata and ready never fire, because it needs to wait for ConnPool to connect to do so, which is noop, but metadata etc works, nothing errors on evaluation unlike the `master` branch one.
debug:
image
seems good, supposedly brave will be keeping v2 manifest, so sockets might work there

ThaUnknown:
@DiegoRBaquero any other concerns?

ThaUnknown:
notably `pako` can be replaced with `fflate` for even faster speeds and lower bundle sizes, but I'm not certain about compatibility with `zlib`

DiegoRBaquero (approved):

PR #2409, opened 2022-11-21
feat: sw renderer tests

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[x] New feature
[ ] Other, please explain:

Added quite strict service worker renderer tests, and fixed bugs that creating those tests revealed.

\-

The electron exclusion is there from when render-media was used, I don't understand why it wouldn't work, so I kept it in. Airtap doesn't serve static files, so we need to create a support server in order to register the service worker as it needs to be a standalone file and in the same scope.
Also used an external torrent as fixtures doesn't provide video anymore :/

### DIFF
--- a/.airtap.yml
+++ b/.airtap.yml
@@ -24,3 +24,4 @@ browserify:
     global: true
     presets: ['@babel/preset-env']
     plugins: ['@babel/plugin-syntax-import-assertions']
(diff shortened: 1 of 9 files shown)


ThaUnknown:
@DiegoRBaquero 😉

DiegoRBaquero (approved):

PR #2411, opened 2022-11-26
Update file-stream to operate on larger chunks.

gdborton:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

I updated file-stream to read in batches of 10mb where possible.

I did this because I'm using the idb-chunk-store instead of the default in memory store. IndexDB allows for huge amounts of additional storage (especially compared to memory), but is slow. Each chunk (16kb) lookup takes 5-25ms. In practice this meant reading a single 7mb file would take up to 3 seconds in my app.

After doing some local testing I found I could read the entire contents of the file in ~20ms by requesting each chunk in parallel.

This change batches requests to 10mb where possible, limiting the memory used, but more fully saturating w/e the async bottleneck is in indexDB, but appends content serially. 

N/A

Error handling? I added a happy path test, but I'm not sure if I got everything correct here. Will be using in my own project in the mean time and will port back any fixes that I have to make there.

### DIFF
--- a/lib/file-stream.js
+++ b/lib/file-stream.js
@@ -50,6 +50,7 @@ class FileStream extends Readable {
     if (!this._reading || this._missing === 0) return cb()
     const maxPieces = Math.floor(this._readLength / this._torrent.pieceLength) | 1
     const piecesToLoad = Math.max(Math.min(maxPieces, this._endPiece - this._piece), 1)
+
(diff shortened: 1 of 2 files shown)


ThaUnknown:
superseeded by https://github.com/webtorrent/webtorrent/pull/2414

PR #2414, opened 2022-11-27
feat, perf: w3c-like File, file-iterator

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Changed file-stream to file-iterator, which allows us to have one codebase for both node and web streams, based on that implemented some w3c-like file functions, this is cool for libraries which use w3c files as interfaces.

Additionally if the file is already downloaded it will go directly from the store via `chunk-store-iterator` rather than via the stream, this is cool because `chunk-store-iterator` extends the underlying stores own iterator if it implements one, allowing for 0 copy and close to 0 overhead data passthrough.

Additionally the iterator supports concurrency, so one could do:
```js
import par from 'it-parallel'

for await (const value of par(file, { concurrency: 5, ordered: true })) {
  // stuff
}
```
or chain more transforms directly on the data reads:
```js
import par from 'it-parallel'
import block from 'block-iterator'

for await (const value of block(par(file, { concurrency: 5, ordered: true }), 10 * 1024 * 1024, { zeroPad: false })) {
  // stuff
}
```
or create a different stream from it
```js
import par from 'it-parallel'
import block from 'block-iterator'
import { Readable } from 'streamx'

const stream = Readable.from(block(par(file, { concurrency: 5, ordered: true }), 10 * 1024 * 1024, { zeroPad: false })) 
```

Made some API's async.

Dependency hygine.

Dropped a Buffer requirement package.
- #2411 
- #2055 [partial progress]

// TODO docs
Should we drop callbacks in the now async API's?
Should we drop the node names for getBuffer and only use arrayBuffer? [partial progress towards #1971]
Is there any point in keeping getBlobURL? 
Should `slice` be implemented as a FilePart proxy?


DiegoRBaquero (approved):

DiegoRBaquero:
> Should we drop callbacks in the now async API's?
Yes

> Should we drop the node names for getBuffer and only use arrayBuffer? [partial progress towards https://github.com/webtorrent/webtorrent/issues/1971]
I'm not sure what you mean by "drop the node names"

> Is there any point in keeping getBlobURL?
because we now have `await File.blob()`?

> Should slice be implemented as a FilePart proxy?
again not sure what you mean

ThaUnknown:
> I'm not sure what you mean by "drop the node names"

get rid of `getBuffer`

PR #2431, opened 2022-12-17
chore: docs

ThaUnknown:

[X] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

This updates docs to v2's changes, a lot of which went undocumented because well.. I wasn't sure of them.

Notable changes:

instead of using `jsdelivr` we now suggest `esm.sh`, they don't actually correctly compile webtorrent YET, but that's likely to change very soon, this is more future-proofing than anything. Upside of esm.sh is that it's packages aren't bundled like when we said to use the built file of webtorrent, meaning module de-duping if they use esm.sh for other modules.

added webpack suggestions with browserify in a lot more places, as a proper webpack config will yield an almost 80% smaller bundle size, with links to out webpack config

in some places I used `navigator.serviceWorker.ready` it's not exactly the correct way of registering the worker, but it's simpler and less code, worst case is that streams won't support cancelling instead use 5s timeouts on the first time the service worker is loaded

fixed a few issues which I found during writing of the docs and with real-world tests

removed all instances of getBlobURL, mainly because blob url's are browser only, while blobs are now cross-platform, so it's simpler to just tell the user to wrap it themselves, the `URL.createObjectURL` API is sync anyways

changed almost all references of Buffer to Uint8Array, this isn't *fully* breaking as buffer extends Uint8Array anyways, with one notable exception that Buffer.slice acts like Uint8Array.subarray


nuzzesick (approved):

jimaek:
Please note that jsDelivr is a production ready multi-cdn with failover systems as well as public download stats allowing you to track usage   https://www.jsdelivr.com/package/npm/webtorrent

And it also has esm support.

ThaUnknown:
this topic was already touched once, jsDelivr fails to bundle webtorrent, and for the module it does it uses sub-par polyfills, esm.sh handles webtorrent bundling correctly with uptime just as good as jsDelivr, it lacks stats but that's about it

jimaek:
I wouldn't call the uptime the same, it's using 1 CDN with 1 DNS provider. If they go down esm.sh goes down with them, which is fine for dev purposes. But not in case of jsDelivr that has stayed online when Cloudflare had outages thanks to production focus https://www.jsdelivr.com/network/infographic

Regarding esm we are working on improvements :)

PR #2442, opened 2023-01-22
fix: standard linting

ThaUnknown:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

migrated from standard cli to eslint, and created a custom eslint config for it
standard cli doesn't support many new features like import assertions etc, this fixes that, it also allows us to globally define globals [kek]
on the surface it might seem like this brings in a lot of new dependencies, but in reality standard already imported all these dependencies and more, so this actually reduces the amt of packages downloaded

this would have been nice to do in package.json's `standard` field, but it doesn't accept `parserOptions` as a field, which means we'd need extra config files regardless

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,0 +1,21 @@
+module.exports = {
+  env: {
+    es2022: true,
+    browser: true,
+    node: true,
+    serviceworker: true
+  },
+  extends: ['standard'],
+  parser: '@babel/eslint-parser',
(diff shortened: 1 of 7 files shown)


SilentBot1 (approved):
Looks good to me. Tests all pass locally and it makes sense to drop standard in favour of `eslint-config-standard` considering the wall hit with `standard`.

PR #2455, opened 2023-01-25
fix: redirect users to add trailing slash in server

### LINKED ISSUE #2454: Links in WebTorrent's server torrent list are incorrect
Webtorrent 2.0.0

Windows 11 22H2, node v18.13.0, npm 8.19.3

**What happened?**
When navigating to [http://localhost:9000/webtorrent](http://localhost:9000/webtorrent) without a trailing slash, clicking on Sintel will redirect you to 
`http://localhost:9000/08ada5a7a6183aae1e09d831df6748d566095a10`
 instead of 
`http://localhost:9000/webtorrent/08ada5a7a6183aae1e09d831df6748d566095a10/`
Because the url is incorrect, the server returns a 404 error.

To be redirected to `http://localhost:9000/webtorrent/08ada5a7a6183aae1e09d831df6748d566095a10/`

**Are you willing to submit a pull request to fix this bug?**
Yes

funniray:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

When a users navigates to `/webtorrent` instead of `/webtorrent/`, it sends them a 301 redirect and appends a trailing slash.

Fixes #2454

I'm unsure if there's a better way to do this. To prevent requiring absolute URLs in the `<a>` tags, this PR will redirect users. I would like to know if there would be issues with this approach.


ThaUnknown:
personally I'm not very fond of redirects like this, I'd prefer if you solved this by appending the missing trailing slash to the url

funniray:
> personally I'm not very fond of redirects like this, I'd prefer if you solved this by appending the missing trailing slash to the url

Sorry if I was slightly vague with why the direct is required.

the `href` is currently pointed to `"${escapeHtml(torrent.infoHash)}"`. Because this is a relative url, if a user browses to `http://localhost/webtorrent` without a trailing slash, they'll get redirected to `http://localhost/${torrent.infoHash}/` instead of the desired `http://localhost/webtorrent/${torrent.infoHash}/`

Whilst this could be fixed by making the `href` point to `${this.pathname}/${escapeHtml(torrent.infoHash)}`, that would require either making serveIndexPage static or passing through the pathname.

ThaUnknown:
again, this really should be fixed without redirects, passing the pathname sounds good

funniray:
Alright. I've updated it to accept the pathname. I also absolute links to the torrent page as a similar issue would occur if the user went to `/webtorrent/infohash` without a trailing slash.

nuzzesick (approved):

lib/server.js
```diff
@@ -15,11 +15,11 @@ class ServerBase {
     this.pendingReady = new Set()
   }
 
-  static serveIndexPage (res, torrents) {
+  static serveIndexPage (res, torrents, pathname) {
     const listHtml = torrents
       .map(torrent => (
       `<li>
-        <a href="${escapeHtml(torrent.infoHash)}">
+        <a href="${escapeHtml(pathname)}/${escapeHtml(torrent.infoHash)}">
```
ThaUnknown, line 22: infoHash is always hex, it doesn't need to be escaped

lib/server.js
```diff
@@ -72,11 +72,11 @@ class ServerBase {
     return res
   }
 
-  static serveTorrentPage (torrent, res) {
+  static serveTorrentPage (torrent, res, pathname) {
     const listHtml = torrent.files
       .map(file => (
     `<li>
-      <a href="./${escapeHtml(file.path)}">
+      <a href="${escapeHtml(pathname)}/${escapeHtml(torrent.infoHash)}/${escapeHtml(file.path)}">
```
ThaUnknown, line 79: same as above

lib/server.js
funniray: I only used escapeHtml as it appeared to be used everywhere in the serveIndex functions. escapeHtml was already being used on [`file.length`](https://github.com/webtorrent/webtorrent/blob/4d0267600b405aad40759e41f1b35130972ba880/lib/server.js#L82), so I assumed the intention was to use it everywhere

ThaUnknown: yeah would be better, it was there before which was kinda my bad, I had a massive workload when I was rewriting all the http server functionality, and it's still gonna be re-written again

ThaUnknown (approved):

PR #2461, opened 2023-01-27
fix: parse-torrent v11

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:


DiegoRBaquero:
This is a breaking change, but v2 is quite new and was broken on the very first versions 🤔

index.js
```diff
@@ -208,19 +208,19 @@ export default class WebTorrent extends EventEmitter {
...
    * @param  {string|Buffer|Object|Torrent} torrentId
-   * @return {Torrent|null}
+   * @return {Promise<Torrent|null>}
    */
-  get (torrentId) {
+  async get (torrentId) {
     if (torrentId instanceof Torrent) {
       if (this.torrents.includes(torrentId)) return torrentId
     } else {
+      const torrents = this.torrents
```
DiegoRBaquero, line 217: why

index.js
```diff
@@ -372,20 +372,18 @@ export default class WebTorrent extends EventEmitter {
    * @param  {string|Buffer|Torrent}   torrentId
    * @param  {function} cb
    */
-  remove (torrentId, opts, cb) {
+  async remove (torrentId, opts, cb) {
```
DiegoRBaquero, line 375: hate this mix of async and cb

docs/api.md
```diff
@@ -201,7 +201,7 @@ destroyed and all torrents are removed and cleaned up when this occurs.
 
 Always listen for the 'error' event.
 
-## `client.remove(torrentId, [opts], [function callback (err) {}])`
+## `await client.remove(torrentId, [opts], [function callback (err) {}])`
```
DiegoRBaquero, line 204: the mix of async and cb no no

index.js
ThaUnknown: this was to keep functionality mostly identical, IMHO the migration from cb -> async should all happen in a separate PR, not some version bump, I'd like to keep it like this for now and address it in a future PR, is that okay or should I remove cb in this one? [imho we really shouldnt]

index.js
ThaUnknown: since parsetorrent is now async, if someone called this function without waiting for async [happens frequently during tests] the torrent it was looking for might not longer exist, which would throw an error, here it just silent fails and returns null instead, TLDR did this so people aren't forced to use the await for this fn

DiegoRBaquero: Ok, needs to be cleaned up

DiegoRBaquero (approved):

index.js
ThaUnknown: @DiegoRBaquero that means?

DiegoRBaquero: Eventually only support `await`, as this IS an `async` function

Issue #1393, opened 2018-05-17
How webseed really works? Do we have to seed the file that has webseed url? 


hz0324:
For example, we have a magnet like this:

`magnet:?xt=urn:btih:000ee0007db5c059931c59471b3dbda97240a607&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=xxx.xx.xxx.xx/000ee0007db5c059931c59471b3dbda97240a607/Secret.png`

The last part is the weebseed url. You can just visit the url and get the file. In such case, do we have to seed the file to make this magnet link work? What will happen if there is no seeder but the webseed url is working?

I have tried to search for this question ,but nothing related found.
I also have tried to download this magnet link while webseed url is working, but no seeders for this file. It din't start the downloading. I am confused.

Please help, thanks.
Thanks.

KayleePop:
Make sure to include `http://` or `https://` at the beginning :wink:

hz0324:
@KayleePop Thank you, I didn't know that.

I have tried again. It is still not downloading if there is no seeders for this file.  Does seeder necessary?
Thanks.

KayleePop:
No, a seeder isn't required.

Try pointing to the directory instead of the file:

`&ws=http://xxx.xx.xxx.xx/000ee0007db5c059931c59471b3dbda97240a607`

feross:
Actually you still need at least one peer in order to fetch the .torrent file metadata, which is not part of the magnet link.

If you want the magnet link to work even when there are no peers, you need to also include a link to the .torrent file in the magnet link using the `xs` URL parameter.

[...]

PR #2487, opened 2023-02-20
server: Support serving the .torrent file

shesek:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

This adds support for serving the `.torrent` metadata file from the webserver via `/webtorrent/<infohash>.torrent`.

For my use-case, I was looking to use the webserver as the magnet `xs` source for the torrent file. Alongside `ws`, this [makes it possible](https://github.com/webtorrent/webtorrent/issues/1393#issuecomment-389805621) to use the webserver to create magnet links with an HTTP fallback that works even when there are no peers on the bittorrent network.


shesek:
It is still relevant :)

Issue #2441, opened 2023-01-17
Webtorrent emitting completed event to trackers multiple times


funniray:
**What version of this package are you using?**
master@https://github.com/webtorrent/webtorrent/commit/14fe83fe0088123b7caab4372217bc151094f9a3

**What operating system, Node.js, and npm version?**
Windows 11, nodejs v16.13.1, npm 8.1.2

**What happened?**
Each time a piece is stored when all selections are fulfilled (eg: removing a selection) causes WebTorrent to announce completed to the trackers. This causes completed to be announced multiple times. I believe this is caused by _checkDone not checking for currently outstanding piece reservations outside of the current selections.

**What did you expect to happen?**
For completed to only be announced once all reservations are downloaded/cancelled.

**Are you willing to submit a pull request to fix this bug?**
yes

**Steps to reproduce**
- Add a magnet URI with `&so=` appended to the end of the URI (Removes default selection)
- Add a file to a `<video>` using `.streamTo`
- Play and then pause the video to add and remove selections
- Observe the tracker to see multiple completed events being announced

gauravsaini:
https://github.com/webtorrent/webtorrent/pull/2512

PR #2512, opened 2023-03-28
 Prevent emitting completed event to trackers multiple time

gauravsaini:
Prevent emitting completed event to trackers multiple time

fixes: https://github.com/webtorrent/webtorrent/issues/2441


ThaUnknown:
hm this is a patch, not a fix, what causes this to happen in the first place i what should be looked at, done should not be called with pending reservations at all.....

PR #2542, opened 2023-05-27
perf: reduce the impact of request idle callback

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Reduces the performance impact of requestIdleCallback, this was done by checking for support only once, and calling it once, for all peers, rather than every time for every single peer. This can slightly reduce the desired effect, but it offers significant performance gains:
before:
image
after:
image
[look at relative % rather than absolute ms]

@Chocobozzz


SilentBot1 (approved):
Looks good to me.

PR #2546, opened 2023-05-28
fix: drop randombytes

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Dropped randombytes, with this PR [and the pending one on protocol] randombytes is now dropped from web version of webtorrent, bundle size decrease


SilentBot1 (approved):
LGTM!

PR #2553, opened 2023-05-30
fix: drop IDB chunk store

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Dropped hybrid chunk store, to use FSA chunk store directly, this drops IDB support.

Lets assume every1 updates their browsers, as all browsers since recently support OPFS https://caniuse.com/mdn-api_storagemanager_getdirectory putting us at 88% coverage.

actually safari doesn't support createWritable https://caniuse.com/mdn-api_filesystemfilehandle_createwritable so this is a 73% coverage :/ maybe lets leave this PR to sit for a while

FSA offers waaaaaay more storage capacity, way better speeds and way smaller bundle sizes.


ThaUnknown:
I guess we can merge this once https://bugs.webkit.org/show_bug.cgi?id=254726 is resolved, since safari is probably the biggest block on this

lib/torrent.js
```diff
@@ -58,6 +58,12 @@ const RECONNECT_WAIT = [1_000, 5_000, 15_000]
 
 const USER_AGENT = `WebTorrent/${VERSION} (https://webtorrent.io)`
 
+// if nodejs or browser that supports FSA
+const FALLBACK_STORE = !process.browser ||
+(typeof navigator !== 'undefined' && navigator.storage?.getDirectory && typeof FileSystemFileHandle !== 'undefined' && FileSystemFileHandle?.prototype?.createWritable)
+  ? FSChunkStore
+  : MemoryChunkStore
```
ThaUnknown, line 65: this is a hellscape, how to do this better?

SilentBot1: This could become:
```js
const FALLBACK_STORE = process.browser &&
  navigator?.storage?.getDirectory &&
  FileSystemFileHandle?.prototype?.createWritable
    ? FSChunkStore
    : MemoryChunkStore;
 ```
But it does assume that in the browser that `navigator` and `FileSystemFileHandle` are both defined, failing that, the undefined checks will still need to exist.

Comment (ThaUnknown) [reply lib/torrent.js:65]: this has potential to fail in node...
i considered
```js
const FALLBACK_STORE = !process.browser
  ? FSChunkStore
  : (typeof navigator !== 'undefined' && navigator.storage?.getDirectory && typeof FileSystemFileHandle !== 'undefined' && FileSystemFileHandle?.prototype?.createWritable)
    ? FSChunkStore // FSAChunkStore
    : MemoryChunkStore

Comment (ThaUnknown) [reply lib/torrent.js:65]: or a separate 
```js
const SUPPORTS_FSA = (typeof navigator !== 'undefined' && navigator.storage?.getDirectory && typeof FileSystemFileHandle !== 'undefined' && FileSystemFileHandle?.prototype?.createWritable)

const FALLBACK_STORE = !process.browser
  ? FSChunkStore // FSChunkStore - nodejs/bun
  : SUPPORTS_FSA
    ? FSChunkStore // FSAChunkStore - browser
    : MemoryChunkStore
```

SilentBot1: imo separate would be better at that point

SilentBot1 (approved):

PR #2564, opened 2023-06-05
fix: utp support bundlers and preprocessors

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

utp needs to conditionally import a module, ESM doesn't support this synchronously, and we need to do it sync
we could avoid .cjs by using node:module but that doesn't sit well with preprocessors like typescript and bundlers which can bundle for node like webpack

this is a bit lazy but works well, and doesn't require us to rewrite all of webtorrent to support async imports


SilentBot1 (approved):
Hopefully we don't need this workaround for too long, as while it gets the job done, it isn't very elegant.

PR #2565, opened 2023-06-05
fix: correctly destroy piped streams in server

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

This fixes an issue which caused us to leak iterators when using node's server

specifics are outlined in https://github.com/mafintosh/streamx/issues/76

since pump was no longer needed it was removed


SilentBot1 (approved):
Looks good to me.

PR #2570, opened 2023-06-06
Implement NAT traversal

### LINKED ISSUE #195: UPnP and NAT-PMP port forwarding

SilentBot1:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

The goal of this PR is to increase discoverability and reachability for NodeJS based clients who wish to receive inbound TCP and UDP connections from peers, the DHT, or optionally to the created web server. This is achieved by implementing NAT traversal into WebTorrent, supporting both the UPNP and PMP methods for port forwarding via the `@silentbot1/nat-api` package. 

This supports chromeapps, but only UPNP, PMP isn't supported as we couldn't figure out how to obtain the default gateway in a chromeapp.

fixes #195

- Testing on a wide range of networks. UPNP and PMP are typically supported by a wide range of routers, and while this has worked in my test environment, ensuring it works on others is important!
- Testing client destruction and recreation. The client will attempt to close all open ports on destruction, which could possibly cause issues when destroying / creating the client rapidly.


index.js
```diff
@@ -75,13 +76,17 @@ export default class WebTorrent extends EventEmitter {
...
+    this.natUpnp = opts.natUpnp || true
+    this.natPmp = opts.natPmp || true
     this.torrents = []
     this.maxConns = Number(opts.maxConns) || 55
     this.utp = WebTorrent.UTP_SUPPORT && opts.utp !== false
 
     this._downloadLimit = Math.max((typeof opts.downloadLimit === 'number') ? opts.downloadLimit : -1, -1)
     this._uploadLimit = Math.max((typeof opts.uploadLimit === 'number') ? opts.uploadLimit : -1, -1)
 
+    this.natTraversal = this.natUpnp ? new natApi({ enablePMP: this.natPmp }) : null
```
ThaUnknown, line 88: as github suggests `NatAPI`, not `natApi`

ThaUnknown: `this.natUpnp && new NatAPI` not `?`

lib/server.js
```diff
@@ -263,6 +266,7 @@ class NodeServer extends ServerBase {
   constructor (client, opts) {
     super(client, opts)
 
+    this.nat = opts.nat || false
```
ThaUnknown, line 269: honestly, we dont need nat in the http server

lib/server.js
```diff
@@ -313,12 +317,31 @@ class NodeServer extends ServerBase {
     this.closed = false
     this.server.on('connection', this.onConnection.bind(this))
     this.server.on('request', this.wrapRequest.bind(this))
+    if (this.client.natTraversal != null && this.nat === true) {
+      this.server.on('listening', () => {
+        try {
+          const port = this.server.address().port
```
ThaUnknown, line 323: this wont actually work, as at this time the port isnt defined, its defined a few lines later on the this._listen.apply

docs/api.md
```diff
@@ -82,6 +84,10 @@ For possible values of `opts.tracker` see the
 For possible values of `opts.blocklist` see the
 [`load-ip-set` documentation](https://github.com/webtorrent/load-ip-set#usage).
 
+For `opts.natUpnp` and `opts.natPmp`, if both are set to `true`, PMP will be attempted first, then fallback to UPNP.
```
ThaUnknown, line 87: NodeJS only.

index.js
```diff
@@ -482,7 +504,16 @@ export default class WebTorrent extends EventEmitter {
     if (this._connPool) {
       // Sometimes server.address() returns `null` in Docker.
       const address = this._connPool.tcpServer.address()
-      if (address) this.torrentPort = address.port
+      if (address) {
+        this.torrentPort = address.port
+        if (this.natTraversal != null) {
+          try {
+            this.natTraversal.map({ publicPort: this.torrentPort, privatePort: this.torrentPort, protocol: this.utp ? null : 'tcp', description: 'WebTorrent Torrent' })
```
ThaUnknown, line 511: this is async, so yo need a .catch() not a try-catch unless you make this function async and use await

index.js
```diff
@@ -123,7 +128,16 @@ export default class WebTorrent extends EventEmitter {
 
       this.dht.once('listening', () => {
         const address = this.dht.address()
-        if (address) this.dhtPort = address.port
+        if (address) {
+          this.dhtPort = address.port
+          if (this.natTraversal != null) {
+            try {
+              this.natTraversal.map({ publicPort: this.dhtPort, privatePort: this.dhtPort, protocol: 'udp', description: 'WebTorrent DHT' })
```
ThaUnknown, line 135: this is async, so yo need a .catch() not a try-catch unless you make this function async and use await

lib/server.js
SilentBot1: Updated to remove NAT for the HTTP server.

lib/server.js
SilentBot1: Thanks for spotting this, as per the above comment, NAT has been removed for the HTTP server.

index.js
SilentBot1: Both have been updated, thanks.

docs/api.md
SilentBot1: The docs have been amended to note this is NodeJS only.

index.js
SilentBot1: Thanks for catching this, my head clearly wasn't on straight when writing this. This has been amended to .catch the error.

index.js
SilentBot1: This has been amended to .catch the error as per the above comment.

index.js
```diff
@@ -15,6 +15,7 @@ import { ThrottleGroup } from 'speed-limiter'
 import ConnPool from './lib/conn-pool.js' // browser exclude
 import Torrent from './lib/torrent.js'
 import { NodeServer, BrowserServer } from './lib/server.js'
+import NatAPI from '@silentbot1/nat-api' // browser exclude
```
ThaUnknown, line 18: cosmetic - library imports then local imports so this should be above any `somepath/file.js` import

ThaUnknown:
question: I understand the mappings get removed from my router if I call destroy() but what about if i just kill the node process, do the mappings stay there until the router is restarted? how is it handled? if they stay there maybe be should disable permanent mappings by default, and have them be opt-in?

ThaUnknown:
now the dependency, these are just code quality pickings, NOT ACTUAL PROBLEMS!, but it would be nice to resolve them:

Biggest gripe: never use `return await value` instead do `return value` its functionally IDENTICAL, while being faster.

https://github.com/SilentBot1/nat-api/blob/master/lib/pmp/index.js#L65-L85
these functions are async when they shouldn't be.... well kinda
portMapping needs to stay async because it has a `throw`, if you were to remove the async you wouldn't be able to use .catch(), however if you don't use .catch() for it, instead wrap it in a try {} catch, you should remove the async and await, since the functions synchronously return a promise, it's functionality would be the same as if they were async, example:
```js
async function stuff () {
  return await doWork()
}
```
is the same as:
```js
function stuff () {
  return doWork()
}
```
except you wrap it in less promises which is a lot quicker, however, this isn't the same:
```js
async function stuff () {
  if (fail) throw err
  return await doWork()
}
```
isn't the same as:
```js
function stuff () {
  if (fail) throw err
  return doWork()
}
```
because on the latter you can't do a .catch() as it can synchronously return an error, which is why you'd need a try{}catch instead to make it the same


https://github.com/SilentBot1/nat-api/blob/master/lib/upnp/device.js#L53-L83
MY EYES, this really should be a map, could be smth along the lines of, however this is my personal opinion, not backed by any best practices, so take this with a massive grain of salt:
```js
if (error) {
  throw new Error(errorMap[Number(errorCode)] || defaultErrorString)
}
```

ThaUnknown:
actual issues:
what the fuck is this? https://github.com/SilentBot1/nat-api/blob/master/lib/upnp/ssdp.js#L49-L57

should this be called repeatedly? https://github.com/SilentBot1/nat-api/blob/master/index.js#L35

can a gateway in this case change? can we instead just cache this value? this library calls `child_proces.exec` so if possible I'd like to avoid calling it as much as possible. I guess if the network changes the default gateway can change, is that right?

additionally I feel like doing this sync, blocks the entire thread, meaning all of webtorrent, and it has potential to be insanely slow, I wouldn't be surprised if it was upwards of 200ms so I feel like calling the async version would be better, however that means an async callback and handling of destroyed state

index.js
```diff
@@ -75,13 +76,22 @@ export default class WebTorrent extends EventEmitter {
...
+    this.natUpnp = opts.natUpnp !== true
+    this.natPmp = opts.natPmp !== true
     this.torrents = []
     this.maxConns = Number(opts.maxConns) || 55
     this.utp = WebTorrent.UTP_SUPPORT && opts.utp !== false
 
     this._downloadLimit = Math.max((typeof opts.downloadLimit === 'number') ? opts.downloadLimit : -1, -1)
     this._uploadLimit = Math.max((typeof opts.uploadLimit === 'number') ? opts.uploadLimit : -1, -1)
 
+    if ((this.natUpnp || this.natPmp) && typeof NatAPI === 'function') {
```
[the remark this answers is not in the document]
SilentBot1: By default the bundler returns an empty object which is truthy:
image

This `typeof` check matches how other optionally excluded requirements are already checked, e.g:

https://github.com/webtorrent/webtorrent/blob/cb31d42e39d51ab7f47cd36551e5370843b06b1b/index.js#L114
https://github.com/webtorrent/webtorrent/blob/cb31d42e39d51ab7f47cd36551e5370843b06b1b/index.js#L126

index.js
```diff
@@ -75,13 +77,22 @@ export default class WebTorrent extends EventEmitter {
     this.tracker = opts.tracker !== undefined ? opts.tracker : {}
     this.lsd = opts.lsd !== false
     this.utPex = opts.utPex !== false
+    this.natUpnp = opts.natUpnp !== true
```
ThaUnknown, line 80:
```suggestion
this.natUpnp = opts.natUpnp ?? true
```

index.js
```diff
@@ -75,13 +77,22 @@ export default class WebTorrent extends EventEmitter {
...
+    this.natPmp = opts.natPmp !== true
```
ThaUnknown, line 81:
```suggestion
    this.natPmp = opts.natPmp ?? true
```

lib/server.js
```diff
@@ -319,6 +319,7 @@ class NodeServer extends ServerBase {
   close (cb = () => {}) {
     this.server.removeAllListeners('connection')
     this.server.removeAllListeners('request')
+    this.server.removeAllListeners('listening')
```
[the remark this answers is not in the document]
SilentBot1: We do listen to the `listening` event ourselves, though I assume when you implemented this, it was to handle if this was called elsewhere, I can remove this if it feels unnecessary?

ThaUnknown:
I'd probably specify this in `opts.natUpnp` and have it accept 2 values, `permanent` and `non-permanent`, if opts.natUpnp has ANY value set that's not falsy, so: 1, true, {}, 'xd' it will treat it as non-permanent and if its set to exactly `'permanent'` it would try doing the fallback so the check would look like
```js
if (opts.natUpnp) {
  const mappingResult = tryNonPermanent()
  if (!mappingResult && opts.natUpnp === 'permanent') {
    tryPermanent()
  }
}
```

docs/api.md
```diff
@@ -82,6 +110,12 @@ For possible values of `opts.tracker` see the
...
+NodeJS only.
```
ThaUnknown, line 114: remove newline

SilentBot1: Resolved.

docs/api.md
```diff
@@ -82,6 +110,12 @@ For possible values of `opts.tracker` see the
...
+
+For `opts.natUpnp`, if set to `permanent`, a permanent ttl will be used for UPNP if the router only supports permanent
+leases. NodeJS only.
+
```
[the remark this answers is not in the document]
SilentBot1: That reads a lot better, thank you, this has now been implemented.

ThaUnknown (approved):

PR #2585, opened 2023-06-15
fix: Revert "fix: correctly destroy piped streams in server"

ThaUnknown:
Reverts webtorrent/webtorrent#2565

makes stream/iterator check less strict


SilentBot1 (approved):

ThaUnknown:
at this point it feels obligatory i fuck up every 3-4 months

PR #2607, opened 2023-07-05
fix: perf: faster stream resolution

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

This fixes an issue that caused iterators to sometimes hang for a very long time, and also cleans up some code.

The notify() function wasn't called for each piece [I think....] which caused the streams to sometimes hang for a very long time before being resolved, this was simply annoying and I shiver at the thought of how long this issue has existed, since this code was originally copied from file streams.

This creates a lot of event calls since it's now done for all pieces, not just the ones we're interested in, but speeds up streams a lot.


SilentBot1 (commented):
Looks good to me, though we should probably add `torrent.on('verified', (index)=>{})` as a new event in the API docs before merging.

SilentBot1 (approved):

Issue #2685, opened 2023-10-17
`add` and `remove` are not async safe


uriva:
**What version of this package are you using?**
^2.1.27

**What operating system, Node.js, and npm version?**

ubuntu 23.4
v19.0.0
8.19.2

**What happened?**
It's possible to get `add` and `remove` to throw if calling them in parallel, even with different hash ids.

**What did you expect to happen?**

`remove` should not throw if it can't find the torrent.

`add` should not throw if the torrent is already added.

**Are you willing to submit a pull request to fix this bug?**
yes

ThaUnknown:
those are async functions, you simply need to run
```js
await client.add()
```
and
```js
await client.remove()
```

uriva:
consider the case where `add` is called from different places in the code. If there is an sync context switch inside `add`, then we can be inside it twice.

`await` is not equivalent to a lock.

ThaUnknown:
fair

uriva:
while we're at it, these functions should probably not have a cb parameter and be async at the same time.

ThaUnknown:
they should be, but not as a part of this issue

this was planned for ages: https://github.com/orgs/webtorrent/projects/1#card-60702582

they became async because other issues were being blocked by it, and dropping callback based calls for webtorrent, means doing it inside close to 20 packages, which requries a lot of testing and rewriting

PR #2689, opened 2023-10-23
make `get` and `remove` synchronous

uriva:
improve code health and avoid misleading developers

simplify some functions and change them to be synchronous

https://github.com/webtorrent/webtorrent/issues/2685 (partly, more work will follow)


uriva:
now I see that parseTorrent has wrong typing, and is actually `async`...

ThaUnknown:
yup, blame definetlyTyped for that, they still have not updated webtorrent types

PR #2691, opened 2023-10-24
Don't throw arbitrary exceptions, added an api method to add with timeout

uriva:


index.js
```diff
@@ -294,6 +304,12 @@ export default class WebTorrent extends EventEmitter {
     return torrent
   }
 
+  addV2 (torrentId, timeout) {
```
krnl0138, line 307: May be better to provide comment & typings

index.js
```diff
@@ -294,6 +304,12 @@ export default class WebTorrent extends EventEmitter {
...
+    return new Promise((resolve, reject)=> {
```
krnl0138, line 308: missing space

uriva: Does this project include an auto formatting setting I've missed or you do the spacing manually?

krnl0138: There're no formatter or eslint rules included in package.json. I assume then you have to adjust to the code style manually

uriva: This would require a lot of effort going forward. Why is this the case?

ThaUnknown: there is, standard which extends eslint

uriva: I'm unfamiliar with a formatter named "standard" or using eslint as a formatter.
can you link to some instructions about how to set this up?

uriva: not sure why you're not answering. I can also use my formatter to format only the lines I've touched but I'm not going to do manual style corrections, this is not 1999 we have better things to do with our time.

uriva:
```
$ npx standard --fix index.js 
standard: Use JavaScript Standard Style (https://standardjs.com)
  /home/uri/uriva/webtorrent/index.js:20:35: Parsing error: Unexpected token assert (null)
```

krnl0138: Check here: https://standardjs.com/rules

If you haven't setup your code editor to check linting on save, you may run `eslint` on files you've been working on to execute rules which standard provides. I believe it will do the trick. I guess these actions could be automated on MR opening with github actions. Yet I'm not sure

Anyway it's still your responsibility to match your code with the code base you're trying to merge to. Prettier is too opinionated and not an industry standard. I may understand why it's not included in the first place

uriva: Again, I don't care if it's prettier or not, I care that it's not manual.

Running standard does not work with your code and eslint --fix causes no changes.

SilentBot1: @uriva I assume this is caused  by you having a different version of `eslint` installed globally vs what is defined in the `package.json`, causing it to possibly not read the `.eslintrc.cjs`.  Does running `npx eslint@8.51.0 .` and `npx eslint@8.51.0 --fix .` while in the root of the project work for you?

uriva: that works ty!

uriva: my version was 8.52.0

index.js
uriva: ptal

index.js
uriva: fyi it wasn't even the version, just missing the `.`

PR #2733, opened 2024-01-12
fix: malformed debug ID

EvanHahn:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

You can see debug messages from WebTorrent with the `DEBUG` environment variable. For example, in this repository:

```sh
DEBUG=webtorrent npm test
```

Debug messages should contain the first 7 characters of the peer ID. However, they currently look like this:

    webtorrent [02undef] listening +7ms

"02undef" seems wrong. After this change, the debug ID looks correct:

    webtorrent [2d57573] listening +7ms

This happened because `peerId` is a string but was mistakenly being treated like a buffer.

No GitHub issues I'm aware of.

No.

### DIFF
--- a/index.js
+++ b/index.js
@@ -67,7 +67,7 @@ export default class WebTorrent extends EventEmitter {
     }
     this.nodeIdBuffer = hex2arr(this.nodeId)
 
-    this._debugId = arr2hex(this.peerId).substring(0, 7)
+    this._debugId = this.peerId.substring(0, 7)
 
     this.destroyed = false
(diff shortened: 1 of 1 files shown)


ThaUnknown:
right, in v1 this did `buffer.toString().toString()` which when I was rewriting in v2 i assumed to be `buffer.toString()` when it was actually a string, good catch

ThaUnknown (approved):

PR #2750, opened 2024-02-09
fix: iterate over destroyed torrent

puttyman:

[ ] Documentation update
[X ] Bug fix
[ ] New feature
[ ] Other, please explain:

Fix an issue when iterating over a file could cause the library to crash due to an unhandled error in promise. This change adds a try/catch around `this._torrent.critical(index, index + this._criticalLength)` which can throw an error if the torrent is destroyed.

There is no open issues related to this one.

This issue would occur when a file read is made while a torrent is being destroyed.

### DIFF
--- a/lib/file-iterator.js
+++ b/lib/file-iterator.js
@@ -51,7 +51,11 @@ export default class FileIterator extends EventEmitter {
             }
           }
           this._torrent.on('verified', listener)
-          return this._torrent.critical(index, index + this._criticalLength)
+          try {
+            return this._torrent.critical(index, index + this._criticalLength)
+          } catch (err) {
+            reject(err)
+          }
         }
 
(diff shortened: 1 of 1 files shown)


puttyman:
Hey @ThaUnknown ,

Thanks for the feedback. I had a hard time trying to reproduce this issue. It appears it occurs mostly when there is an active read to a file while a torrent destroy has been initiated.

I'll see if I can dig this out further and provide more info. One note with this fix the issue does not surface again.

ThaUnknown:
yeah webtorrent isn't simple, finding edge cases like this is insanely hard

puttyman:
I am closing this pull request as I could not see it happening more often. It seems this issue occurred a few times when the host is under pressure (limited RAM and CPU resource left)

Issue #2755, opened 2024-02-17
Webtorrent downloads deselected pieces after re-adding a torrent


detarkende:
**What version of this package are you using?**
2.1.36

**What operating system, Node.js, and npm version?**
macOS Sonoma 14.2.1 (23C71)
Node.js v20.11.1

**What happened?**
I'm using the free Sintel torrent - downloaded from webtorrent.io to test partial downloading and seeding.
Here's what I do:
```js
/** @type {WebTorrent.Torrent} */
const torrent = await new Promise((resolve, reject) => {
    const torrent = client.add(
        TORRENT_PATH,
        {
            path: DOWNLOADS_PATH,
        },
        (torrent) => {
            console.log(
                `Torrent ${torrent.name} - ${torrent.infoHash} verified and added.`,
            );
            resolve(torrent);
        },
    );
    torrent.on('metadata', () => {
        console.log(
            `Deselecting all files and pieces for ${torrent.name} - ${torrent.infoHash}`,
        );
        torrent.files.forEach((file) => file.deselect());
        torrent.deselect(0, torrent.pieces.length - 1, 0);
    });
    torrent.on('error', reject);
});
```
I add the torrent, deselect all files and pieces, then wait for the file verification to finish.
After this, I have a server endpoint that serves partial content. I grab the start and end pieces from the Range header, then use those to create the readable stream:
```js
/** @type {ReadableStream} */
const fileStream = file.stream({start, end});
const nodeStream = Readable.fromWeb(fileStream);
nodeStream.pipe(res);
```
This means that only the content that the user requests is actually downloaded, the rest of the files should not get downloaded. Meanwhile, I should be a partial seeder if possible, to seed as much as I can.

**everything up until this point works correctly.**

The issue happens when I restart my script:
Webtorrent verified pieces (I log out the progress on each `download` event), but then proceeds to download the rest of the torrent. Even though I have deselected everything and didn't create a ReadableStream to the files yet. 
Straight after adding the torrent, it starts downloading the full content.

**What did you expect to happen?**
I would expect Webtorrent not to start the download after the verification, since everything should be deselected.

**Are you willing to submit a pull request to fix this bug?**
Yes, although I am not very familiar with the Bittorrent protocol, so I might need some guidance.

<details>
<summary>Here's the complete script, in case it is useful to you:</summary>

```js
import WebTorrent from 'webtorrent';
import Express from 'express';
import mime from 'mime';
import { Readable } from 'stream';

[...]

Issue #2594, opened 2023-06-17
Enabling protocol encryption stalls downloads


anma-dev:
**What version of this package are you using?**
webtorrent version 2.1.9

**What operating system, Node.js, and npm version?**
The problem occurs on macOS and Linux as far as I know.
Node 19.4.0
Npm 9.7.1

**What happened?**
This issue is related to bittorrent protocol encryption. When enabling protocol encryption the torrent never starts downloading and it hangs indefinitely.

Steps to installation

- npm init
- npm install webtorrent network-address

the code:

`````
#!/usr/bin/env node
import WebTorrent from "webtorrent";
import networkAddress from "network-address";

const torrentID = "...";
const client = new WebTorrent({
  secure: true,
});

client.on("error", (err) => {
  console.log(err);
});

function startServer(torrent) {
  if (torrent.ready) startTorrent(torrent);
  else torrent.once("ready", () => startTorrent(torrent));
}

function startTorrent(torrent) {
  // start the streaming torrent-to-http server
  let server = client.createServer();
  server.listen(0, () => {
    console.log("server listen");
    const port = server.address().port;
    const urlSuffix = ":" + port;
    const info = {
      torrentKey: torrent.key,
      localURL: "http://localhost" + urlSuffix,
      networkURL: "http://" + networkAddress() + urlSuffix,
      networkAddress: networkAddress(),
    };
    console.log(info.networkAddress);
  });
}

const torrent = client.add(torrentID);

startServer(torrent);

`````

I think the problem is with this while loop that never ends and is always waiting for setGenerators to be set to true, which never happens: https://github.com/webtorrent/bittorrent-protocol/blob/v4.1.7/index.js#L1157

A potential solution is to change this line at https://github.com/webtorrent/webtorrent/blob/v2.1.5/lib/peer.js#L130 for:

`````
        if (secure && this.retries === 0 && !this.sentPe1) {
          this.sendPe1();
          this.wire._determineHandshakeType() // added
      }
`````

I'm not sure why this works but the script no longer stalls. I'm not sure how to check that encryption is actually being used after this change.

**What did you expect to happen?**
Specifying the secure option should enable protocol encryption without interrupting the download of torrented content.

**Are you willing to submit a pull request to fix this bug?**
Yes

ThaUnknown:
oops.

might be my fault, i dont think we have tests for this which is an issue too

ThaUnknown:
[...]

Issue #2581, opened 2023-06-13
perf: use uint8.subarray instead of uint8.slice


ThaUnknown:
when migrating to uint8 from node buffer, code which used .slice was not changed, node:buffer's slice creates a subarray, but uint8.slice creates a copy, this needs to be updated to use .subarray, this would be another very big performance gain

ljluestc:
// Original code with Node.js Buffer
const originalBuffer = Buffer.from(/* some data */);
const subBuffer = originalBuffer.slice(10, 20); // Creates a subarray

// Updated code using Uint8Array
const uint8Array = new Uint8Array(/* some data */);
const subArray = uint8Array.subarray(10, 20); // Creates a view without copying

Issue #2467, opened 2023-01-28
Abstract web seed


ThaUnknown:
The issue of webseeds having poor performance has been raised multiple times in many ways, this is mainly because it treats webseeds as peers, not webseeds

Banou26:
> Notable issues are that these would likely need to bypass speed limiter, otherwise it would simply keep the data in memory as browser http streams don't really care if the request stream request is requesting more data or not, it just pulls everything at max speed

For servers that respond with a `'Cache-Control': 'no-store'` header, the fetch's stream's backpressure is properly handled, as per
https://bugs.chromium.org/p/chromium/issues/detail?id=480746 on chromium
and
https://bugzilla.mozilla.org/show_bug.cgi?id=1808868 (current issue, recently opened)

You can try it out yourself here https://github.com/Banou26/webtorrent-issue-2467 or here's a video otherwise of that same repo in action on chrome 

https://user-images.githubusercontent.com/5209149/215358312-3aa5eb93-5d67-4da7-8660-d0edd02a08b0.mp4

ThaUnknown:
started work at https://github.com/webtorrent/webtorrent/tree/abstract-web-seed

PR #2757, opened 2024-02-28
Refactor selections with non-overlapping data structure

detarkende:

- [X] Documentation update
- [X] Bug fix
- [ ] New feature
- [ ] Other, please explain:

### WIP - please give me feedback

The goal of this PR is to introduce a new data structure that handles piece selections. The new `Selection` class allows you to add intervals to the selection and remove them without creating overlapping selection items.

Upon insertion, the class checks if the code overlaps with any existing selections. If it does, then it merges the two selections (the new, inserted selection remains intact and the existing one(s) are truncated or cut into pieces).

This also means that (for example) if the user runs
```js
torrent.deselect(0, torrent.pieces.length - 1);
```
then all the pieces are actually deselected.

The long-term goal of this change is to also fix the issue with `file.deselect()` but that's a story for a later day.

It relates to #2755.

As Webtorrent is a pretty large project, I don't understand all of it. I would like reviewers to help me poke holes into this implementation. If there are any aspects of the app that this breaks that I didn't think of, please let me know so I can hopefully extend my solution.

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -91,7 +91,7 @@ export default class File extends EventEmitter {
 
   deselect () {
     if (this.length === 0) return
-    this._torrent.deselect(this._startPiece, this._endPiece, false)
+    this._torrent.deselect(this._startPiece, this._endPiece)
   }
 
   [Symbol.asyncIterator] (opts = {}) {
--- a/lib/selection.js
+++ b/lib/selection.js
@@ -5,6 +5,7 @@
  * @property {number} offset
  * @property {number} priority
  * @property {function} notify
+ * @property {boolean} isStreamSelection
  */
 
 export class Selection {
@@ -17,10 +18,27 @@ export class Selection {
   /**
    * @public
    * @param {{from: number, to: number}} item Interval to be removed from the selection
+   * @param {boolean} [isStreamSelection]
    */
-  remove (item) {
+  remove (item, isStreamSelection = false) {
     for (let i = 0; i < this._items.length; i++) {
       const existing = this._items[i]
+
+      if (existing.isStreamSelection) {
+        if (isStreamSelection) {
+          // If both are stream selections and they match, then we remove the first matching item, then we break the loop
+          if (existing.from === item.from && existing.to === item.to) {
+            this._items.splice(i, 1)
+            // for stream selections, we only remove one item at a time
+            // ergo we break the loop after removing the first matching item
+            break
+          }
+        } else {
+          // we only remove stream selections when the `isStreamSelection` flag is true and they match
+          continue
+        }
+      }
+
       if (isLowerIntersecting(item, existing)) {
         existing.to = item.from - 1
         continue
@@ -46,13 +64,16 @@ export class Selection {
   /**
    * @public
    * @param {SelectionItem} newItem
+   * @param {boolean} [isStreamSelection]
    */
-  insert (newItem) {
+  insert (newItem, isStreamSelection = false) {
     if (newItem.from > newItem.to) {
       throw new Error('Invalid interval')
     }
-    this.remove(newItem)
-    this._items.push(newItem)
+    if (!isStreamSelection) {
+      this.remove(newItem)
+    }
+    this._items.push({ ...newItem, isStreamSelection })
   }
 
   /**
--- a/lib/selection.js
+++ b/lib/selection.js
@@ -1,0 +1,181 @@
+/** A selection of pieces to download.
+ * @typedef {Object} SelectionItem
+ * @property {number} from
+ * @property {number} to
+ * @property {number} offset
+ * @property {number} priority
+ * @property {function} notify
+ */
+
+export class Selection {
+  /**
+   * @type {Array<SelectionItem>}
+   * @private
+   */
+  _items = []
+
+  /**
+   * @public
+   * @param {{from: number, to: number}} item Interval to be removed from the selection
+   */
+  remove (item) {
+    for (let i = 0; i < this._items.length; i++) {
+      const existing = this._items[i]
+      if (isLowerIntersecting(item, existing)) {
+        existing.to = item.from - 1
+        continue
+      }
+      if (isUpperIntersecting(item, existing)) {
+        existing.from = item.to + 1
+        continue
+      }
+      if (isInsideExisting(item, existing)) {
+        const existingStart = { ...existing, to: item.from - 1 }
+        const existingEnd = { ...existing, from: item.to + 1 }
+        this._items.splice(i, 1, existingStart, existingEnd)
+        continue
+      }
+      if (isCoveringExisting(item, existing)) {
+        this._items.splice(i, 1)
+        i -= 1
+        continue
+      }
+    }
+  }
+
+  /**
+   * @public
+   * @param {SelectionItem} newItem
+   */
+  insert (newItem) {
+    if (newItem.from > newItem.to) {
+      throw new Error('Invalid interval')
+    }
+    this.remove(newItem)
+    this._items.push(newItem)
+  }
+
+  /**
+   * @public
+   * @param {(a: SelectionItem, b: SelectionItem) => number} sortFn
+   */
+  sort (sortFn = (a, b) => a.from - b.from) {
+    this._items.sort(sortFn)
+  }
+
+  /**
+   * @public
+   */
+  isEmpty () {
+    return this._items.length === 0
+  }
+
+  /**
+   * @public
+   */
+  get length () {
+    return this._items.length
+  }
+
+  /**
+   * Takes an index and returns the item at that index
+   * @public
+   * @param {number} index
+   */
+  get (index) {
+    return this._items[index]
+  }
+
+  /**
+   * Takes an index and an item and replaces the item at the index with the new item
+   * @public
+   * @param {number} index
+   * @param {SelectionItem} item
+   */
+  set (index, item) {
+    this._items[index] = item
+  }
+
+  /**
+   * Takes two indexes and swaps the items at those indexes
+   * @public
+   * @param {number} i
+   * @param {number} j
+   */
+  swap (i, j) {
+    const temp = this._items[i]
+    this._items[i] = this._items[j]
+    this._items[j] = temp
+  }
+
+  /**
+   * @public
+   */
+  clear () {
+    this._items.length = 0
+  }
+
+  * [Symbol.iterator] () {
+    const self = this
+    for (let i = 0; i < this._items.length; i++) {
+      const item = this._items[i]
+      yield {
+        ...item,
+        remove () {
+          self._items.splice(i, 1)
+          i -= 1
+        }
+      }
+    }
+  }
+}
+
+/**
+ * Examples:
+ * existing: 1-10 | 1-10
+ * new_item: 8-12 | 10-15
+ * @param {SelectionItem} newItem
+ * @param {SelectionItem} existing
+ * @returns {boolean} returns true if the new item's lower end is intersecting with the existing item
+ */
+export function isLowerIntersecting (newItem, existing) {
+  return (newItem.from <= existing.to) && (newItem.from > existing.from) && (newItem.to > existing.to)
+}
+
+/**
+ * Examples:
+ * existing: 20-25 | 20-25
+ * new_item: 15-22 | 15-20
+ * @param {SelectionItem} newItem
+ * @param {SelectionItem} existing
+ * @returns {boolean} returns true if the new item's upper end is intersecting with the existing item
+ */
+export function isUpperIntersecting (newItem, existing) {
+  return (newItem.to >= existing.from) && (newItem.to < existing.to) && (newItem.from < existing.from)
+}
+
+/**
+ * Examples:
+ * existing: 10-20 | 10-20 | 10-20
+ * new_item: 12-15 | 20-20 | 15-20
+ * @param {SelectionItem} newItem
+ * @param {SelectionItem} existing
+ * @returns {boolean} returns true if the new item is completely inside the existing item
+ */
+export function isInsideExisting (newItem, existing) {
+  const existingIntervalSize = existing.to - existing.from
+  const newItemIntervalSize = newItem.to - newItem.from
+  return newItem.from >= existing.from && newItem.to <= existing.to && (newItemIntervalSize < existingIntervalSize)
+}
+
+/**
+ * Examples:
+ * existing: 10-20 | 10-20 | 10-20
+ * new_item: 10-21 | 09-20 | 10-20
+ * @param {SelectionItem} newItem
+ * @param {SelectionItem} existing
+ * @returns {boolean} returns true if the new item is covering the existing item
+ */
+export function isCoveringExisting (newItem, existing) {
+  return newItem.from <= existing.from && newItem.to >= existing.to
+}
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -20,7 +20,7 @@ export class Selections {
   _items = []
 
   /**
-   * @param {SelectionItem} item Interval to be removed from the selection
+   * @param {MinimalSelectionItem & {isStreamSelection: boolean, priority?: number}} item Interval to be removed from the selection
    */
   remove (item) {
     for (let i = 0; i < this._items.length; i++) {
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -8,16 +8,20 @@
  * @typedef {MinimalSelectionItem & {
  *  offset: number,
  *  priority: number,
- *  notify: function,
  *  isStreamSelection?: boolean
  * }} SelectionItem
  */
 
+/**
+ * @typedef {MinimalSelectionItem & {notify: function}} NotificationItem
+ */
+
 export class Selections {
-  /**
-   * @type {Array<SelectionItem>}
-   */
+  /** @type {Array<SelectionItem>} */
   _items = []
+
+  /** @type {Array<NotificationItem>} */
+  _notifications = []
 
   /**
    * @param {MinimalSelectionItem & {isStreamSelection: boolean}} item Interval to be removed from the selection
@@ -67,13 +71,43 @@ export class Selections {
    * @param {SelectionItem} newItem
    */
   insert (newItem) {
-    if (newItem.from > newItem.to) {
+    const { from, to, notify, isStreamSelection, ...restItem } = newItem
+    if (from > to) {
       throw new Error('Invalid interval')
     }
-    if (!newItem.isStreamSelection) {
-      this.remove(newItem)
+    if (!isStreamSelection) {
+      this.remove({ from, to, isStreamSelection })
     }
-    this._items.push({ ...newItem })
+    this._items.push({ from, to, isStreamSelection, ...restItem })
+    this._notifications.push({ from, to, notify })
+  }
+
+  /**
+   * Returns all the notifications that in any way intersect with the given range
+   * @param {number} from
+   * @param {number} to
+   */
+  getNotificationsByRange (from, to) {
+    const range = { from, to }
+    return this._notifications.filter(notification => {
+      const isNotEmpty = (notification.to - notification.from) > 0
+      const isIntersecting = isCoveringExisting(range, notification) ||
+        isInsideExisting(range, notification) ||
+        isLowerIntersecting(range, notification) ||
+        isUpperIntersecting(range, notification)
+      return isNotEmpty && isIntersecting
+    })
+  }
+
+  /**
+   * Removes a notification from the list of notifications by reference
+   * @param {NotificationItem} notification
+   */
+  removeNotification (notification) {
+    const index = this._notifications.findIndex(n => n === notification)
+    if (index !== -1) {
+      this._notifications.splice(index, 1)
+    }
   }
 
   /**
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -29,7 +29,7 @@ export class Selections {
       if (existing.isStreamSelection) {
         if (item.isStreamSelection) {
           // If both are stream selections and they match, then we remove the first matching item, then we break the loop
-          if (existing.from === item.from && existing.to === item.to) {
+          if (existing.from === item.from && existing.to === item.to && existing.priority === item.priority) {
             this._items.splice(i, 1)
             // for stream selections, we only remove one item at a time
             // ergo we break the loop after removing the first matching item
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -5,12 +5,12 @@
  */
 
 /** A selection of pieces to download.
- * @typedef {Object} SelectionItem
- * @extends MinimalSelectionItem
- * @property {number} offset
- * @property {number} priority
- * @property {function} notify
- * @property {boolean} isStreamSelection
+ * @typedef {MinimalSelectionItem & {
+ *  offset: number,
+ *  priority: number,
+ *  notify: function,
+ *  isStreamSelection?: boolean
+ * }} SelectionItem
  */
 
 export class Selections {
@@ -20,15 +20,14 @@ export class Selections {
   _items = []
 
   /**
-   * @param {{from: number, to: number}} item Interval to be removed from the selection
-   * @param {boolean} [isStreamSelection]
+   * @param {SelectionItem} item Interval to be removed from the selection
    */
-  remove (item, isStreamSelection = false) {
+  remove (item) {
     for (let i = 0; i < this._items.length; i++) {
       const existing = this._items[i]
 
       if (existing.isStreamSelection) {
-        if (isStreamSelection) {
+        if (item.isStreamSelection) {
           // If both are stream selections and they match, then we remove the first matching item, then we break the loop
           if (existing.from === item.from && existing.to === item.to) {
             this._items.splice(i, 1)
@@ -66,16 +65,15 @@ export class Selections {
 
   /**
    * @param {SelectionItem} newItem
-   * @param {boolean} [isStreamSelection]
    */
-  insert (newItem, isStreamSelection = false) {
+  insert (newItem) {
     if (newItem.from > newItem.to) {
       throw new Error('Invalid interval')
     }
-    if (!isStreamSelection) {
+    if (!newItem.isStreamSelection) {
       this.remove(newItem)
     }
-    this._items.push({ ...newItem, isStreamSelection })
+    this._items.push({ ...newItem })
   }
 
   /**
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -80,10 +80,6 @@ export class Selections {
     this._items.sort(sortFn)
   }
 
-  isEmpty () {
-    return this._items.length === 0
-  }
-
   get length () {
     return this._items.length
   }
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -19,7 +19,6 @@
 export class Selections {
   /** @type {Array<SelectionItem>} */
   _items = []
-
   /** @type {Array<NotificationItem>} */
   _notifications = []
 
@@ -27,14 +26,12 @@ export class Selections {
    * @param {MinimalSelectionItem & {isStreamSelection?: boolean}} item Interval to be removed from the selection
    */
   remove (item) {
-    for (let i = 0; i < this._items.length; i++) {
-      const existing = this._items[i]
-
+    for (const existing of this) {
       if (existing.isStreamSelection) {
         if (item.isStreamSelection) {
           // If both are stream selections and they match, then we remove the first matching item, then we break the loop
           if (existing.from === item.from && existing.to === item.to) {
-            this._items.splice(i, 1)
+            existing.remove()
             // for stream selections, we only remove one item at a time
             // ergo we break the loop after removing the first matching item
             break
@@ -45,28 +42,9 @@ export class Selections {
         }
       }
 
-      if (isLowerIntersecting(item, existing)) {
-        existing.to = Math.max(item.from - 1, 0)
-        continue
-      }
-      if (isUpperIntersecting(item, existing)) {
-        existing.from = item.to + 1
-        continue
-      }
-      if (isInsideExisting(item, existing)) {
-        const existingStart = { ...existing, to: Math.max(item.from - 1, 0) }
-        const existingEnd = { ...existing, from: item.to + 1 }
-        if (existingStart.to - existingStart.from > 0) {
-          this._items.splice(i, 1, existingStart, existingEnd)
-        } else {
-          this._items.splice(i, 1, existingEnd)
-        }
-        continue
-      }
-      if (isCoveringExisting(item, existing)) {
-        this._items.splice(i, 1)
-        i -= 1
-        continue
+      const clearedExisting = clearIntersection(item, existing)
+      if (clearedExisting !== null) {
+        existing.replaceWith(clearedExisting)
       }
     }
   }
@@ -86,36 +64,20 @@ export class Selections {
     if (notify) this._notifications.push({ from, to, notify })
   }
 
-  /**
-   * Returns all the notifications that in any way intersect with the given range
-   * @param {MinimalSelectionItem} range
-   */
-  getNotificationsByRange (range) {
-    return this._notifications.filter(notification => {
-      if ((notification.to - notification.from) <= 0) return false
-      return isCoveringExisting(range, notification) ||
-        isInsideExisting(range, notification) ||
-        isLowerIntersecting(range, notification) ||
-        isUpperIntersecting(range, notification)
-    })
-  }
-
-  /**
-   * Removes a notification from the list of notifications by reference
-   * @param {NotificationItem} notification
-   */
-  removeNotification (notification) {
-    const index = this._notifications.indexOf(notification)
-    if (index !== -1) {
-      this._notifications.splice(index, 1)
-    } else {
-      throw new Error('Notification not found')
+  /** @param {MinimalSelectionItem} range */
+  callNotifyInRange (range) {
+    for (const notification of this._notifications) {
+      if (isLowerIntersecting(range, notification) ||
+            isUpperIntersecting(range, notification) ||
+            isInsideExisting(range, notification) ||
+            isCoveringExisting(range, notification)
+      ) {
+        notification.notify()
+      }
     }
   }
 
-  /**
-   * @param {(a: SelectionItem, b: SelectionItem) => number} sortFn
-   */
+  /** @param {(a: SelectionItem, b: SelectionItem) => number} sortFn */
   sort (sortFn = (a, b) => a.from - b.from) {
     this._items.sort(sortFn)
   }
@@ -124,28 +86,11 @@ export class Selections {
     return this._items.length
   }
 
-  /**
-   * Takes an index and returns the item at that index
-   * @param {number} index
-   */
+  /**  @param {number} index */
   get (index) {
     return this._items[index]
   }
 
-  /**
-   * Takes an index and an item and replaces the item at the index with the new item
-   * @param {number} index
-   * @param {SelectionItem} item
-   */
-  set (index, item) {
-    this._items[index] = item
-  }
-
-  /**
-   * Takes two indexes and swaps the items at those indexes
-   * @param {number} i
-   * @param {number} j
-   */
   swap (i, j) {
     const temp = this._items[i]
     this._items[i] = this._items[j]
@@ -156,16 +101,33 @@ export class Selections {
     this._items.length = 0
   }
 
+  /** @returns {Generator<SelectionItem & {remove: () => void, replaceWith: (MinimalSelectionItem[]) => void}>} */
   * [Symbol.iterator] () {
     for (let i = 0; i < this._items.length; i++) {
       const item = this._items[i]
-      yield {
-        ...item,
-        remove: () => {
-          this._items.splice(i, 1)
-          i -= 1
+
+      item.replaceWith = (replacingItems) => {
+        this._items.splice(i, 1, ...replacingItems)
+        i = i - 1 + replacingItems.length // adjust i after splice
+
+        if (replacingItems.length !== 0) return
+        // when a selection is finished and removed, we need to also remove this interval from the notifications
+        for (let j = 0; j < this._notifications.length; j++) {
+          const notification = this._notifications[j]
+          const clearedNotification = clearIntersection(item, notification)
+          if (clearedNotification !== null) {
+            this._notifications.splice(j, 1, ...clearedNotification)
+            j = j - 1 + clearedNotification.length // decrement j to offset splice
+          }
         }
       }
+
+      item.remove = () => item.replaceWith([])
+
+      yield item
+
+      delete item.replaceWith
+      delete item.remove
     }
   }
 }
@@ -218,4 +180,34 @@ export function isInsideExisting (newItem, existing) {
  */
 export function isCoveringExisting (newItem, existing) {
   return newItem.from <= existing.from && newItem.to >= existing.to
+}
+
+/**
+ * Takes two intervals and returns null if no changes are needed, or an array of intervals that remain
+ * after the existing interval is truncated or removed to make space for the new interval
+ * @param {MinimalSelectionItem} item
+ * @param {MinimalSelectionItem} existing
+ * @return {MinimalSelectionItem[] | null}
+ */
+export const clearIntersection = (item, existing) => {
+  if (isLowerIntersecting(item, existing)) {
+    existing.to = Math.max(item.from - 1, 0)
+    return [existing]
+  }
+  if (isUpperIntersecting(item, existing)) {
+    existing.from = item.to + 1
+    return [existing]
+  }
+  if (isInsideExisting(item, existing)) {
+    const returnedArray = []
+    const existingStart = { ...existing, to: Math.max(item.from - 1, 0) }
+    if (existingStart.to - existingStart.from > 0) returnedArray.push(existingStart)
+    const existingEnd = { ...existing, from: item.to + 1 }
+    if (existingEnd.to - existingEnd.from > 0) returnedArray.push(existingEnd)
+    return returnedArray
+  }
+  if (isCoveringExisting(item, existing)) {
+    return []
+  }
+  return null
 }
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -8,7 +8,7 @@
  * @typedef {MinimalSelectionItem & {
  *  offset: number,
  *  priority: number,
- *  notify: function
+ *  notify?: function
  *  isStreamSelection?: boolean
  * }} SelectionItem
  */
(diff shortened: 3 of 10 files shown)


ThaUnknown:
code quality is good, I haven't look at the actual functionality tho, it would be good to actually explain what you did, and how the algorithm changed, you also cannot get rid of priority.

ThaUnknown:
also if you're already addessing select/deselect behavior it would be good if you included what <https://github.com/webtorrent/webtorrent/pull/2115> tries to accomplish, not exactly its code, but it's functionality

detarkende:
## Explanation of my changes:

To make the _selections array non-overlapping, I abstraced it away into a separate class.
This class internally has the same array, but its `insert` and `remove` methods include some extra logic to keep items separate:

When we remove an interval, we loop over the existing items and check if there's any overlap with the removed interval. There can be 4 types of overlapping:
- `isLowerIntersecting`: Only the lower end of the removed interval overlaps 
    - Example: we are removing `8-12`, but `1-10` is already selected, so we cut the end of the existing selection off.

- `isUpperIntersecting`: Only the upper end of the removed interval overlaps 
    - Example: we are removing `10-22`, but `20-25` is already selected, so we cut the beginning of the existing selection off.
 
- `isInsideExisting`: The removed interval is specifically inside an existing larger selection. 
    - Example: we are removing `10-15`, but `1-30` is already selected, so we carve out the removed section from the middle of the existing one -> this leaves us with 2 selections with a hole between them.

- `isCoveringExisting`: This is the simplest case. If the removed interval covers an entire existing selection, or if they're an exact match size-wise, then we simply remove the existing selection.

Insertion is very simple compared to this: we call the `remove` method with the newly inserted interval to make sure that there's no overlapping items, then we push the selection to the array.

---

### On the topic of priorities:

I understand why we assign priorities to pieces when they are selected, but I don't see why we would need them upon deselection. 

From recently reading around I got the feeling that it's for making sure that we don't deselect intervals that were selected by file streams -- therefore in the past we would only deselect the given pieces if the start and end points matched exactly and also the priority. 

Is this right? If so, then wouldn't it make more sense to distinguish stream-selections from non-stream selections with a boolean flag instead of their priority? If this is the case, then I'll adjust my code to make sure that only files can deselect file streams and not end users accidentally. Let me know if I'm wrong on this.

ThaUnknown:
> I understand why we assign priorities to pieces when they are selected, but I don't see why we would need them upon deselection.

your assumption is roughly correct, but it doesn't apply to only streams, imagine webtorrent's HTTP server being used as a media server, and having say 10 people watching 1 video as it's being downloaded/streamed from BitTorrent

if multiple people are at the same point in time in the video, and one of them stops watching it unloads the video, which sends a cancel request to the range that was requested, thus deselecting the pieces for ALL 9 other people that are watching it, permanently stalling the playback for 9 people until someone reloads it

this ofc applies to not only streams, but manually running selection ranges, your code works for single selections, but when you have multiple unrelated processes making selections, it falls apart

hope that clears it up

detarkende:
I think I finally understand. So the file stores one iterator for each stream created. If there are multiple requests as the same time, there might be 2 iterators for the same (or overlapping) piece intervals. When the file iterator's `destroy` method is called, it deselects the selected pieces. Which means that if selections cannot overlap, then one iterator might clear its own selection, but with that also clear the selection for all other iterators.

I think I might be able to solve this by keeping a separate array of selections specifically for these cases. This would ensure that existing features don't break. I'll try to write the code now and push it as soon as I can. I'll try to setup an example project as well, to test that my solution actually works.

ThaUnknown:
yeup, u got it right.

detarkende:
@ThaUnknown Thanks for the guidance so far. I have implemented 2 things:
1. Streams create specific stream-selections, which are marked with a true `isStreamSelection` flag. These can only be deselected by streams, and only one at a time. The actual implementation of this is mostly the same as in #2115, since that has already received a round of code review - so I assume it's not a terrible way to do it 😄 
2. In #2115, @feross suggested a new option in the torrent options: the `deselect` option. I added that as well.

I copied the tests from Alex's PR and modified them somewhat + just to be on the safe side, I created a test server and connected to it with 2 clients, then disconnected one of them. The other client's selection remained in the `_selections` and its video playback continued without any issues.

I read somewhere that you should never modify the git history, so I basically sent everything in a separate commit, let me know if I should squash my commits though.

lib/torrent.js
```diff
@@ -35,6 +35,7 @@ import utp from './utp.cjs' // browser exclude
 import WebConn from './webconn.js'
 
 import info from '../package.json' assert { type: 'json' }
+import { Selection } from './selection.js'
```
ThaUnknown, line 38: this should be above the info import, right under webconn

lib/torrent.js
```diff
@@ -688,7 +690,9 @@ export default class Torrent extends EventEmitter {
       : this.pieceLength
     this.pieces[index] = new Piece(len)
     this.bitfield.set(index, false)
-    this.select(index, index, 1)
+    if (!this._startAsDeselected) {
```
ThaUnknown, line 693: inline this without {}

lib/torrent.js
```diff
@@ -1019,7 +1025,7 @@ export default class Torrent extends EventEmitter {
     this._drain()
   }
 
-  select (start, end, priority, notify) {
+  _select (start, end, priority, notify, isStreamSelection) {
```
ThaUnknown, line 1028: notify = noop, isStreamSelection = false) {

lib/torrent.js
```diff
@@ -1029,36 +1035,37 @@ export default class Torrent extends EventEmitter {
...
       from: start,
       to: end,
       offset: 0,
       priority,
       notify: notify || noop
-    })
+    }, isStreamSelection)
 
-    this._selections.sort((a, b) => b.priority - a.priority)
+    this._selections.sort((a, b) => a.priority - b.priority)
```
ThaUnknown, line 1046: why was the order reversed?

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+ * @typedef {Object} SelectionItem
+ * @property {number} from
+ * @property {number} to
+ * @property {number} offset
+ * @property {number} priority
+ * @property {function} notify
+ * @property {boolean} isStreamSelection
+ */
+
+export class Selection {
```
ThaUnknown, line 11: this should be Selections plural, since it handles many selections

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * @type {Array<SelectionItem>}
+   * @private
```
ThaUnknown, line 14: skip these @private and @public jsdocs, we'd use # if we wanted to declare private, but we don't because we want people to access internal functions if they wish to, which is why _ is used for private, for strict engines such as TS, touching this will throw errors, which is annoying

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+   * @param {(a: SelectionItem, b: SelectionItem) => number} sortFn
+   */
+  sort (sortFn = (a, b) => a.from - b.from) {
+    this._items.sort(sortFn)
+  }
+
+  /**
+   * @public
+   */
+  isEmpty () {
```
ThaUnknown, line 90: this is VERY redundant

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * @public
+   */
+  clear () {
+    this._items.length = 0
+  }
+
+  * [Symbol.iterator] () {
+    const self = this
```
ThaUnknown, line 140: no.... don't use self
do remove: () => {}

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+    for (let i = 0; i < this._items.length; i++) {
+      const item = this._items[i]
+      yield {
```
ThaUnknown, line 143: this.... works? but feels a bit too hacky? if u figure out a cleaner way to do this then great, if not oh well it's fine

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+ * Examples:
+ * existing: 10-20 | 10-20 | 10-20
+ * new_item: 10-21 | 09-20 | 10-20
+ * @param {SelectionItem} newItem
+ * @param {SelectionItem} existing
+ * @returns {boolean} returns true if the new item is covering the existing item
+ */
+export function isCoveringExisting (newItem, existing) {
+  return newItem.from <= existing.from && newItem.to >= existing.to
+}
```
ThaUnknown, line 202: you don't need return types on any of these as tslib will detect that this always returns bool, the returns comments you could move into the bottom of the top comment

lib/torrent.js
```diff
@@ -1029,36 +1035,37 @@ export default class Torrent extends EventEmitter {
 
     this._debug('select %s-%s (priority %s)', start, end, priority)
 
-    this._selections.push({
+    this._selections.insert({
```
ThaUnknown, line 1038: SelectionItem implements `isStreamSelection` so simply pass it in the object, rather than creating a new function parameter?

lib/selection.js
ThaUnknown: these params should also be `{from: number, to: number}` or `MinimalSelectionItem` if you want to make something like that, not SelectionItem

ThaUnknown (changes requested):
This PR is high quality, so I'm going to be VERY strict on it :)

hope that doesn't demotivate you

detarkende:
I hope so too. Sure, go for it 😄

ThaUnknown:
Webtorrent's selections also have 1 more issue, I'm not gonna ask you to fix it, but it would be insanely welcome as it's the last thing that sucks in this implementation: webtorrent selects full pieces, rather than specific bytes.

This becomes an issue for example when there's a torrent with 2 files, and file 1 ends at 1/2 piece and file 2 starts at where that 1/2 piece ended, selecting and deselecting the files, impacts the other file's selections.
this is especially an issue when creating file streams/iterators, as in theory removing an iterator from the start of the 2nd file, can remove the selection for the end of the 1st file, which ignores metadata download, and effectively stalls all playback etc

The way I could see this solved, is that internally we'd select byte ranges, but users would still request pieces.
This means that file selections aka torrent.file.select would select ranges, and deselecting it would also deselect ranges, then the indexes specific selections target would be exposed as parts of those ranges.

This is hard to explain but TLDR: internally selections should be byte ranges, however they should be exposed to users and the selection algorithms as piece numbers.

If you're interested in this, and didn't understand my shit explanations, ask, and I'll put more effort into properly explaining this.

lib/selection.js
```diff
@@ -0,0 +1,202 @@
...
+  _items = []
+
+  /**
+   * @public
+   * @param {{from: number, to: number}} item Interval to be removed from the selection
+   * @param {boolean} [isStreamSelection]
+   */
+  remove (item, isStreamSelection = false) {
+    for (let i = 0; i < this._items.length; i++) {
+      const existing = this._items[i]
```
ThaUnknown, line 25: you do not check priorities here, remove should remove by priorities!!!!!!!!!!!

lib/selection.js
detarkende: I don't consider this to be too hacky to be honest. The other solution would be to call `.splice` on `_selections._items` from outside of this file, which does feel hacky tbh.
Plus this allows us to use `for (.. of ..)` loops and still remove items without having to deal with the indexes. If you think of a better solution, I'd be happy to replace the current one, but I also don't really see this as an issue either, since it's easily understandable and simple.

lib/selection.js
detarkende: The return types are not there necessarily there to provide the types, but rather to leave a comment on **when** this type returns true. As you can see they all have a _"returns true if the new items..."_ explanation after the `@returns {boolean}`. To me it makes sense to put these comments to the `@returns` section, but let me know if you'd still like me to change it.

detarkende: Also, I like the idea of having a smaller `MinimalSelectionItem` typedef, so thanks. I added that too.

lib/torrent.js
detarkende: got it, thanks. fixed it

lib/torrent.js
detarkende: done.

lib/torrent.js
detarkende: fixed.

lib/torrent.js
detarkende: honestly no clue. probably just an accident, but good catch. I reverted to the old order

lib/selection.js
detarkende: You're right, thanks for this. I removed them all

lib/selection.js
detarkende: can't really argue with that haha 😄 removed it.

lib/selection.js
detarkende: thanks, good idea. fixed.

lib/torrent.js
detarkende: That's a good point. I removed the extra argument from `insert` and `select`, `deselect`. It's now determined by the shape of the object.

detarkende:
> Webtorrent's selections also have 1 more issue, I'm not gonna ask you to fix it, but it would be insanely welcome as it's the last thing that sucks in this implementation: webtorrent selects full pieces, rather than specific bytes.

Personally, I like the idea of having byte-range selections. I haven't given this too much thought yet, but I imagine it could be used to specifically not download any files, other than the selected one(s). For me this was an issue previously because when a torrent has 2 large files and there's a piece that contains data from both files, it creates the second file too (even when I haven't selected it). Now this shouldn't be an issue, since I didn't actually download the second file's contents, but to the file system it is indicated that the second file is also a full file (let's say 10GB). Now I know it's not actually taking up that space, but it's still annoying.

That said, I think that would be a pretty big undertaking and considering that this PR has also grown to be around ~450 LOC in changes, I would prefer to keep those changes in a separate PR.

I can't guarantee to work on it after this PR, because I would first like to address some other issues - namely fixing the very out of date `@types/webtorrent` declaration files, or maybe adding more jsdoc annotations to the code, but I might give it a go later.

TLDR: I like the idea and support it, but I think it should be in a separate PR. I won't start working on it right away, but if I have time later, I might take a stab at it.

lib/selection.js
detarkende: I added the priority check to the section that compares `isStreamSelection` selections

For regular `select` and `deselect`, priority shouldn't matter:
- If it's an insertion, then it should overwrite whatever was there previously.
- If we are using `deselect`, then the user shouldn't have to know beforehand what the current selection's priority is. Especially when you deselect a range that contains multiple existing ranges - they could all be different selections.

detarkende:
I'm going to check out the failing tests, but I adressed your comments. 

Thank you for the detailed review! I don't think there are too many details left to debate, so we might ship this thing quite quickly 🚀 _Who said open-source is slow?_ 😄

lib/selections.js
```diff
@@ -152,35 +134,26 @@ export class Selection {
 }
 
 /**
- * Examples:
- * existing: 1-10 | 1-10
- * new_item: 8-12 | 10-15
- * @param {SelectionItem} newItem
- * @param {SelectionItem} existing
+ * @param {MinimalSelectionItem} newItem
```
ThaUnknown, line 137: aw man those explanations were good :(

ThaUnknown:
> but I think it should be in a separate PR

very fair

> Who said open-source is slow? 

it is, I just sacrifice a lot of free time for FOSS, as you see the previous PR is still open

lib/selections.js
detarkende: you mean the examples? I can add them back, I just thought it made things kind of messy, but if you found them helpful, then they're probably worth keeping

ThaUnknown: for you maybe, but for some1 who hasnt worked on this in the future it's useful

detarkende: okay, sure. I added them back in then :)

lib/torrent.js
```diff
@@ -1029,36 +1033,38 @@ export default class Torrent extends EventEmitter {
...
+    this._debug('deselect %s-%s', start, end)
 
-    for (let i = 0; i < this._selections.length; ++i) {
-      const s = this._selections[i]
-      if (s.from === start && s.to === end && s.priority === priority) {
-        this._selections.splice(i, 1)
-        break
-      }
-    }
+    this._selections.remove({ from: start, to: end, isStreamSelection })
```
ThaUnknown, line 1059: :/ this throws typeerror during static analysis
```
Argument of type '{ from: any; to: any; isStreamSelection: boolean; }' is not assignable to parameter of type 'SelectionItem'.
  Type '{ from: any; to: any; isStreamSelection: boolean; }' is missing the following properties from type '{ offset: number; priority: number; notify: Function; isStreamSelection?: boolean | undefined; }': offset, priority, notifyts(2345)
```

ThaUnknown (changes requested):
yup, this is gonna be a breaking change, but I think it's needed and I don't know if its the best idea, @SilentBot1 what do u think

test/client-deselect.js
```diff
@@ -0,0 +1,109 @@
...
+
+  client1.seed(fixtures.leaves.content, {
+    name: 'Leaves of Grass by Walt Whitman.epub',
+    announce: [],
+    store: MemoryChunkStore
+  }, () => {
+    client2.add(parsedTorrent, { store: MemoryChunkStore, ...addTorrentProps }, (torrent) => {
+      onTorrent(torrent)
+
+      torrent.addPeer(`localhost:${client1.torrentPort}`)
```
ThaUnknown, line 25: you're adding a peer by IP, this is a no-go in browser, move this test to test/node

lib/torrent.js
detarkende: Thanks, I updated it to 
```js
/**
* @param {MinimalSelectionItem & {isStreamSelection: boolean, priority?: number}} item 
*/

File: lib/file.js [hunk 16]
```diff
@@ -91,7 +91,7 @@ export default class File extends EventEmitter {
 
   deselect () {
```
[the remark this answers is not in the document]
detarkende: I replied below: https://github.com/webtorrent/webtorrent/pull/2757#discussion_r1511719326

lib/torrent.js
```diff
@@ -1029,36 +1033,38 @@ export default class Torrent extends EventEmitter {
...
 
     this._updateSelections()
   }
 
+  deselect (start, end) {
```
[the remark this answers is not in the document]
ThaUnknown: sorry, there's some overlap between what I want and find missing in WebTorrent, and what you're doing, so I partially wanted you to do it. Selections should be unique by priority levels, imagine a torrent client UI, where u can download different files, at different priorities, but if you want to remove that elevated priority from a file, you'd deselect the file.... instead of removing the selection at that priority

will adding a new selection at a lower priority for only part of an older higher selection override only the overlapping part?

I hadn't had much time to delve into the code, so I was making rough assumptions by quick glances

if that's the case then the if  `if (existing.from === item.from && existing.to === item.to && existing.priority === item.priority)` where I asked u to add priorities is stupid and it was simply my lack of understanding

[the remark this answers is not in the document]
ThaUnknown: image
@detarkende my wonderful paint skills

lib/selections.js
```diff
@@ -0,0 +1,184 @@
...
+   * @param {MinimalSelectionItem & {isStreamSelection: boolean, priority?: number}} item Interval to be removed from the selection
+   */
+  remove (item) {
+    for (let i = 0; i < this._items.length; i++) {
+      const existing = this._items[i]
+
+      if (existing.isStreamSelection) {
+        if (item.isStreamSelection) {
+          // If both are stream selections and they match, then we remove the first matching item, then we break the loop
+          if (existing.from === item.from && existing.to === item.to && existing.priority === item.priority) {
```
ThaUnknown, line 32: i was stupid, this priority check isnt needed

detarkende: Are you sure? This check is for stream selections only. Those are still on the old logic: from, to and prio have to match and if you have multiple selections for the same range, then you have to call deselect for each.

lib/selections.js
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * @type {Array<SelectionItem>}
+   */
+  _items = []
+
+  /**
+   * @param {MinimalSelectionItem & {isStreamSelection: boolean, priority?: number}} item Interval to be removed from the selection
+   */
+  remove (item) {
+    for (let i = 0; i < this._items.length; i++) {
```
ThaUnknown, line 26: should this be a `for (const existing of this) {` then later use either existing.remove() or this.swap()?

ThaUnknown: nvm it wouldn't be swap, but remove could be used

lib/selections.js
ThaUnknown: but stream selections cannot be created with priorities, unless some1 does it manually ,which isn't it's intended use case? or am I overlooking smth

ThaUnknown: sorry i mean the if statement stays, but the priority check is useless

detarkende: Oh, okay. I didn't think it through and thought there might be stream selections with differing priorities, but that's not the case. In that case, I removed the priority comparison from the if statement.

lib/selections.js
detarkende: we also need `i` for the `splice` call on line 55.

test/node/client-deselect.js
```diff
@@ -0,0 +1,109 @@
...
+    onDone: (torrent) => {
+      t.equal(torrent.pieces.filter((a) => a === null).length, torrent.pieces.length)
+    }
+  })
+})
+
+test('client.select: partial torrent', function (t) {
+  t.plan(3)
+
+  let lastPiece
```
ThaUnknown, line 54: this is touch missleading, torrent's have a concept of "lastPiece" which is "the last piece in the torrent which doesn't have the length of a normal piece, instead is the remaining data" so this could be treated as lastPieceLength, better be safe and rename this to `lastPieceIndex` everywhere in this test file

lib/selections.js
ThaUnknown: ah no, streams simply select needed data with the default unset priority, and data that's needed for the stream load RIGHT NOW is set to critical, which is highest priority

test/node/client-deselect.js
detarkende: didn't know about that, plus this test is not originally mine. thanks for the info. renamed them all

lib/torrent.js
```diff
@@ -549,7 +551,7 @@ export default class Torrent extends EventEmitter {
       })
     } else {
       // start off selecting the entire torrent with low priority
-      if (this.pieces.length !== 0) {
+      if (this.pieces.length !== 0 && !this._startAsDeselected) {
         this.select(0, this.pieces.length - 1, false)
```
ThaUnknown, line 555: what about this, why false?

ThaUnknown (commented):
One issue stands with these changes in total, `notify` can be swallowed or overwritten, that's kinda bad.... when you merge or overwrite selections, you should also merge their notify functions? imagine multiple functions ask for notify on different priority levels, all but one will be lost....

lib/torrent.js
detarkende: I didn't change this line, but I guess in this case I could replace the `false` to `0`. They're basically the same thing here.
Replaced in last commit

ThaUnknown:
yeah @detarkende you'll need to "merge" callback/notify functions when selections are merged/overwritten, this will call the notify function when it's not really necessary, but its better to call it more than less, I think it would be roughly like this:
```js
const existingStart = { ...existing, to: item.from - 1, notify: ()=>{item.notify();existing.notify()} }
```
probably? you'd need to double check this, and ofc it would be in other places too

SilentBot1:
The code looks overall high quality, thanks for taking the time to PR this @detarkende.

Through my testing, I only found one real issue, and one thing that may cause issues:

1. The first being that the `&so=` magnet query parameter is preventing any selections from being made:
```js
client.add('magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent&so=0', {deselect: false})
// No selection made - incorrect

client.add('magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent&so=1')
// No selection made - incorrect

client.add('magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent&so=5', {deselect: true})
// No selection made - I would say this is correct, explicit deselect should take priority over &so=
```

2. The second being that two overlapping sections at the start of the piece range leaves a negative selection:
```js
client.torrents[0].select(0, 100)
client.torrents[0].select(0, 50)
```
which leads to an unfulfillable selection:
```js
Selections {
  _items: [
    {
      from: 0,
      to: -1,
      offset: 0,
      priority: 0,
      notify: [Function: noop],
      isStreamSelection: false
    },
    ...
  ]
}
```

Other than that, it looks like a nice addition!

detarkende:
> yeah @detarkende you'll need to "merge" callback/notify functions when selections are merged/overwritten, this will call the notify function when it's not really necessary, but its better to call it more than less, I think it would be roughly like this:
> 
> ```js
> const existingStart = { ...existing, to: item.from - 1, notify: ()=>{item.notify();existing.notify()} }
> ```
> 
> probably? you'd need to double check this, and ofc it would be in other places too

@ThaUnknown Thanks for the idea. I'm currently trying to make it work with an independent notifications array inside Selections. It's kinda the same idea what #2115 does. I have some issues with it atm, but I'll try to fix them. If I can't I'll probably use your idea.

detarkende:
@ThaUnknown pushed some changes that address the notifications issue. I believe it solves the problem, but maybe I didn't think of some edge case. lmk what you think.

detarkende:
@SilentBot1 Thank you for the review. You've got a good eye for bugs, I would've never caught those two by myself.
I addressed both of them in my last 2 commits. Let me know what you think

lib/selections.js
```diff
@@ -0,0 +1,222 @@
...
+      return isNotEmpty && isIntersecting
+    })
+  }
+
+  /**
+   * Removes a notification from the list of notifications by reference
+   * @param {NotificationItem} notification
+   */
+  removeNotification (notification) {
+    const index = this._notifications.findIndex(n => n === notification)
```
ThaUnknown, line 111: .indexOf()

lib/selections.js
```diff
@@ -0,0 +1,222 @@
...
+  insert (newItem) {
+    const { from, to, notify, isStreamSelection, ...restItem } = newItem
+    if (from > to) {
+      throw new Error('Invalid interval')
+    }
+    if (!isStreamSelection) {
+      this.remove({ from, to, isStreamSelection })
+    }
+    this._items.push({ from, to, isStreamSelection, ...restItem })
+    this._notifications.push({ from, to, notify })
```
ThaUnknown, line 86: I don't like this, you're essentially using the old range behavior, and it has it's own separate state, not like the rest, it's way too complex

lib/torrent.js
```diff
@@ -1302,35 +1306,38 @@ export default class Torrent extends EventEmitter {
...
       const oldOffset = s.offset
 
       // check for newly downloaded pieces in selection
       while (this.bitfield.get(s.from + s.offset) && s.from + s.offset < s.to) {
         s.offset += 1
       }
 
-      if (oldOffset !== s.offset) s.notify()
+      if (oldOffset !== s.offset) {
+        this._selections.getNotificationsByRange(s.from, s.from + s.offset).forEach(({ notify }) => {
```
ThaUnknown, line 1318: this... is gonna be insanely slow when there are a lot of selections?, also use for of, not forEach

ThaUnknown:
I'm not a fan of this notification code, because it's complexity skyrocketed compared to the other things, and I can't review it by simply looking at it, and I don't have the time to test it

lib/selections.js
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * Returns all the notifications that in any way intersect with the given range
+   * @param {number} from
+   * @param {number} to
+   */
+  getNotificationsByRange (from, to) {
+    const range = { from, to }
+    return this._notifications.filter(notification => {
+      const isNotEmpty = (notification.to - notification.from) > 0
```
ThaUnknown, line 97:
```suggestion
      // isNotEmpty
      if ((notification.to - notification.from) > 0) return false
```

ThaUnknown:
@detarkende as expected, the notification code is too complex and already broken somehow, in a way I can't find, I fixed it not emitting notify for first pieces, made it not run when no value is specified by removing noops, but silent found another issue where removing a manually created selection didn't remove it's notification? there's fuckery afoot

ThaUnknown:
also, you made me put actual effort into this, I hope you're proud of yourself..... should tell you that this is an actually important change that's been a long standing issue :p

ThaUnknown:
couldn't we get rid of offset like in this PR? https://github.com/webtorrent/webtorrent/pull/2115/files#diff-43f8f193b859b154ddc4cb89276649d35e2eb59859b68eae0c2ad58f7306aa59R1211

ThaUnknown:
silent will test this PR, look into that offset thing, and then we'll merge this if there are no issues

detarkende:
> @detarkende as expected, the notification code is too complex and already broken somehow

You seem unsatisfied with the current solution. Do you think we should handle notifications alongside selection ranges and merge them, when we merge the selections? 

I'm definitely not against that, but I also saw that you committed changes to the current version, so now I'm hesitant to change it to your initial idea.

Do you want to fix the current version, or scrap it and move to a different solution?

detarkende:
> couldn't we get rid of offset like in this PR? https://github.com/webtorrent/webtorrent/pull/2115/files#diff-43f8f193b859b154ddc4cb89276649d35e2eb59859b68eae0c2ad58f7306aa59R1211

I tried getting rid of the offset, but ran into a bunch of small bugs that were difficult to debug. Not sure what causes them, I was working on something else at the time.

I can look into it, but I personally kinda like the current solution, where we keep the original selection's size and the progress separate.

lib/torrent.js
```diff
@@ -1302,23 +1306,28 @@ export default class Torrent extends EventEmitter {
...
       const oldOffset = s.offset
 
       // check for newly downloaded pieces in selection
       while (this.bitfield.get(s.from + s.offset) && s.from + s.offset < s.to) {
         s.offset += 1
       }
 
-      if (oldOffset !== s.offset) s.notify()
+      if (oldOffset !== s.offset) {
+        for (const { notify } of this._selections.getNotificationsByRange({ from: s.from, to: s.from + s.offset })) {
```
ThaUnknown, line 1318: question: shouldnt this be notificationsIntersecting?

lib/torrent.js
```diff
@@ -1302,23 +1306,28 @@ export default class Torrent extends EventEmitter {
...
       if (s.to !== s.from + s.offset) continue
       if (!this.bitfield.get(s.from + s.offset)) continue
 
-      this._selections.splice(i, 1) // remove fully downloaded selection
-      i -= 1 // decrement i to offset splice
-
-      s.notify()
+      s.remove() // remove fully downloaded selection
+      // remove notifications when we remove selection.
+      for (const notification of this._selections.getNotificationsByRange(s)) {
```
ThaUnknown, line 1327: and shouldnt this be notificationsInside?

the issue I'm seeing is that notifications which are touching, but arent containing a range, will get removed too, meaning they won't emit for other bigger ranges?

ThaUnknown:
> You seem unsatisfied with the current solution. Do you think we should handle notifications alongside selection ranges and merge them, when we merge the selections?

yeah quite so, as you can see there are already issues and uncertainties....

lib/selections.js
```diff
@@ -0,0 +1,221 @@
...
+   */
+  insert (newItem) {
+    const { from, to, notify, isStreamSelection, ...restItem } = newItem
+    if (from > to) {
+      throw new Error('Invalid interval')
+    }
+    if (!isStreamSelection) {
+      this.remove({ from, to, isStreamSelection })
+    }
+    this._items.push({ from, to, isStreamSelection, ...restItem })
```
ThaUnknown, line 85: same here, if its a stream selection, push it to _streamItems not _items

detarkende:
I don't have enough time today or tomorrow, but I'll come back to it on Friday and hopefully fix everything that's left. Thank you for your patience.

ThaUnknown:
silent found an issue in this, but hasnt reported it, idk why
```js
> client.torrents[0].select(0, 1514, 1, ()=>{})
undefined
> client.torrents[0]._selections
Selections {
  _items: [
    {
      from: 0,
      to: 1514,
      isStreamSelection: false,
      offset: 0,
      priority: 1
    }
  ],
  _notifications: [ { from: 0, to: 1514, notify: [Function (anonymous)] } ]
}
> client.torrents[0].deselect(0, 1514)
undefined
> client.torrents[0]._selections
Selections {
  _items: [],
  _notifications: [ { from: 0, to: 1514, notify: [Function (anonymous)] } ]
}
```

ThaUnknown:
yeaaaaaaah, this is what I meant with insane complexity

from my POV: 80% of the PR and 90% of the complexity is now the selection notifications, which are rarely gonna be used in the first place
it's complexity is now high enough where I don't think I'll be able to understand and review all the potential cases

I really think we should have some simpler system like merging, and introduce higher accuracy for noti in a separate PR

detarkende:
@ThaUnknown, look, I agree with you. I don't like that notifications is this complex. I wouldn't say that it's 80% of this PRs complexity, but it's too much for such a small feature.

I'll be honest, I searched through dozens of github repos that use webtorrent and none of them use this feature. I also asked around on reddit - no answer so far. If it was up to me, I'd remove it in a second. We're changing the select/deselect behaviour anyway, why not get rid of this unused feature in exchange for sensible selec/deselect? But I understand that we should avoid breaking changes, even for such small things.

I'm willing to properly test the current implementation, which should be as close to the original implementation as we can get, though it is complex.

---

<details>
<summary>
The other path we can consider (besides removing and the current implementation) is that we could change the behaviour slightly (dropdown)
</summary>

Before the current solution we simply stored 1 notification callback per selection and just called that. That was simple and understandable, but I think the issue would be the following:

```js
torrent.select(0, 20, 0, () => console.log('downloading [0 - 20]'))
torrent.select(10, 40, 0, () => console.log('downloading [10 - 40]'))
```
The second selection truncates the first one, so it's kinda false that it's "downloading [0 - 20]", since it's actually "downloading [0 - 9]".

Logging is pretty much the only legit use case for the notify function, so at least that should be accurate.

So what if we use this simple solution, but we pass the `from` and `to` values to `notify`?

```js
torrent.select(0, 20, 0, (from, to) => console.log(`downloading [${from} - ${to}]`))
```

This would always report it accurately and it would keep the implementation "stupid simple". I think it's worth considering.

---

My issue with merging selections is that it would be even less accurate than not doing anything. If we would merge notifications, it would call the `[0-20]` notification even when the downloaded pieces are actually between 21-40. I think in that case the "no-change" solution might be more correct, since it would never "lie", it would just not call between 10-20.

</details>

ThaUnknown:
I agree that I'd also like to drop this feature, it was so unusable that I created a `verified` even because it's simply easier to work with, and a lot faster, but unfortunately we can't do that, because I was told that breaking functionality should be a major release, and a major release needs to be agreed upon by everyone, this doesn't warrant that.

That said we can simply neuter it, as I mentioned merging callbacks, instead of adding this `from` `to` stuff would be a lot easier, be a lot safer, and sure in a few cases it would be called when not necessary, but it's still called rarely enough where that doesn't matter, and since the callback doesn't give any data, the developer needs to check the bitfield if the piece he's looking for is verified, so the developer's end functionality wouldn't change

detarkende:
@ThaUnknown I took your advice and went with the merging solution. I am however thinking about splitting the inserted selection at the intersections and only merging the notify function of the section that's actually overlapping. It's just a thought for now, but that would mean that the notify function merging would be much more precise, though at the expense of creating more objects in the `_items` array. 

Let me know what you think. I can provide some sketches if that helps.

lib/selections.js
```diff
@@ -53,28 +65,26 @@ export class Selections {
...
-  /** @param {MinimalSelectionItem} range */
-  callNotifyInRange (range) {
-    for (const notification of this._notifications) {
-      if (isLowerIntersecting(range, notification) ||
-            isUpperIntersecting(range, notification) ||
-            isInsideExisting(range, notification) ||
-            isCoveringExisting(range, notification)
-      ) {
-        notification.notify()
+    newItem.notify = newItem.notify ?? noop
```
ThaUnknown, line 71: don't noop the notifiy, if it doesn't exist, skip it, less function calls

lib/selections.js
```diff
@@ -53,28 +65,26 @@ export class Selections {
...
+    if (!newItem.isStreamSelection) {
+      const { notify: oldNotify } = newItem
+      // Merge notifications of intersecting items into the new item's notify function
+      const intersectingNotifyFunctions = new Set(
```
ThaUnknown, line 75: is there any point in de-duplicating them? it's unlikely it will ever happen and it forces you to iterate over the items 3 times, instead of once

lib/selections.js
detarkende: Got it. removed the noop, instead the notify functions are called like `s.notify?.()`.

lib/selections.js
detarkende: It was only there to merge all the noops, since they would all point to the same reference. But since noops are removed, I replaced this with a single `for..of` loop.

lib/selections.js
```diff
@@ -0,0 +1,185 @@
...
+  /**
+   * @param {SelectionItem & NotificationItem} newItem
+   */
+  insert (newItem) {
+    if (newItem.from > newItem.to) {
+      throw new Error('Invalid interval')
+    }
+    if (!newItem.isStreamSelection) {
+      const { notify: oldNotify } = newItem
+      // Merge notifications of intersecting items into the new item's notify function
```
ThaUnknown, line 73: shouldn't this entire thing be skipped if notify doesn't exist on the item or am I missunderstanding?

detarkende: i don't think so. even though the inserted item doesn't have a notify function, if the merged items do, then we should merge those into the inserted one.

lib/selections.js
```diff
@@ -0,0 +1,185 @@
...
+      const intersectingNotifyFunctions = []
+      for (const existing of this._items) {
+        if (existing.notify && isIntersecting(newItem, existing)) {
+          intersectingNotifyFunctions.push(existing.notify)
+        }
+      }
+      if (intersectingNotifyFunctions.length > 0) {
+        newItem.notify = () => {
+          intersectingNotifyFunctions.forEach(fn => fn())
+          oldNotify()
```
ThaUnknown, line 83: this will call undefined

ThaUnknown (approved):

SilentBot1:
Hey, just after some further testing from my side:
```javascript
const { default: WebTorrent } = await import("./index.js") 
client.add("magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent", {deselect: true})
let server = client.createServer()
server.listen(8443)
```
And running the command:
```bash
curl http://localhost:8443/webtorrent/08ada5a7a6183aae1e09d831df6748d566095a10/Sintel/Sintel.mp4 -O
```
Appears to create two stream selections:
image

Both selections are updated / destroyed as the stream progresses/closes:
image

This was the only thing I've found, everything else looks good and appears to work as expected / designed and I'll be happy to approve once the above is resolved.

detarkende:
@SilentBot1 - Thanks for the review. I tried to reproduce your situation, but I have to say, I don't know what I'm missing, but it only inserts 1 selection for me.

This is my script:
```js
import WebTorrent from './index.js'
import MemoryChunkStore from 'memory-chunk-store'

client.add('magnet:?xt=urn:btih:08ada5a7a6183aae1e09d831df6748d566095a10&dn=Sintel&tr=udp%3A%2F%2Fexplodie.org%3A6969&tr=udp%3A%2F%2Ftracker.coppersurfer.tk%3A6969&tr=udp%3A%2F%2Ftracker.empire-js.us%3A1337&tr=udp%3A%2F%2Ftracker.leechers-paradise.org%3A6969&tr=udp%3A%2F%2Ftracker.opentrackr.org%3A1337&tr=wss%3A%2F%2Ftracker.btorrent.xyz&tr=wss%3A%2F%2Ftracker.fastcast.nz&tr=wss%3A%2F%2Ftracker.openwebtorrent.com&ws=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2F&xs=https%3A%2F%2Fwebtorrent.io%2Ftorrents%2Fsintel.torrent', { deselect: true, store: MemoryChunkStore })
const server = client.createServer()
server.listen(8443)

/** @type {import('./lib/torrent.js').default} */
const torrent = client.torrents[0]

setInterval(() => {
  console.log(torrent._selections)
}, 1000)
```

And this is the output for me, when I run your `curl` command in a separate terminal instance:

image

Can you please try it again? I have rebased the repo now, but that shouldn't have an effect on this (theoretically).

I tested this with node 16, 18 and 20. Which version are you using? If you have the time, we could check in a video call or something, but I'm just out of ideas on how to reproduce this 😄

SilentBot1:
Hey @detarkende, I'm now also unable to replicate this (on Node v18.16.0) after the rebase and a `git reset --hard detarkende/2755-non-overlapping-selections`, so I'm assuming this was a skill issue on my half, but this now looks all good to go.

SilentBot1 (approved):

detarkende:
Thank you @ThaUnknown and @SilentBot1! Let's ship this thing 🚀 

@ThaUnknown, to me it shows that you have 2 unresolved threads, but I cannot find them. Could you please check and maybe reapprove?

detarkende:
Thanks everyone! It says that only those with write access to webtorrent can merge this PR, so from my end if either of you have this permission, we can merge this 😄 🚀

ThaUnknown:
it will be merged when I have the time.

ThaUnknown:
yeah add that 1 test, and I'll merge this, this is defo an A tier pull request

test/node/client-deselect.js
```diff
@@ -0,0 +1,109 @@
...
+    t,
+    onTorrent: (torrent) => {
+      lastPieceIndex = Math.floor((torrent.pieces.length - 1) / 2)
+      torrent.deselect(0, lastPieceIndex)
+    },
+    onDone: (torrent) => {
+      t.equal(torrent.pieces.filter((a) => a === null).length, (torrent.pieces.length - 1 - lastPieceIndex))
+    }
+  })
+})
```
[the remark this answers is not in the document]
detarkende: @ThaUnknown I just pushed the test you asked for, but please WAIT. I might have found an issue.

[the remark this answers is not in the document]
detarkende: @ThaUnknown I believe I have fixed the issue that I found this morning. pushed the code just now. All tests passed in my local plus I added a few new ones, so I think we're pretty much done.

ThaUnknown:
good PR, well documented, clean simple code, includes typings for static analysis

ThaUnknown:
@detarkende if you're interested in working on other things, in order of priority:
- #2465
- #2594 had a pull request, but it was 50% finished, and abandoned
- #2685
- #2082
- #2467
- #2581

I wouldn't work on types, webtorrent needs to be re-written to use async/await instead of callbacks [this has been on the roadman for years], which will break all the types anyways

ThaUnknown:
you can fix that, by re-writing this to use byte ranges internally, not pieces :)

detarkende:
@ThaUnknown yep, I'll take a look at it this week. I might have to rewrite these tests altogether because they are kind of weird anyway.

PR #2766, opened 2024-03-27
Added WebTorrent add and remove events and docs update

LostBeard:

[X] Documentation update
[  ] Bug fix
[X] New feature
[  ] Other, please explain:

Added the events 'add' and 'remove' that are fired when a torrent is added to or removed from client.torrents. 

This allows attaching to torrent events that may be emitted before the client 'torrent' event is emitted.

I am in a situation where I want to use a custom wire extension on every torrent. I initially tried to listen on the client.on('torrent') event but sometimes the Torrent 'wire' event was firing before the 'torrent' event so it was missed, and my wire extension could not be added via wire.use(). Attaching a 'wire' event handler to the torrent immediately after getting it from client.add() and client.seed() avoided this issue but it meant this had to be done to the torrent after calling client.add or client.seed. Which makes using and sharing custom wire extensions more cumbersome than it needs to be.

This is a very simple modification. I have added an update to the docs and the code.


LostBeard:
I added a WebTorrent bug fix that was made apparent while testing my initial pull request,

index.js
```diff
@@ -411,11 +412,14 @@ export default class WebTorrent extends EventEmitter {
...
     if (typeof opts === 'function') return this._remove(torrent, null, opts)
-    this.torrents.splice(this.torrents.indexOf(torrent), 1)
+    var index = this.torrents.indexOf(torrent);
+    if (index === -1) return
+    this.torrents.splice(index, 1)
     torrent.destroy(opts, cb)
     if (this.dht) {
       this.dht._tables.remove(torrent.infoHash)
     }
+    this.emit('remove', torrent);
```
SilentBot1, line 422:
```suggestion
    this.emit('remove', torrent)
```

LostBeard:
I made the requested changes. Not sure if I missed something. It says "Merging is blocked" "Merging can be performed automatically once the requested changes are addressed."

SilentBot1 (approved):

PR #2774, opened 2024-04-15
Make "always choke seeders" optional

QuixThe2nd:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Turn "always choke seeders" into an option (enabled by default).

When disabled, the client won't choke other seeders while it's seeding.

Although the average person wouldn't want to change this setting, it is useful for custom [bittorrent extensions](http://github.com/webtorrent/bittorrent-protocol) which would like to keep peers unchoked. I would appreciate a PR as my bittorrent extension still needs seeding clients to communicate.

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1199,8 +1199,7 @@ export default class Torrent extends EventEmitter {
         if (!wire.peerPieces.get(i)) return
       }
       wire.isSeeder = true
-      if (this.alwaysChokeSeeders)
-        wire.choke() // always choke seeders
+      if (this.alwaysChokeSeeders) wire.choke() // always choke seeders
     }
 
     wire.on('bitfield', () => {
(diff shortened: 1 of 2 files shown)


lib/torrent.js
```diff
@@ -88,6 +88,7 @@ export default class Torrent extends EventEmitter {
     this._destroyStoreOnDestroy = opts.destroyStoreOnDestroy || false
     this.store = null
     this.storeOpts = opts.storeOpts
+    this.alwaysChokeSeeders = opts.alwaysChokeSeeders !== undefined ? opts.alwaysChokeSeeders : true;
```
SilentBot1, line 91: This can simply become:
```suggestion
    this.alwaysChokeSeeders = opts.alwaysChokeSeeders ?? true
```

SilentBot1 (changes requested):
This looks fine to me, outside of a two small changes. Please address the suggestion and get this new option to the [docs](https://github.com/webtorrent/webtorrent/blob/master/docs/api.md#clientaddtorrentid-opts-function-ontorrent-torrent-), with a short blurb of what it does, and I'll be happy to approve this.

lib/torrent.js
QuixThe2nd: Sure I'll change it. I used `var !== undefined` because the same style is being used elsewhere in the code.
e.g. https://github.com/webtorrent/webtorrent/blob/c8ffdda0d0a03571bab2644cb5e0033d47486294/lib/torrent.js#L87C1-L87C91

SilentBot1 (approved):

QuixThe2nd:
Thank you :)

Issue #2279, opened 2022-03-17
Add JSdocs comments to improve readability in the codebase.


Bruce-Hopkins:
**What version of this package are you using?**
1.8.6

**What problem do you want to solve?**
In #2222 it was mentioned that the codebase doesn't have functions and methods don't have comments for documentation.

> I've found the codebase hard to get my head around, especially since functions aren't commented so it's hard to know what they are intended to do, and therefore how to fix them without breaking something.

I've found the same issue when adding my own contributions to the codebase. I believe that the code can benefit from [JSdocs](https://jsdoc.app/) style comments to the classes for both contributors and users.

**What do you think is the correct solution to this problem?**
To add JSdocs style documentation to all (or most) of the functions and methods in the codebase.

**Are you willing to submit a pull request to implement this change?**
Gladly!

DiegoRBaquero:
I'd be happy to have code docs! If you are up for it, I'm happy to review

Bruce-Hopkins:
Thanks! I'll get started on it.

Bruce-Hopkins:
@DiegoRBaquero I'm currently working on the PR to implement more docs but I'm not sure which methods I should document,

I noticed a common pattern in the codebase. Many of the methods which start with underscores don't have jsdocs styles comments for them:

https://github.com/webtorrent/webtorrent/blob/38a6c72a367686c5d919768aab1e603d72058220/index.js#L477

While the ones without underscores do:

https://github.com/webtorrent/webtorrent/blob/38a6c72a367686c5d919768aab1e603d72058220/index.js#L468-L472

Is there a reason for this? Should I add comments to both?

PR #2787, opened 2024-05-15
feat: add TypeScript defination files

### LINKED ISSUE #2782: wrong Return Value type of client.get() in @types/webtorrent packages
## What version of this package are you using?

`@types/webtorrent@0.109.8`

## What operating system, Node.js, and npm version?

- OS: Windows 10 Pro 22H2
- Node.js: 18.x
- PNPM: 9.x

## What happened?

Latest WebTorrent has an async `client.get()` method:

https://github.com/webtorrent/webtorrent/blob/41cb62b3b7095742f553de169bf1ce8e319ea6cb/docs/api.md?plain=1#L231-L233

but its Type Defination package has no `Promise` type on the return value of `client.get()`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0474990ae3c1b3745cc7bd6dd83f69c2f574f4de/types/webtorrent/index.d.ts#L136

## What did you expect to happen?

The Type Defination package has the same type of the Implement package.

## Are you willing to submit a pull request to fix this bug?

I may submit a PR after this bug being confirmed by @webtorrent team.

TechQuery:
## What is the purpose of this pull request?

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Other, please explain:

## What changes did you make?

- Add built-in TypeScript defination to replace outdated `@types/webtorrent` package: DefinitelyTyped/DefinitelyTyped#69619

## Which issue (if any) does this pull request address?

- fix #2782

## Is there anything you'd like reviewers to focus on?

We may ignore the `dist` folder in Git, and only publish it to NPM, then check published files in CDN pages:
1. https://www.npmjs.com/package/webtorrent?activeTab=code
2. https://www.jsdelivr.com/package/npm/webtorrent?tab=files&path=dist


SilentBot1:
Hey @TechQuery,

Not to discourage you from what is in the PR, which is great to see, but a lot of these types appear to be half-complete, or too vague to be helpful to end users, e.g:

`constructor(opts?: {});` is an object but with none of the actual options defined,
`declare const VERSION: any;` which is a string,
`torrentPort: any;` which is a number,
`tracker: any;` which can be false, true, null, or any of the required / optional options described in bittorrent-tracker [here](https://github.com/webtorrent/bittorrent-tracker?tab=readme-ov-file#client).

Since many of these types are from child packages, these can likely be ignored for now, but ensuring the types that are actually from WebTorrent (not the dependencies) are both useful and accurate should be a priority for this PR.

If you have any questions on the above, please don't hesitate to let me know.

TechQuery:
> a lot of these types appear to be half-complete, or too vague to be helpful to end users

@SilentBot1 I did know those issues, but I just use `tsc` to generate Type Defination from JS codes with incomplete JSDoc annotation, which needs to be added first. For example:
https://github.com/webtorrent/webtorrent/blob/320f9296fca2315d6d775af7e90f1dc67c91186d/index.js#L45-L48

package.json
```diff
@@ -164,6 +166,7 @@
     "build-chromeapp": "webpack --config ./scripts/chromeapp.webpack.js",
     "build-js": "webpack --config ./scripts/browser.webpack.js",
     "build-js-worker": "webpack --config ./scripts/worker.webpack.js",
+    "build-d-ts": "rm -rf dist/**/*.d.ts  &&  tsc --emitDeclarationOnly",
```
SilentBot1, line 169: I recommend using something like `npx rimraf -g dist/**/*.d.ts` to ensure all scripts are cross-platform.

SilentBot1:
Sorry @TechQuery, I misunderstood what the intentions for this PR were, I'll give this a review now.

package.json
SilentBot1, line 169: This should likely also be added to the script `build` on line 165, e.g.
```
"build": "npm run build-js && npm run build-js-worker && npm run build-chromeapp && npm run build-d-ts",
```
To ensure it's built during a release, we'll also need to create a PR [here](https://github.com/webtorrent/semantic-release-config/blame/411b9e580dcd805eee651b91f9483d838c2a7fbd/.releaserc.js#L37) to add `dist/**/*.d.ts` to `assets` to ensure the built files are pushed back to GitHub.

SilentBot1 (commented):
Overall this looks good to me.

@ThaUnknown do you have any opinion on this? e.g. should the `.d.ts` files be placed into a separate `types` folder, rather than `dist`, so we can be more consistent across the other repos?

package.json
TechQuery: > To ensure it's built during a release, we'll also need to create a PR [here](https://github.com/webtorrent/semantic-release-config/blame/411b9e580dcd805eee651b91f9483d838c2a7fbd/.releaserc.js#L37) to add `dist/**/*.d.ts` to `assets` to ensure the built files are pushed back to GitHub.

Does it support Glob Pattern? We have lots of `.d.ts`.

ThaUnknown:
yeah, these are really bad because a lot of JSDoc is missing, I'd likely not approve this until the JSDoc is improved, which also would require a bunch of other changes, I think this PR is quite premature

ThaUnknown:
the more I think about it, the less sense this PR makes, you generate types from JSDoc.... just use the JSDoc? it provides the same functionality to end developers? I think what you actually want to do is fix @definetlytyped, which you should simply make a PR to remove the outdated types so editors like vscode don't inherit the types automatically

ThaUnknown:
@TechQuery yeah, I'm closing this in favor of plain JSDoc https://github.com/webtorrent/webtorrent/issues/2279

Feel free to make a PR on definetlytyped repo to remove webtorrent types, and I'll give it a +1

TechQuery:
> I think what you actually want to do is fix @definetlytyped, which you should simply make a PR to remove the outdated types

@ThaUnknown I tried first, but failed: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69619#issuecomment-2113220261

ThaUnknown:
no, you tried updating them, not removing them

Issue #2777, opened 2024-04-20
Do not use the old `import ... assert` syntax.


nicolo-ribaudo:
This package is using the legacy "import assertions" proposal, to import a JSON file in https://github.com/webtorrent/webtorrent/blob/c8ffdda0d0a03571bab2644cb5e0033d47486294/index.js#L20.

That proposal now uses a different syntax: `import info from './package.json' with { type: 'json' }`. The old syntax is only supported in Node.js and Chrome, and they are both planning to remove support for it (Node.js 22 next week, and Chrome 126 in June), so it will make this package not work anymore.

For more info, see https://github.com/tc39/proposal-import-attributes/issues/135.

ThaUnknown:
it was used because it was compatible with webpack, while `with` wasnt, feel free to PR a fix

PR #2797, opened 2024-05-31
Replace import ... assert calls with import ... with

ripdog:
Fixes compatibility with Node 22

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

I replaced the deprecated(?) `assert` statement with `with`. This has fixed use of the lib with Node 22. I haven't tested it any further.
#2777
Mentioned in the issue, webpack compatibility might be an issue?

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ import ConnPool from './lib/conn-pool.js' // browser exclude
 import Torrent from './lib/torrent.js'
 import { NodeServer, BrowserServer } from './lib/server.js'
 
-import info from './package.json' assert { type: 'json' }
+import info from './package.json' with { type: 'json' }
 const VERSION = info.version
 
 const debug = debugFactory('webtorrent')
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -35,7 +35,7 @@ import utp from './utp.cjs' // browser exclude
 import WebConn from './webconn.js'
 import { Selections } from './selections.js'
 
-import info from '../package.json' assert { type: 'json' }
+import info from '../package.json' with { type: 'json' }
 
 const debug = debugFactory('webtorrent:torrent')
 const MAX_BLOCK_LENGTH = 128 * 1024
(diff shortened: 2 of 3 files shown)


ThaUnknown:
seems like node16 doesnt support this syntax, so this PR is gonna sit for at least a year or two until node16 is no longer supported

ripdog:
Sorry, I'm not sure I understand. Node 16 was unsupported on 2023-09-11, according to Wikipedia. Is Webtorrent supporting old and potentially insecure Node versions, encouraging users to continue using them against the recommendation of the Node developers? 

Or is this an attempt to stay compatible with stable Debian or similar super-slow distros? In that case, surely it is the distro packagers' responsibility to ship old versions of webtorrent and take on the security maintenance burden themselves? That's how it is for basically all 'traditional' libraries, as I understand it. The library stays compatible with current versions of their runtime/dependencies, and distro packagers who want to ship old runtime/dependencies take up the slack themselves. 

I want to point out that webtorrent is currently broken on current versions of node, and gives an incredibly cryptic error message. Rolling release distro? Broken. Windows/Mac user who downloads latest Node/NPM from nodejs.org? Broken.

ThaUnknown:
No longer supported by WebTorrent, up untill a few days ago, WebTorrent supported node 12. It's not encouraged use, it's supporting legacy implementations, once you work with actual software you'll understand. I'd prefer if you wouldn't use manipulative messages in your future responses.

WebTorrent works fine on LTS of node [20.14.0], so I'm not sure what you're trying to say, unless you mean node22, by which I'll reply with are you suggesting that WebTorrent should support "potentially insecure Node versions"?

ripdog:
It was not my intention to use 'manipulative messages'. Your library is currently broken with the current version of its runtime. You justify keeping it broken by wishing to keep it running on versions of Node which have been abandoned and deprecated by *the Node developers*. If you're going to talk about 'actual software', perhaps you should go talk to the Node team about how their deprecation policy makes them not 'actual software'.

And why are you implying that the latest, currently supported Node version is 'potentially insecure'?

There's nothing wrong with people using old versions of software if they want to and understand the risks. But why is webtorrent going so far to encourage people to use old runtime versions? It would be perfectly fine to leave a branch for distro maintainers who want to ship Node 16 and keep the main branch functioning with the currently supported versions of Node?

PR #2805, opened 2024-06-07
fix: Inability to default to the utp peer type

SilentBot1:

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Other, please explain:

Currently, due to `host`only being defined when `client.blocked` is truthy and `client.blocked` only being truthy when `opts.blocklist` is passed to the construction of the client, we could only successfully default to utp in edge cases.

https://github.com/webtorrent/webtorrent/blob/45b3b1acd7b45a184ecc6f93ed33d94c21ae430b/lib/torrent.js#L843-L870
https://github.com/webtorrent/webtorrent/blob/45b3b1acd7b45a184ecc6f93ed33d94c21ae430b/index.js#L170-L183

In this PR, I make changes to ensures we always populate `host`, which allows the correct peer type to be defaulted.

### Note: Defaulting to uTP in WebTorrent's current implementation, if a peer only supports TCP, it would take >21 seconds to fall back to a TCP connection. Due to this, I would recommend holding off this for the moment.


ThaUnknown:
this isn't a new feature, but a bugfix, the 21s wait for `tcpOutgoing` is annoying, however that was the original intention with the library, non-utp connections are rare enough that this doesn't matter, the tcp connection delay should be addressed in a separate PR

ThaUnknown (approved):

PR #2843, opened 2024-07-29
fix: custom length file buffer

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

added support for custom lengths for file buffer, previously this hardcoded file length

the math, needs to be verified


SilentBot1 (approved):
The math looks good from my local testing.

PR #2851, opened 2024-08-04
feat: drop IDB

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

Dropped hybrid chunk store, to use FSA chunk store directly, this drops IDB support.

Lets assume every1 updates their browsers, as all browsers since recently support OPFS https://caniuse.com/mdn-api_storagemanager_getdirectory putting us at 94% coverage.

safari doesn't support createWritable https://caniuse.com/mdn-api_filesystemfilehandle_createwritable so it is polyfilled via some hacky code

FSA offers waaaaaay more storage capacity, way better speeds and way smaller bundle sizes [if not for safari].


SilentBot1 (approved):
I've tested using Safari on OSX and Chrome on Windows, and both are working as expected, with small and larger files. 

LGTM!

PR #2870, opened 2024-10-02
Added an api method to add with timeout #2691

uriva:

[ ] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

1. added support for timeout in the api
2. avoid throw exceptions on innocuous actions


ThaUnknown:
this can be done externally by a developer by listening to the `ready` event, this pr also does bad stuff like version locking random commits and changing styling where it shouldnt, no.

uriva:
What is "the ready event"? I don't see anything in the docs about this, so how would one go about using this?

uriva:
Regarding the other things, I think they can be fixed if  you want to cooperate.

uriva:
Found the ready event
Thanks

PR #2878, opened 2024-10-29
Add an option to store the bitfield in the file system

maury91:

[x] Documentation update
[ ] Bug fix
[x] New feature
[ ] Other, please explain:

Add an option to store the bitfield in the file system, the bitfield will be used instead of hashing in verifyPieces.
This is several orders of magnitude faster compared to hashing all the pieces to know what pieces are missing, it has two drawbacks:
1. It creates an extra file in the file system
2. If an external program changes the content of the downloaded files the final result could become corrupted


ThaUnknown:
this is a no, mainly because this won't work in browser, however this PR could be modified to accept a pre-loaded bitfield, as a torrent create parameter, this would allow a developer to do:

```js
const field = client.torrents[0].bitfield
await client.torrents[0].destroy()
await client.add(infoHash, { bitfield: field })
```
or smth similar

ThaUnknown:
additionally with custom bitfield, webtorrent should check if a file exists, cuz its possible that passing a full bitfield while a file doesnt exist, could attempt to read a file that doesnt exist, that might need to be fixed on the chunk store's side tho

maury91:
agree on the solution that is more browser friendly, however, there are two more things to take in consideration:

1. The developer will need to listen to the `verified` event in order to save the bitfield to a more long-term place ( I'm thinking this solution is useful in case the process quits unexpectly )
2. Webtorrent should add a second argument to the`verified` event during warmup to prevent backfeeding

maury91:
I changed the implementation to allow a custom bitfield, to verify if the file exists I check if the first existing piece of that is correct.

At most, it will check one piece for each file, if we want to save up more time the hashing for each one of this pieces can be removed

ThaUnknown:
adding the `isStartup` is redundant, you can instead just do:
```js
torrent.on('verified', index => {
  const isStartup = !torrent.ready
  if (isStartup) return // ignore
  // do stuff with index
}
```
I also think the initial bitfield should be a buffer, not a full bitfield, but I'm uncertain if that's a good idea implementation wise.

if a store verification fails, it shouldn't error and destroy the entire torrent, just mark the respective file as unverified [WATCH OUT FOR PIECE OVERLAP]

there's a LOT of code overlap between verify pieces and verify pieces using bitfield, can this be de-duplicated?

overall this is a good PR, and a good feature, just needs some slight polishing

maury91:
It's not destroying the torrent when the bitfield verification falls, the error is catched in the final callback of parallelLimit, there it checks for error, disables the bitfield and falls back to normal verification, in this way if there's an incongruence it just assumes the bitfield is not accurate and can't use it for verification so it has to resort to hashing every piece

lib/torrent.js
```diff
@@ -638,24 +639,70 @@ export default class Torrent extends EventEmitter {
...
-        if (this.destroyed) return cb(new Error('torrent is destroyed'))
+  _verifyPiecesUsingBitfield (cb) {
+    const piecesToCheck = new Set()
+    this.files.forEach(file => {
+      let checkFile = true
+      for (let i = file._startPiece; i <= file._endPiece; ++i) {
+        if (this._startupBitfield.get(i)) {
+          if (checkFile) {
+            piecesToCheck.add(i)
+            checkFile = false
```
ThaUnknown, line 670: shouldn't this have a break? I think the point of this is to look if any of the bitfield pieces are in the file's range, once we find one we don't need to continue iterating? that allows u to get rid of `checkFile`

lib/torrent.js
```diff
@@ -638,24 +639,70 @@ export default class Torrent extends EventEmitter {
...
+        if (!isVerified) {
+          this._startupBitfield = null
+          return cb(new Error('piece is invalid, force recheck'))
+        }
+        cb(null)
+      })
+    }), FILESYSTEM_CONCURRENCY, (err) => {
+      if (err) {
+        this._startupBitfield = null
+        return this._verifyPieces(cb)
```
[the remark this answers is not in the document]
maury91: The question is "Why this file differ from what the bitfield says?", it can have two answers:
1. This specific file has been modified
2. The bitfield is not the correct one for this torrent

I think we may need more check, first of all the length of the bitfield, it should be `Math.ceil(pieces.length / 8)`
By verifying that the bitfield is the correct length the second answer is more improbable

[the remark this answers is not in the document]
ThaUnknown: oh yeah, you're right, it should defo check if the bitfield length matches the torrent's bitfield size, if it doesn't don't even run any of this code

as for file being modified, this doesn't check that, realistically it only checks if it exists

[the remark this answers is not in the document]
maury91: It does check that the first downloaded piece of the file matches the expected hash, it isn't perfect but the chance of two totally different files with same hash in the same piece is extremely low

lib/torrent.js
maury91: My bad, after re-reading the code we don't want a break there because it needs to continue to mark the rest of the pieces of the file as valid or no

ThaUnknown: oh right, i see, i'm stupid, I thought you did this differently, you copy the contents of the startup bitfield into the torrent's bitfield, imho the torrent's bitfield should be the startup bitfield, and if something goes wrong the pieces from it should be invalidated, this is also why i think the initial bitfield should be a buffer, the developer doesnt really care about the bitfield class when he wants to store it, just it's buffer

ThaUnknown:
https://github.com/webtorrent/webtorrent/blob/1b631cd226ef66202fbfbbc695019e6c6183a22f/lib/torrent.js#L582

```js
    this.bitfield = new BitField(this._startupBitfield ?? this.pieces.length)
```

lib/torrent.js
```diff
@@ -638,24 +639,71 @@ export default class Torrent extends EventEmitter {
...
+      for (let i = file._startPiece; i <= file._endPiece; ++i) {
+        if (this.bitfield.get(i)) {
+          if (checkFile) {
+            piecesToCheck.add(i)
+            checkFile = false
+          }
+          this._debug('piece verified %s', i)
+          this._markVerified(i)
+        } else {
+          this._markUnverified(i)
```
ThaUnknown, line 676: marking it as unverified isnt necessary? you only need a `this.bitfield.set(i, false)` rest can be skipped

lib/torrent.js
```diff
@@ -638,24 +639,71 @@ export default class Torrent extends EventEmitter {
...
+          this._debug('piece invalid %s', i)
+        }
+      }
+    })
+    parallelLimit([...piecesToCheck].map((index) => cb => {
+      this._verifyPiece(index, (err, isVerified) => {
+        if (err) return cb(err)
+        if (!isVerified) {
+          this._startupBitfield = null
+          return cb(new Error('piece is invalid, force recheck'))
```
ThaUnknown, line 686:
```suggestion
          return cb(index)
        }
        cb(null)
      })
    }), FILESYSTEM_CONCURRENCY, (err) => {
      if (err) {
        const filesToRecheck = []
        for (const index of err) {
          for (const file of this.files) {
            if (file.includes(err)) {
              filesToRecheck.push(file)
            }
          }
        }
        return rescanSpecificFiles(filesToRecheck, cb)
      }
      cb(null)
```

I think something like this *might* work?

ThaUnknown: actually something along the lines of
```js
let checkFile = 2
if (checkFile) {
  piecesToCheck.add(i)
  if (checkFile-- === 1) {
    piecesToCheck.delete(i - 1)
  }
}
```
could probably do the job

ThaUnknown:
still needs test, I think the best way is to create a few big files of garbage, write them to a folder, create a torrent from them via `createTorrent`, and add the torrent, with a store path to the folder you wrote into, then you can create a test of having the all the files but skipping the rescan with a preloaded bitfield, then missing one of the file but having the rest should only rescan the file [you can check bitfield and check that only values for the one file were false], then having all files, but bitfield missing a bunch of pieces should fill the bitfield, [I think this test fails currently because pieces not existing in the initial bitfield aren't checked]

ThaUnknown:
I'm starting to debate if so much safety is actually a good idea, this is more logic than it's worth, while adding more and more safety looses it's benefits of speed... I think I had a bad approach to this

ThaUnknown:
sorry for being so strict, this is a really good feature and I want to make sure it's done right, it's something i thought of a while ago myself, so I'm very enthusiastic about some1 else actually doing it

maury91:
Don't worry, it's a good thing being strict, at the moment I'm on holiday so I didn't got a chance to address the new comments.

So what we want is:

1. Do a check on the second piece of the file if is bigger than one piece ( or at least a piece that does not overlap )
2. If the check fails because the flle does not exists or the content is different only that file will be re-checked

ThaUnknown:
yep.

lib/torrent.js
maury91: I rewrote it differently:

```js
const piecesToCheck = new Set()
    const piecesToFilesMap = new Map()
    this.files.forEach(file => {
      const previouslyVerifiedFilePieces = []
      for (let i = file._startPiece; i <= file._endPiece; ++i) {
        if (this.bitfield.get(i)) {
          previouslyVerifiedFilePieces.push(i)
          if (!piecesToFilesMap.has(i)) {
            piecesToFilesMap.set(i, [])
          }
          piecesToFilesMap.get(i).push(file)
          this.emit('verified', i)
        } else {
          this._markUnverified(i)
          this._debug('piece invalid %s', i)
        }
      }
      const nonOverlappingPiece = previouslyVerifiedFilePieces.find(piece => piecesToFilesMap.get(piece).length === 1)
      if (nonOverlappingPiece) {
        piecesToCheck.add(nonOverlappingPiece)
      } else if (previouslyVerifiedFilePieces.length >= 1) {
        piecesToCheck.add(previouslyVerifiedFilePieces[0])
      }
    })
```

it is more complex but it prioritizes truly non-overlapping pieces.

Also, I don't see a big problem with overlapping pieces, it is verifying them using the hashes and not just their existence, so if a piece has the correct hash and spans across multiple files all those files are correct.

Considering the most probable use ( start and stop of an application that uses webtorrent and re-loads partially downloaded torrents ) the chances of someone tampering with the contents of those files are close to 0, and what truly can happen is that the download folder will be deleted

maury91:
I added the logic we discussed and also some JSDoc comments, the JSDoc comments did have a positive effect on the IDE, and I could look in the future in a progressive re-write in Typescript

lib/torrent.js
```diff
@@ -638,24 +704,110 @@ export default class Torrent extends EventEmitter {
...
+   * @private
+   */
+  _verifyPiecesUsingBitfield (cb) {
+    const piecesToCheck = new Set()
+    const piecesToFilesMap = new Map()
+    this.files.forEach(file => {
+      const previouslyVerifiedFilePieces = []
+      for (let i = file._startPiece; i <= file._endPiece; ++i) {
+        if (this.bitfield.get(i)) {
+          previouslyVerifiedFilePieces.push(i)
```
ThaUnknown, line 751: wow i hate this, actually, i despise this, this is way too much, i have 0 clue what even is happening here at a glance, idk why you wrote it this way

lib/torrent.js
```diff
@@ -638,24 +704,110 @@ export default class Torrent extends EventEmitter {
...
+      }
+      const nonOverlappingPiece = previouslyVerifiedFilePieces.find(piece => piecesToFilesMap.get(piece).length === 1)
+      if (nonOverlappingPiece) {
+        piecesToCheck.add(nonOverlappingPiece)
+      } else if (previouslyVerifiedFilePieces.length >= 1) {
+        piecesToCheck.add(previouslyVerifiedFilePieces[0])
+      }
+    })
+    const filesToCheck = new Set()
+    parallelLimit([...piecesToCheck].map((index) => cb => {
```
ThaUnknown, line 770: mmm sucks u cannt use this._verifyPiecesUsingHash here

lib/torrent.js
ThaUnknown: I liked the 2nd piece approach better, because the only pieces that can be overlapping is first or last, 2nd means its not first, if its last, oh well you cant do anything about it anyways, but at least we know its not first

ThaUnknown:
> I added the logic we discussed and also some JSDoc comments, the JSDoc comments did have a positive effect on the IDE, and I could look in the future in a progressive re-write in Typescript

yeah, I also considered a TS rewrite, actually started a bit, but man, webtorrent is NOT build for that, it does so many assumptions about types its insane, the lib/torrent.js alone is insanely hacky

lib/torrent.js
maury91: The problem is that we don't know exactly what is downloaded in advance so we can't really check the second piece.

Imagine the file is split in 10 pieces, and the download mode is rarest first, and it fully downloaded 3rd piece and 5th piece. The 2nd piece is not downloaded so we can't use it for verification, and the second of the downloaded pieces is in reality the 5th piece.

The way I wrote it:

1. Find what pieces are downloaded for that file ( in our example it would be 3rd and 5th )
2. Find the first piece that is not overlapping ( it would be the 3rd aka 1st )
3. Fallback case, if only one piece is downloaded use that one even if overlaps

ThaUnknown: why....? the if the 2nd piece is in the preloaded bitfield, then it has been downloaded, if its not then it will check if the file exists using 1st piece

ThaUnknown: sorry, my code doesnt check the 2nd piece explicity, just a `2nd piece for the file` we dont know which piece in the file it is, it could be the 14th piece, because the file only has 10th piece and 14th piece downloaded, that's fine too

maury91: I'm thinking of the edge casa when we have a single piece. 

The other solution would be to collect these pieces and if we have more than one we use the second one, in the case we have only one piece we use the first one.

I used a reverse map to know if the piece is multiple files for the edge case when the file has only one single piece downloaded and it overlaps ( it could be because we downloaded only the first piece or the file is smaller than one piece ( I imagine that in a torrent with a 4K movie and various subtitles, the subtitles files could fit all in one file ))

ThaUnknown: then it never reaches the 2nd part of the check which replaces the first piece with the second?

maury91: It should be able to reach the last condition in the case where there's only one piece or all the piece overlap ( first and last piece downloaded )

ThaUnknown: yes, what i suggest, adds the first piece to check, then if the 2nd piece exists on the next iteration, remvoes the first piece and adds the 2nd piece in its place

ThaUnknown: what you got *could* be better? i'm not sure because the code is so convoluted i can't really see what it does at a glance

maury91: I'm thinking of a way to make the code very understandable


```js
// First step, optimistically mark what is verified and what is not by blindly trusting the bitfield
    this.files.forEach(file => {
      for (let i = file._startPiece; i <= file._endPiece; ++i) {
        if (this.bitfield.get(i)) {
          this.emit('verified', i)
        } else {
          this._markUnverified(i)
          this._debug('piece invalid %s', i)
        }
      }
    })
    // Second step, we construct support structures to choose what pieces should be slowly verified using hashing
    const piecesToFilesMap = new Map()
    // We are going to use this to discover what pieces are unique to each file
    const downloadedPiecesByFile = this.files.map(file => {
      const downloadedPieces = []
      for (let i = file._startPiece; i <= file._endPiece; ++i) {
        if (this.bitfield.get(i)) {
          if (!piecesToFilesMap.has(i)) {
            piecesToFilesMap.set(i, [])
          }
          piecesToFilesMap.get(i).push(file)
          downloadedPieces.push(i)
        }
      }
      return downloadedPieces
    })
    // Third step, using the support structures we are going to choose what pieces to verify
    const uniquePiecesByFile = downloadedPiecesByFile.map((downloadedPieces) => {
      // If there's only one piece we don't care if it's unique
      if (downloadedPieces.length === 1) {
        return downloadedPieces
      }
      // piecesToFilesMap tells us how many files share that piece, if it's only one than it's unique to that file
      return downloadedPieces.filter(piece => piecesToFilesMap.get(piece).length === 1)
    })
    const piecesToCheck = new Set(uniquePiecesByFile.map(pieces => pieces[0]))
```

I believe this is very robust and at the same time should be understandable

lib/torrent.js
maury91: I would love to, but the logic inside is different, `_verifyPiecesUsingHash` just sets the piece as verified or not, but the logic in line 770 is to check if the piece that the bitflied says it's verified is truly verified

ThaUnknown: yep, i'm just making this for myself while thinking it over

lib/torrent.js
ThaUnknown: that's way too much code, way too many iterations for something way too simple, i'm telling you, simply
``js
for (const file of this.files) {
  for (let i = file._startPiece; i <= file._endPiece; ++i) {
    let checkFile = 2
    if (checkFile) {
      piecesToCheck.add(i)
      if (checkFile-- === 1) {
        piecesToCheck.delete(i - 1)
      }
    }
  }
}
```
will do most of the same work, while being loads faster and easier to understand

File: lib/torrent.js (hunk 6 above)
Comment (maury91) [reply lib/torrent.js:770]: To re-use it there's the option to have `_verifyPiecesUsingHash` return in the callback a boolean array with the verification status

```js
 this._verifyPiecesUsingHash([...piecesToCheck], (err, piecesVerified) => {
      if (err) {
        return cb(err)
      }
      piecesVerified.forEach((isVerified, index) => {
        if (!isVerified) {
          const filesOnPiece = piecesToFilesMap.get(index)
          filesOnPiece.forEach(file => {
            filesToCheck.add(file)
          })
        }
      })
      // Fifth step, if we need to recheck files we are going to recheck them
      if (filesToCheck.length) {
        const piecesToRecheck = []
        filesToCheck.forEach((file) => {
          for (let i = file._startPiece; i <= file._endPiece; ++i) {
            if (piecesToRecheck.indexOf(i) === -1) {
              piecesToRecheck.push(i)
            }
          }
        })
        return this._verifyPiecesUsingHash(piecesToRecheck, cb)
      }
      cb(null)
    })
```

lib/torrent.js
maury91: > piecesToCheck.delete(i - 1)

We have no insurance it was the piece immediately before, so we need another support variable for it

maury91: This should do the trick:

```js
const piecesToCheck = new Set()
    // First step, optimistically mark what is verified and what is not by blindly trusting the bitfield
    this.files.forEach(file => {
      let checkFile = 2
      let pieceToCheckForThisFile = null
      for (let i = file._startPiece; i <= file._endPiece; ++i) {
        if (this.bitfield.get(i)) {
          if (checkFile) {
            pieceToCheckForThisFile = i
            checkFile--
          }
          this.emit('verified', i)
        } else {
          this._markUnverified(i)
          this._debug('piece invalid %s', i)
        }
      }
      piecesToCheck.add(pieceToCheckForThisFile)
    })
```

maury91:
I rewrote it using the simpler logic of just checking the "second" piece, it also uses _verifyPiecesUsingHash

lib/torrent.js
```diff
@@ -36,6 +36,43 @@ import { Selections } from './selections.js'
...
+/**
+ * @typedef TorrentOpts
+ * @type {object}
+ * @property {Array<string>=} announce - Torrent trackers to use (added to list in .torrent or magnet uri)
+ * @property {Array<string>=} urlList - Array of web seeds
+ * @property {string=} path - Folder to download files to (default=`/tmp/webtorrent/`)
+ * @property {boolean=} addUID - (Node.js only) If true, the torrent will be stored in it's infoHash folder to prevent file name collisions (default=false)
+ * @property {FileSystemDirectoryHandle=} rootDir - *(browser only)* if supported by the browser, allows the user to specify a custom directory to stores the files in, retaining the torrent's folder and file structure
+ * @property {boolean=} skipVerify - If true, client will skip verification of pieces for existing store and assume it's correct
+ * @property {Uint8Array|Buffer|Array<number>=} bitfield -  Preloaded numerical array/buffer to use to know what pieces are already downloaded (any type accepted by UInt8Array constructor is valid)
```
ThaUnknown, line 56: Uint8Array, not Buffer, never rely on buffer if possible here, because wt runs in browser, which doesnt have buffer

lib/torrent.js
```diff
@@ -638,24 +704,117 @@ export default class Torrent extends EventEmitter {
...
+   * @private
+   */
+  _verifyPiecesUsingBitfield (cb) {
+    const piecesToCheck = new Set()
+    const piecesToFilesMap = new Map()
+    // First step, optimistically mark what is verified and what is not by blindly trusting the bitfield
+    // and construct a list of pieces to verify ( max 1 piece for each file, in some edge cases that piece could overlap )
+    this.files.forEach(file => {
+      let checkFile = 2
+      let pieceToCheckForThisFile = null
```
ThaUnknown, line 751: good catch, I overlooked this!

lib/torrent.js
maury91: However Node has it, so this type has some types for the browser and some for Node, ideally it should overlap with the types allowed in `new UInt8Array(<<<here>>>)`

lib/torrent.js
maury91: Rewrote it a different way, we don't need the data in the callback cause we already have the info in the mutated bitfield

lib/torrent.js
```diff
@@ -638,24 +704,117 @@ export default class Torrent extends EventEmitter {
...
+      }
+      piecesToCheck.add(pieceToCheckForThisFile)
+    })
+    // Second step, for each piece that needs to be verified we verify it using hashing
+    this._verifyPiecesUsingHash([...piecesToCheck], (err, piecesVerified) => {
+      if (err) {
+        return cb(err)
+      }
+      const filesToCheck = new Set()
+      piecesVerified.forEach((isVerified, index) => {
```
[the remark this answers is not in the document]
maury91: addressed

lib/torrent.js
maury91: This is from `lib.es5.d.ts` ( Typescript internals )

```
    new (length: number): Uint8Array;
    new (array: ArrayLike<number> | ArrayBufferLike): Uint8Array;
    new (buffer: ArrayBufferLike, byteOffset?: number, length?: number): Uint8Array;
```

I'm sure we don't want it to be a simple number cause it would make no sense ( even if it would work ), so the correct types are:

`ArrayLike<number` and `ArrayBufferLike`

```ts
interface ArrayLike<T> {
    readonly length: number;
    readonly [n: number]: T;
}

interface ArrayBufferTypes {
    ArrayBuffer: ArrayBuffer;
}
type ArrayBufferLike = ArrayBufferTypes[keyof ArrayBufferTypes];
```

maury91: I changed `Buffer` to `ArrayBuffer` cause it's a superset

ThaUnknown: Buffer is Uint8Array with some extra functions, buffer extends uint8array, which extends typedarray, arraybuffer is a representation of data that cannot be accessed without a view, buffer doesnt exist in browsers so as a rule of thumb uint8array is used everywhere in webtorrent

docs/api.md
```diff
@@ -125,6 +125,7 @@ If `opts` is specified, then the default options (shown below) will be overridde
   storeOpts: Object,         // Custom options passed to the store
   addUID: Boolean,           // (Node.js only) If true, the torrent will be stored in it's infoHash folder to prevent file name collisions (default=false)
   skipVerify: Boolean,       // If true, client will skip verification of pieces for existing store and assume it's correct
+  bitfield: Buffer,          // Preloaded numerical array/buffer to use to know what pieces are already downloaded (any type accepted by UInt8Array constructor is valid)
```
ThaUnknown, line 128: update here too

lib/torrent.js
```diff
@@ -555,11 +612,11 @@ export default class Torrent extends EventEmitter {
 
     // Select only specified files (BEP53) http://www.bittorrent.org/beps/bep_0053.html
     if (this.so && !this._startAsDeselected) {
-      this.files.forEach((v, i) => {
+      for (const file of this.files ) {
```
ThaUnknown, line 615: you mass replaced, and broke this, here the `i` from forEach was used

maury91: good catch, I used a macro and didn't notice it wasn't using `file` but the index

docs/api.md
maury91: done, it's `UInt8Array` now

ThaUnknown:
great, i see no issues with this, just tests now and we can merge this

ThaUnknown:
this seems to fail tests :/

ThaUnknown:
any updates on this?

maury91:
I need to check why the test hangs, unfortunately, I don't have much time to dedicate to it, I will get into fixing and creating tests locally when I have more bandwidth

ThaUnknown:
ah, unfortunate, sorry for pressing then

ThaUnknown:
boo, bad bot, i'll fix this soon

ThaUnknown (approved):

maury91:
Thanks for taking care of it! I'm sorry I couldn't do it myself, my free time simply dropped to near zero in the last months

ThaUnknown:
nah you did the bulk of the work, i just had to write the tests and fix 1 line of code u accidentally broke, thanks for the pr

PR #2892, opened 2024-11-17
fix: uTP error on setup

ThaUnknown:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

it's possible for uTP to error on `listen` which disables utp, however it hangs the client indefinetly, as onListening never fired, this fixes that, the errors were also swallowed, thats not a good thing, tcp server logs its error on destroy, however utp server didnt! this was a bad implementation!
onListening is safe to fire right away on utp error, as utp can only error once tcp server has already started listening

### DIFF
--- a/lib/conn-pool.js
+++ b/lib/conn-pool.js
@@ -42,8 +42,10 @@ export default class ConnPool {
       this._client._destroy(err)
     }
 
-    this._onUTPError = () => {
+    this._onUTPError = (err) => {
       this._client.utp = false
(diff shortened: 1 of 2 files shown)


ThaUnknown:
this is also likely the cause of some tests hanging for a long time without erroring, this will resolve that issue, likely help us track down the issue in utp-native in the future, as the error is quite rare

SilentBot1 (approved):

PR #2906, opened 2024-12-21
Fix protocol encryption stalling downloads

vishwamartur:
Related to #2594

Fix protocol encryption stalling downloads in WebTorrent.

* Modify `lib/peer.js` to include a call to `this.wire._determineHandshakeType()` after sending `Pe1` in the `onConnect` method.
* Add a new test file `test/node/protocol-encryption.js` to verify protocol encryption.
* Add a test case to check if enabling protocol encryption does not stall downloads using two WebTorrent instances connecting to each other with the `secure` option enabled.

### DIFF
--- a/lib/peer.js
+++ b/lib/peer.js
@@ -127,9 +127,15 @@ export default class Peer extends EventEmitter {
 
     if (this.swarm) {
       if (this.type === 'tcpOutgoing') {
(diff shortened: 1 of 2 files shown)


ThaUnknown:
unfortunately your tests fail

PR #2926, opened 2025-02-13
log warning instead of exception when calling critical on destroyed torrent

### LINKED ISSUE #2923: exception
2.5.17

ubuntu node 22.11.0

**What happened?**

```
Error: torrent is destroyed
    at Torrent.critical (file:///node_modules/webtorrent/lib/torrent.js:1078:31)
    at pump (file:///node_modules/webtorrent/lib/file-iterator.js:54:32)
    at Torrent.listener (file:///node_modules/webtorrent/lib/file-iterator.js:50:15)
    at Torrent.emit (node:events:518:28)
    at Torrent._markVerified (file:///node_modules/webtorrent/lib/torrent.js:695:10)
    at file:///node_modules/webtorrent/lib/torrent.js:1779:18
    at /node_modules/immediate-chunk-store/index.js:21:7
    at end (/node_modules/run-parallel/index.js:21:15)
    at done (/node_modules/run-parallel/index.js:25:10)
    at each (/node_modules/run-parallel/index.js:31:7)
```
crashes my server, although as far as I can tell I'm not using the torrent after it's destroyed.

not to crash my server, especially if I can't catch this error or debug it

uriva:
instead of throwing when critical is called, log a warning (as this crashes the entire server and cannot be caught)

fixes #2923

### DIFF
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1076,7 +1076,7 @@ export default class Torrent extends EventEmitter {
 
   critical (start, end) {
     if (this.destroyed) {
-      console.warning('ignoring critical request, torrent is destroyed', start, end)
+      this.emit('warning', new Error(`ignoring critical on destroyed torrent ${start}-${end}`))
       return
     }
 
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -1075,7 +1075,10 @@ export default class Torrent extends EventEmitter {
   }
 
   critical (start, end) {
-    if (this.destroyed) throw new Error('torrent is destroyed')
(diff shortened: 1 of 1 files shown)


uriva:
yup
@SilentBot1 @ThaUnknown maybe you can help?

ThaUnknown:
this a patch, not a fix, it doesn't fix the underlying issue that causes this problem and will not be accepted

uriva:
I pinged you because you didn't respond on the issue or the PR

What do you suggest? Without this change the server crashes suddenly.

PR #2981, opened 2025-09-30
fix: Do not verify pieces which are not contained in any selection

### LINKED ISSUE #331: --select / -s downloads the whole torrent instead of the specified file
5.1.3 (2.6.10)

Microsoft Windows [Version 10.0.19045.5131] / 24.3.0 / 11.4.2

**What happened?**
When using --select / -s the whole torrent still gets downloaded

To only download / stream the selected file

**Are you willing to submit a pull request to fix this bug?**
No

crass:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

Update `_verifyPiecesUsingHash` to only verify pieces that are covered in a selection range. Also update the comment documenting `rescanFiles` to reflect that `rescanFiles` only validates the selected pieces.

When selecting a subset of files to download, the Selections object is initially populated correctly with ranges covering only the needed pieces. However, when file verification is enabled (the default), all pieces will be verified. On initial downloads all of the pieces will fail verification. This causes the `_markUnverified` method to be called for all pieces, which in turn adds selection ranges for all the pieces. Thus all the pieces are selected, even though only a subset was desired.

 From my reading its okay to rely on the selections as a list of pieces that we care about verifying. Am I wrong? Is it appropriate to change the behavior of `rescanFiles`? I think so. If you want to rescan all the files, just select all the files.

Also, I believe this fixes [webtorrent-cli #331](https://github.com/webtorrent/webtorrent-cli/issues/331).

### DIFF
--- a/lib/file.js
+++ b/lib/file.js
@@ -19,6 +19,7 @@ export default class File extends EventEmitter {
     this.size = file.length
     this.type = mime.getType(this.name) || 'application/octet-stream'
     this.offset = file.offset
+    this.selected = file.selected || false
 
     this.done = false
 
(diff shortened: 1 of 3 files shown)


crass:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?

Yes, still relevant.

PR #2982, opened 2025-09-30
Add selectOnly option for torrent creation

crass:

[ ] Documentation update
[ ] Bug fix
[X] New feature
[ ] Other, please explain:

This add proper support for selecting a subset of torrent files during creation. It support Arrays and BEP53 compliant strings. Backwards compatibility is kept with applications that used the now depreciated method of setting the `torrent.so` property.

Are there other places that documentation should be updated? I'm not sure if the `torrent.so` interface was documented (I can't find it). If so, it should be labeled as depreciated.


crass:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?

I wish there was an answer other than no the last question.

crass:
Damn shame. Looks like Webtorrent is dead.

PR #331, opened 2015-05-27
Web (HTTP) Seeding

olalonde:
WIP (do not merge). See https://github.com/feross/webtorrent/issues/67
Depends on this PR: https://github.com/feross/bittorrent-swarm/pull/19

I started working on a web seed implementation. The implementation right now is a bit naive (it maps every bittorrent protocol requests to HTTP `Range` requests) and could probably be optimized.

- [x] downloads from web seeds listed in `url-list`
- [ ] test in web browser
- [ ] test what happens when regular peers and web peer

PR #3011, opened 2026-03-18
fix: prevent _markUnverified from re-selecting deselected file pieces

semikolon:
## Summary

Fixes #331 — `--select` / `file.deselect()` doesn't prevent downloading unselected files.

**Root cause**: `_markUnverified` (line 890 of `lib/torrent.js`) unconditionally calls `this.select(index, index)` when a piece is marked as unverified. This re-adds pieces from files the user explicitly deselected, causing the entire torrent to download regardless of file selection.

The only existing guard (`this._startAsDeselected`) only applies when the torrent was created with `opts.deselect = true` — it doesn't protect files deselected after torrent creation (the `-s` flag scenario in webtorrent-cli).

## Fix

- Add `Selections.contains(pieceIndex)` method that checks whether a piece falls within any active non-stream selection range
- Guard the `this.select(index, index)` call in `_markUnverified` with `this._selections.contains(index)`
- Unverified pieces from deselected files are no longer re-added to the download queue

## Changes

- `lib/torrent.js`: 1-line change — add `contains()` guard before re-selecting
- `lib/selections.js`: Add `contains(pieceIndex)` method (~15 lines)
- `test/selections.js`: 3 new test cases for `contains()` (range matching, stream selection filtering, empty selections)

## Test results

All 99 selection tests pass (96 existing + 3 new).

## Notes

- PR #2981 identified the same root cause but proposed a much larger architectural rewrite (4+ years stale, unmergeable). This is a minimal targeted fix
- Shared pieces at file boundaries are handled correctly: if file A and B share a piece and only B is selected, the piece remains in B's selection range, so `contains()` returns true
- Performance is negligible: `contains()` iterates the selections list (typically < 10 items) per unverified piece

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/selections.js
+++ b/lib/selections.js
@@ -141,6 +141,21 @@ export class Selections {
     this._items.length = 0
   }
 
+  /**
+   * Check if a piece index falls within any non-stream selection range.
+   * Used by _markUnverified to avoid re-selecting pieces from deselected files.
+   * @param {number} pieceIndex
+   * @returns {boolean}
+   */
+  contains (pieceIndex) {
+    for (const item of this._items) {
+      if (!item.isStreamSelection && pieceIndex >= item.from && pieceIndex <= item.to) {
+        return true
+      }
+    }
+    return false
+  }
+
   /** @returns {Generator<SelectionItem & {remove: function}>} */
   * [Symbol.iterator] () {
     for (let i = 0; i < this._items.length; i++) {
--- a/lib/torrent.js
+++ b/lib/torrent.js
@@ -887,7 +887,7 @@ export default class Torrent extends EventEmitter {
       : this.pieceLength
     this.pieces[index] = new Piece(len)
     this.bitfield.set(index, false)
-    if (!this._startAsDeselected) this.select(index, index)
+    if (!this._startAsDeselected && this._selections.contains(index)) this.select(index, index)
     for (const file of this.files) {
       if (file.done && file.includes(index)) file.done = false
     }
--- a/test/selections.js
+++ b/test/selections.js
@@ -82,6 +82,40 @@ test('Selections', (t) => {
     }
   }
 
+  t.test('contains should return true for pieces inside a selection range', (s) => {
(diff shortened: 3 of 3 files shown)


semikolon:
## Verification Results

Tested with a 27-file season pack (Legion Complete Series S01-S03, 12,010 pieces × 2MB, ~24GB total) selecting only file index 4 (S01E05, 859MB, pieces 1987-2416).

### Before patch (unpatched webtorrent 2.8.5):
- **7 unwanted files** downloaded across all 3 seasons (~4.1GB actual disk usage)
- S01E01 (456MB), S01E04 (1.1GB), S02E01 (1.1GB), S02E02 (908MB), S03E04 (511MB), plus fragments

### After patch:
- **Target file (S01E05)**: 859MB — correct ✅
- **Adjacent file (S01E04)**: 1.1GB **apparent** size but only **244KB actual disk usage** — sparse file from one shared boundary piece (piece 1987 spans both files) ✅
- **Adjacent file (S01E06)**: 1.5MB — single shared boundary piece ✅
- **All other 24 files**: not downloaded at all ✅
- **Total disk waste**: ~1.7MB (two boundary pieces)

### How the fix works with BEP 53

The torrent's `_onMetadata` correctly uses the BEP 53 path when `torrent.so` is set: only `files[4].select()` is called, creating a selection range of pieces 1987-2416. The `else` branch (`select(0, pieces.length-1)`) does not run.

Without this patch, `_markUnverified` during piece verification would call `select(index, index)` on every unverified piece, effectively overriding the BEP 53 selection and requesting all 12,010 pieces. With the patch, `_selections.contains(index)` returns false for pieces outside the 1987-2416 range, so they stay deselected.

The 244KB and 1.5MB on adjacent files are expected: pieces 1987 and 2416 span file boundaries (S01E04→S01E05 and S01E05→S01E06), and BitTorrent downloads full pieces. The filesystem creates sparse files for the adjacent file portions.

PR #3039, opened 2026-05-22
chore: improve tests perf and speed

ThaUnknown:

[ ] Documentation update
[ ] Bug fix
[ ] New feature
[x] Other, please explain:

changes to tests so they dont take 15 minutes....

fixes a lot of problems:

browser tests will sometimes fail as airtap is flakey, currently it needs to re-run the very long node tests to re-try tape tests, just for them to occasionally fail
this is fixed by making browser tests a standalone action and make them timeout within 30s not 5min

npm install took over a minute
switched to pnpm, now takes 15s, requires dropping of node 16 for security, could use an older version of pnpm that supports node16, but doesn't offer all the security improvements pnpm brings over npm, so decided to drop it
this also not commits the package lock so there should be no more dependency resolving issues in actions, where it would work locally and not in actions
additionally using a dual-node version setup could be possible, where pnpm uses a newer node version for install and older for running tests, but i dont have experience with that
since pnpm's dependency resolution is more strict for security, tests needed sine packages to be imported in the root package rather than sub-dependencies, such as events, randombytes, simple-get, in the future the tests could be re-written to drop those, but that's out of scope

timeouts in node tests took a long time, specifically noPeersInterval time, and http server connections making the server linger for a few seconds
simple fix, force-close connections before closing http servers, and lower noPeersIntervalTime


.gitignore
```diff
@@ -1,5 +1,4 @@
 node_modules/
-package-lock.json
```
DiegoRBaquero, line 2: Why?

ThaUnknown: because otherwise the cache action fails, it has been broken for 4 years now

DiegoRBaquero (approved):

